PR #259, opened 2019-10-15
Fix POST gameroom/propose endpoint

eloaverona:
Currently, when a gameroom is proposed, the Gameroom UI passes back metadata including the connection URL, like so:
```
POST /gamerooms/propose

{
    “alias”: “Acme + Bubba”,
    “members”: [
        {
           “identity”: “bubba-node-000”,
            “metadata”: {
                  “organization”: “Bubba Bakery”,
                  “endpoint”: “tls://splinterd-node-bubba:8044”,
             }
        }
    ],    
}
```
The Gameroom REST API should instead lookup the URL based on the identity string alone to prevent the browser from using arbitrary values (correctness should be enforced by the gameroom REST API). And thus, the "members" list should just be a list of identities. This PR applies these changes.

### DIFF
--- a/examples/gameroom/daemon/src/rest_api/routes/gameroom.rs
+++ b/examples/gameroom/daemon/src/rest_api/routes/gameroom.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 use std::collections::HashMap;
 
-use actix_web::{error, web, Error, HttpResponse};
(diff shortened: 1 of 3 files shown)


eloaverona:
Closing this, and reopening at https://github.com/Cargill/splinter/pull/268 as I had to create a new fork after the repo went public

PR #263, opened 2019-10-18
Enforce a node id is unique on a network

agunde406:
Before it was possible, when using Trust Authorization,
for a new node to "steal" the node id of an already connected
node. This change will remove a connection if it tries to
authorize for a node id that is already connected and
authorized.


agunde406:
Merge conflicts, need to reopen from new fork.

PR #267, opened 2019-10-22
Check for batch status after batch is submitted

eloaverona:
When a batch is submitted via the `/gamerooms/{circuit_id}/batches` endpoint in Gameroomd it will submit the batch and then wait for the batch to be processed and return the batch status (default max wait is 30secs).

For now, if a transaction fails the UI will show a toast with a message about the invalid transaction. In the future a more robust implementation for logging and storing the batch status should be implemented.

### DIFF
--- a/examples/gameroom/daemon/openapi.yml
+++ b/examples/gameroom/daemon/openapi.yml
@@ -633,6 +633,13 @@ paths:
           required: true
           schema:
             type: string
+        - name: wait
+          in: query
+          description: Time gameroomd should wait for the batch to be processed. If set to "false" gameroomd will not wait.
+          required: false
+          schema:
+            type: integer
+            default: 30
       requestBody:
         content:
           application/octet-stream:
@@ -640,21 +647,42 @@ paths:
               type: string
               format: binary
       responses:
-          202:
-            description:  The payload was accepted
+          200:
+            description:  A list with the batch statuses of the submitted batches.
             content:
               application/json:
                 schema:
                   properties:
                     data:
-                      type: string
-                      example: "The payload was submitted successfully"
+                      type: array
+                      items:
+                        $ref: '#/components/schemas/BatchStatus'
+          202:
+            description:  Returned if any of the batches are still pending after timeout or if wait is set to false. Response includes a list with the batch statuses of the submitted batches.
+            content:
+              application/json:
+                schema:
+                  properties:
+                    data:
+                      type: array
+                      items:
+                        $ref: '#/components/schemas/BatchStatus'
           400:
             description: Request was malformed
             content:
               application/json:
                 schema:
-                  $ref: '#/components/schemas/Error'
+                  allOf:
+                    - $ref: '#/components/schemas/Error'
+                    - type: object
+                      properties:
+                          data:
+                            description: List of BatchStatus
+                            type: array
+                            items:
+                              $ref: '#/components/schemas/BatchStatus'
+
+
           500:
             description: Internal server error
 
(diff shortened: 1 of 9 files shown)


libsplinter/src/service/scabbard/state.rs
```diff
@@ -358,10 +358,14 @@ impl From<InvalidTransactionResult> for InvalidTransaction {
 pub struct BatchInfo {
     pub id: String,
     pub status: BatchStatus,
-    #[serde(skip_serializing)]
+    #[serde(skip, default = "get_time_now")]
```
peterschwarz: Can you just use `"System::now"` as this value? Also, probably a good idea to use `skip_deserializing`

examples/gameroom/daemon/openapi.yml
```diff
@@ -629,22 +629,31 @@ paths:
...
       requestBody:
         content:
           application/octet-stream:
             schema:
               type: string
               format: binary
       responses:
-          202:
-            description:  The payload was accepted
+          200:
```
peterschwarz: In other implementations (sawtooth, etc) I think we still have this as a 202

libsplinter/src/service/scabbard/state.rs
eloaverona: I can use `System::now` I'll make that change. `skip` skips both serialization and deserialization.  So I thought it was the best option to keep the logic of the original `skip_serializing` while also doing the `skip_deserializing`. 

https://serde.rs/field-attrs.html#skip

examples/gameroom/daemon/openapi.yml
dplumb94: I think it should be 200 if the batch status was waited on and it came out as a success, and 202 if wait is set to false. 202 implies that we don't know if the request succeeded or not, just that it was accepted.

eloaverona: updated

examples/gameroom/daemon/openapi.yml
```diff
@@ -1060,3 +1069,38 @@ components:
...
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        transaction_id:
+                          type: string
+                          example: f4e147ff464013deccb3e68bb8619beffb29ff86b401257c93bcf8ef76d7ca173fa84b4f4a58414ad2d00a2c9f810cbb726e01cd26ebd44720239d9d35853099
+                        error_message:
+                          type: string
+                          example: "Wasm contract retruned invalid transaction: xo, 0.3.3"
```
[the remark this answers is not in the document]
eloaverona: Also, added a commit to fix the type in the splinterd openapi

examples/gameroom/daemon/src/rest_api/routes/submit.rs
```diff
@@ -64,39 +77,263 @@ pub fn submit_scabbard_payload(
...
+                        if batches_info
+                            .iter()
+                            .any(|batch_info| match batch_info.status {
+                                BatchStatus::Pending => true,
+                                BatchStatus::Valid(_) => true,
+                                _ => false,
+                            })
+                            && Instant::now().duration_since(start_time) < Duration::from_secs(wait)
+                        {
+                            sleep(Duration::new(1, 0));
```
peterschwarz: What's this sleep for?

eloaverona: I thought it would be better to wait a little before sending the request again. So, that gameroomd doesn't end up sending too many requests to the batch status endpoint in case the batches take long to be committed.

peterschwarz: Ah, I guess a comment and using `Duration::from_secs(1)` might improve clarity.

libsplinter/src/service/scabbard/state.rs
```diff
@@ -358,7 +358,7 @@ impl From<InvalidTransactionResult> for InvalidTransaction {
 pub struct BatchInfo {
     pub id: String,
     pub status: BatchStatus,
-    #[serde(skip_serializing)]
+    #[serde(skip, default = "SystemTime::now")]
```
agunde406: reasoning for this change should be added to the commit

examples/gameroom/daemon/src/rest_api/routes/submit.rs
```diff
@@ -64,39 +77,263 @@ pub fn submit_scabbard_payload(
...
+                            Err(err) => {
+                                return Either::B(future::err(RestApiResponseError::InternalError(
+                                    format!("Failed to parse response body {}", err),
+                                )))
+                            }
+                        };
+
+                        let message = match body_value.get("message") {
+                            Some(value) => value.as_str().unwrap_or("Request was malformed."),
+                            None => "Request malformed.",
```
agunde406: Is there a reason this is "Request malformed" instead of "Request was malformed."?

eloaverona: No reason... I changed them to say the same thing.

libsplinter/src/service/scabbard/state.rs
eloaverona: addressed

examples/gameroom/daemon/src/rest_api/routes/submit.rs
eloaverona: addressed

PR #271, opened 2019-10-23
Add health feature to splinterD

RyanLassigBanks:
Adds Health service skeleton, connects it to the SplinterD, and gates it behind the `health` feature.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+[features]
+experimental = ["health"]
+health = ["reqwest", "serde_json"]
+
 [package]
 name = "splinter-cli"
 version = "0.3.6"
@@ -26,7 +30,9 @@ protobuf = "2"
 simple_logger = "0.4.0"
 libc = "0.2"
 splinter = { path = "../libsplinter" }
+reqwest = { version = "0.9", optional = true }
 sawtooth-sdk = "0.3"
 serde = "1.0"
+serde_json = { version ="1.0", optional = true }
 serde_yaml = "0.8"
 serde_derive = "1.0"
--- a/cli/src/action/mod.rs
+++ b/cli/src/action/mod.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 pub mod admin;
+#[cfg(feature = "health")]
+pub mod health;
 
 use std::collections::HashMap;
 
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -28,7 +28,7 @@ const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 fn run() -> Result<(), CliError> {
-    let matches = clap_app!(myapp =>
+    let mut app = clap_app!(myapp =>
         (name: APP_NAME)
         (version: VERSION)
         (author: "Cargill")
@@ -62,8 +62,32 @@ fn run() -> Result<(), CliError> {
                 (@arg quiet: -q --quiet "do not display output")
             )
         )
-    )
-    .get_matches();
+    );
+
+    #[cfg(feature = "health")]
+    {
+        use clap::{Arg, SubCommand};
+
+        app = app.subcommand(
+            SubCommand::with_name("health")
+                .about("displays information about network health")
+                .subcommand(
+                    SubCommand::with_name("status")
+                        .about(
+                            "displays a node's version, endpoint, node id, and a list\n\
+                             of endpoints of its connected peers",
+                        )
+                        .arg(
+                            Arg::with_name("url")
+                                .short("U")
+                                .takes_value(true)
+                                .help("URL of node"),
+                        ),
+                ),
+        );
+    }
+
+    let matches = app.get_matches();
 
     let logger = match matches.occurrences_of("verbose") {
         0 => simple_logger::init_with_level(LogLevel::Warn),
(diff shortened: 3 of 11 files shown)


cli/src/main.rs
```diff
@@ -62,6 +62,13 @@ fn run() -> Result<(), CliError> {
                 (@arg quiet: -q --quiet "do not display output")
             )
         )
+        (@subcommand health =>
+            (about: "Splinter commands for inspecting network integrety and health")
```
peterschwarz: Should match the starting case of the other descriptions - same for the `status` subcommand.

splinterd/src/routes/status.rs
```diff
@@ -12,24 +12,35 @@
...
 
 pub fn get_status(
     node_id: String,
     endpoint: String,
+    peer_map: Arc<RwLock<PeerMap>>,
 ) -> Box<dyn Future<Item = HttpResponse, Error = Error>> {
+    let connected_peers = if let Ok(inner) = peer_map.read() {
+        inner.peer_endpoints()
+    } else {
+        return Box::new(HttpResponse::InternalServerError().finish().into_future());
```
peterschwarz: Can we log the error here?  Otherwise, we won't know about the internal error.

cli/src/main.rs
```diff
@@ -62,6 +62,13 @@ fn run() -> Result<(), CliError> {
...
+            (@subcommand status =>
+                (about: "Check status of node")
```
chenette: The description seems too terse. How about clarifying the specific information that `status` displays?

For example:
"shows a node's connected peers, {status?} endpoint, node ID, and version {of what?}"

Or, with more details:  
"queries the status endpoint for a node's connected peers, node ID, and other status information"

_Style note: The first word should start with a lower-case letter and end with "s" to be consistent with the other descriptions (keygen & keyregistry say "generates")._

cli/src/main.rs
```diff
@@ -62,6 +62,14 @@ fn run() -> Result<(), CliError> {
                 (@arg quiet: -q --quiet "do not display output")
             )
         )
+        (@subcommand health =>
+            (about: "displays information about network health")
```
vaporos: The first letter should be capitalized so output in "carto run -- --help" is consistent.

vaporos: This description probably becomes outdated when we add the active tests. Maybe "Health commands" to match "Administrative commands", though something slightly more descriptive (but not verbose) might be better.

cli/src/main.rs
```diff
@@ -62,6 +62,14 @@ fn run() -> Result<(), CliError> {
...
+            (@subcommand status =>
+                (about: "displays a node's version, endpoint, node id, and a list \
```
vaporos: Shorten this so it fits in <80 columns. This isn't always possible to achieve w/clap but where we can, we should. Maybe "Display node-level health information".

cli/src/main.rs
```diff
@@ -62,6 +62,14 @@ fn run() -> Result<(), CliError> {
...
+                of the endpoints of its connected peers")
+                (@arg url: -u +takes_value "URL of node")
```
vaporos: Add the long argument '--url'. Not sure if this should be -U or -u. I seem to recall -U being more popular for this but haven't found specific examples yet.

vaporos: Should we say "URL of node's health REST API" to clarify which "URL"?

cli/src/main.rs
peterschwarz: If that's the case, it should be done in an additional commit, as the other messages are lowercase

vaporos:
The status information should differentiate between "peers" (other splinter nodes) and connected services.

{
  "peers": [ {
    "endpoint": "tls://172.22.0.8:35612",
    "status": "connected"
  },
  {
 "endpoint": "tls://172.22.0.9:35612",
    "status": "disconnected",
    "error": "Connection refused",
  }],
  "service_processors": [
      {
        "endpoint": "inproc://admin-service"
        ...
      }
    ]
  ],
  "endpoint": "0.0.0.0:8043",
  "node_id": "bubba-node-000",
  "version": "0.3.3"
}

We should also add in a list of circuits and their health status:

"circuits": [
{
    "name": "foo",
    "members": [
        "tls://172.22.0.8:35612",
        "tls://172.22.0.9:35612"
    ],
    "peering_status": "incomplete"
}]

cli/src/action/mod.rs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 pub mod admin;
+pub mod health;
```
vaporos: This new part of the CLI should be gated by a feature.

cli/Cargo.toml
```diff
@@ -26,7 +26,9 @@ protobuf = "2"
 simple_logger = "0.4.0"
 libc = "0.2"
 splinter = { path = "../libsplinter" }
+reqwest = "0.9"
```
vaporos: These should probably be pulled in by the health feature, when enabled.

libsplinter/src/network/mod.rs
```diff
@@ -51,7 +51,7 @@ impl NetworkMessage {
     }
 }
 
-struct PeerMap {
+pub struct PeerMap {
```
vaporos: This feels like it should remain an implementation detail of the network layer, and if we expose it, we should consider if this is truly the right data structure across the API boundary. In any case, this does not differentiate between the types of "peers" (splinter<->splinter) vs (splinter<->service).

cli/src/main.rs
chenette: Isn't there a way to provide multi-line documentation without violating clap's line-length restrictions? I am in favor of clear & helpful information over short lines.

splinterd/Cargo.toml
```diff
@@ -43,6 +44,7 @@ tempdir = "0.3"
 
 [features]
 default = []
+health-service-enabled = [ "health" ]
```
peterschwarz: this feature should probably be just `health-service` (as enabled state is determined by if the feature is included in the build)

splinterd/src/daemon.rs
```diff
@@ -358,19 +370,43 @@ impl SplinterDaemon {
         })
         .expect("Error setting Ctrl-C handler");
 
-        main_loop_join_handle
-            .join()
-            .map_err(|_| StartError::ThreadError("Unable to join main loop".into()))?;
+        #[cfg(feature = "health")]
```
peterschwarz: Does this block need cover all the join logic?

splinterd/src/daemon.rs
```diff
@@ -475,6 +511,70 @@ impl SplinterDaemon {
...
+#[cfg(feature = "health")]
+fn start_health_service(
+    mut transport: InprocTransport,
+    health_service: HealthService,
+    running: Arc<AtomicBool>,
+) -> Result<service::JoinHandles<Result<(), service::error::ServiceProcessorError>>, StartError> {
+    let start_health_service: std::thread::JoinHandle<
+        Result<service::JoinHandles<Result<(), service::error::ServiceProcessorError>>, StartError>,
+    > = thread::spawn(move || {
+        let incoming_capacity = 8;
```
peterschwarz: might as well make these constants, like the admin service equivalents.

splinterd/Cargo.toml
vaporos: Also, move it below experimental below and add to the experimental block.

PR #274, opened 2019-10-23
Increase Gameroom paging limit

shannynalayna:
Increases the default paging limit to 1000 and also fixes the tests to reflect this modification.

### DIFF
--- a/examples/gameroom/daemon/src/rest_api/routes/mod.rs
+++ b/examples/gameroom/daemon/src/rest_api/routes/mod.rs
@@ -37,6 +37,8 @@ use serde::{Deserialize, Serialize};
 
 pub const DEFAULT_LIMIT: usize = 100;
 pub const DEFAULT_OFFSET: usize = 0;
+const MAX_LIMIT: usize = 1000;
+
 const QUERY_ENCODE_SET: &AsciiSet = &CONTROLS
     .add(b' ')
     .add(b'"')
@@ -126,7 +128,7 @@ pub fn get_response_paging_info(
     link: &str,
     query_count: usize,
 ) -> Paging {
-    let limit = limit as i64;
+    let limit = validate_limit(limit);
     let offset = offset as i64;
     let query_count = query_count as i64;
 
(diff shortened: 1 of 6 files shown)


examples/gameroom/daemon/src/rest_api/routes/mod.rs
```diff
@@ -174,44 +174,46 @@ mod tests {
...
+            create_test_paging_response(DEFAULT_LIMIT, DEFAULT_OFFSET, 1000, 0, 9000);
         assert_eq!(test_paging_response, generated_paging_response);
     }
 
     #[test]
     fn test_50offset_paging_response() {
-        // Create paging response from default limit, offset of 50, and a total of 1000
-        let test_paging_response = get_response_paging_info(DEFAULT_LIMIT, 50, TEST_LINK, 1000);
-        let generated_paging_response = create_test_paging_response(DEFAULT_LIMIT, 50, 150, 0, 900);
+        // Create paging response from default limit, offset of 500, and a total of 10000
```
dplumb94: Is it necessary to update the tests besides `test_default_paging_response()`? If so, the names of the tests should be updated as well

examples/gameroom/gameroom-app/src/store/api.ts
```diff
@@ -79,7 +79,7 @@ export async function gameroomPropose(
 }
 
 export async function listGamerooms(): Promise<Gameroom[]> {
-  const response = await gameroomAPI.get('/gamerooms');
+  const response = await gameroomAPI.get('/gamerooms?limit=1000');
```
peterschwarz: Does the server side have a max limit?

shannynalayna: No, the limit gets passed directly to the database query. I don't see any limit maximum documented from diesel.

peterschwarz: We should add a maximum on the server-side for this, as well.  Perhaps in a separate commit.

examples/gameroom/daemon/src/rest_api/routes/mod.rs
```diff
@@ -166,6 +166,15 @@ pub fn get_response_paging_info(
     }
 }
 
+pub fn validate_limit(limit: usize) -> i64 {
+    let max = 1000 as usize;
```
peterschwarz: This could be a `const`

PR #277, opened 2019-10-24
Add design system initial commit

dnewh:


zapplebee:
The readme should be updated from the Create React App default to be relevant to the project.

zapplebee:
eslint config looks good.

canopy/design-system/src/styles/layout.scss
```diff
@@ -0,0 +1,119 @@
+.margin-s {
```
zapplebee: This should be put into to a function so it is easy to manage changing the values and keep it DRY

canopy/design-system/.gitignore
```diff
@@ -0,0 +1,26 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
vaporos: .gitignore should only be done at the top-level of the repo (as per our convention).

canopy/design-system/README.md
```diff
@@ -0,0 +1,68 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
vaporos: Wrap at <80 columns.

canopy/design-system/src/App.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
```
vaporos: Need copyright headers on all files.

canopy/design-system/yarn.lock
```diff
@@ -0,0 +1,11387 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
vaporos: No autogenerated files should be checked into the repo (at least on master).

vaporos:
Do we need to maintain .eslint config in every subdirectory?

vaporos:
Please remove the .keep file in a separate commit.

canopy/design-system/yarn.lock
dnewh: resolved

canopy/design-system/src/App.js
dnewh: resolved

canopy/design-system/README.md
dnewh: resolved

canopy/design-system/src/styles/layout.scss
dnewh: resolved

dnewh:
> Do we need to maintain .eslint config in every subdirectory?

Yes, they are similar but not exactly the same between the projects and also have dependencies in each project

canopy/design-system/README.md
```diff
@@ -0,0 +1,89 @@
+This project was bootstrapped with
```
vaporos: Is this README.md content actually useful? Looks like maybe it was autogenerated and mostly points to facebook URLs and isn't really canopy/design-system specific.

dnewh: It is the generic create-react-app README and will get updated later but it does, for now, list the relevant commands to run the app

PR #279, opened 2019-10-24
Add intial implementation of heartbeats to the Splinter Network

agunde406:
This commits adds a loop to Network that will send
heartbeats to every peer based on a provided heartbeat
interval. This will ensure that connections are not
incorrectly closed. 

Currently, every peer will receive
the heartbeat even if the network has received a message
from that peer recently. This will be changed in a future PR.

The default is 30 seconds. Setting the heartbeat_interval to 0 turns off heartbeats.

### DIFF
--- a/libsplinter/src/network/mod.rs
+++ b/libsplinter/src/network/mod.rs
@@ -33,14 +33,14 @@ use crate::mesh::{
 use crate::transport::Connection;
 
 #[derive(Debug)]
-pub struct NetworkMessage {
+pub struct NetworkMessageWrapper {
     peer_id: String,
     payload: Vec<u8>,
 }
 
-impl NetworkMessage {
+impl NetworkMessageWrapper {
     pub fn new(peer_id: String, payload: Vec<u8>) -> Self {
-        NetworkMessage { peer_id, payload }
+        NetworkMessageWrapper { peer_id, payload }
     }
     pub fn peer_id(&self) -> &str {
         &self.peer_id
@@ -229,7 +229,7 @@ impl Network {
         Ok(())
     }
 
-    pub fn recv(&self) -> Result<NetworkMessage, RecvError> {
+    pub fn recv(&self) -> Result<NetworkMessageWrapper, RecvError> {
         let envelope = self.mesh.recv()?;
         let peer_id = match rwlock_read_unwrap!(self.peers).get_peer_id(envelope.id()) {
             Some(peer_id) => peer_id.to_string(),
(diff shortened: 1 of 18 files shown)


splinterd/src/main.rs
```diff
@@ -192,6 +196,9 @@ fn main() {
         .or_else(|| config.peers())
         .unwrap_or_default();
 
+    let heartbeat_interval = value_t!(matches.value_of("heartbeat_interval"), u64)
+        .unwrap_or_else(|_| config.heartbeat_interval().unwrap_or(30));
```
peterschwarz: magic number

libsplinter/src/network/mod.rs
```diff
@@ -157,11 +160,41 @@ pub struct Network {
...
+            heartbeat_message.set_message_type(NetworkMessageType::NETWORK_HEARTBEAT);
+            heartbeat_message.set_payload(heartbeat);
+            let heartbeat_bytes = heartbeat_message
+                .write_to_bytes()
+                .map_err(|_| NetworkStartUpError("cannot create NetworkMessage".to_string()))?;
+            let _ = thread::spawn(move || {
+                let interval = Duration::from_secs(heartbeat_interval);
+                thread::sleep(interval);
+                loop {
+                    let peers = rwlock_read_unwrap!(heartbeat_network.peers).peer_ids();
```
peterschwarz: This will ping all peers, temp or otherwise

libsplinter/src/network/mod.rs
```diff
@@ -157,11 +160,41 @@ pub struct Network {
...
+            })?;
+            let mut heartbeat_message = NetworkMessage::new();
+            heartbeat_message.set_message_type(NetworkMessageType::NETWORK_HEARTBEAT);
+            heartbeat_message.set_payload(heartbeat);
+            let heartbeat_bytes = heartbeat_message
+                .write_to_bytes()
+                .map_err(|_| NetworkStartUpError("cannot create NetworkMessage".to_string()))?;
+            let _ = thread::spawn(move || {
+                let interval = Duration::from_secs(heartbeat_interval);
+                thread::sleep(interval);
```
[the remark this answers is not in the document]
agunde406: It will go into the next PR.

libsplinter/src/network/mod.rs
agunde406: I meant to not turn on the auth guard around the heartbeats so this shouldn't matter, will update

splinterd/src/main.rs
agunde406: fixed

libsplinter/src/network/mod.rs
agunde406: fixed

PR #288, opened 2019-10-27
Add repo and namespace variables to installed compose file

rbuysse:
This allows us to easily publish to multiple registries.


rbuysse:
These values default to empty. We can change the defaults by adding values to the `.env` file at the root of the repository.

PR #289, opened 2019-10-28
Simplify Service where services can be connected

agunde406:
Before a Service could be connected at any number of allowed nodes. However to alert the rest of the network of their location, the other nodes would need to be explicitly told which node the service was connected on and there was not a catch up procedure implemented if the other node missed the message. This would cause part of the network to not know where services were connected or how to talk to them. This was also seen on circuit start up where a circuit may not be up yet on one of the nodes.

We have decided to enforce that the service will be accessible at the node listed in their allowed nodes (with an error returned if not there) since this information is in the circuit definition. We have also decided to limit the allowed nodes list to one node. Multiple nodes may be supported in the future (to support rolling upgrades for example).


peterschwarz:
Tested

PR #300, opened 2019-10-30
Update responsive side nav

dnewh:
Updates css classes and variables to provide necessary functionality to create a responsive side nav bar.

### DIFF
--- a/canopy/design-system/src/styles/navigation.scss
+++ b/canopy/design-system/src/styles/navigation.scss
@@ -28,14 +28,24 @@
       color: $text-primary;
     }
   }
-}
 
-.nav-item-expandable {
-  .arrow {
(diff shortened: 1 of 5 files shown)


canopy/design-system/src/styles/navigation.scss
```diff
@@ -44,10 +54,21 @@
...
+
+    .nav-logo {
+      width: 4rem;
+      height: 4rem;
+    }
 
-  &.open {
-    .nested {
-      height: fit-content;
+    &.hidden {
```
dplumb94: are we using `hidden` for all elements we don't want to display? How does that work with `is-open`?

dnewh: updated

PR #301, opened 2019-10-31
Add CircuitManagementPayload verification

shannynalayna:
Adds a method in the admin service to verify the CircuitManagementPayload fields, including the payload and header. Also adds signature verification in the `propose_change` method, where this was previously unchecked.

### DIFF
--- a/libsplinter/src/admin/shared.rs
+++ b/libsplinter/src/admin/shared.rs
@@ -1396,7 +1396,10 @@ mod tests {
         AuthorizationMessage, AuthorizationMessageType, AuthorizedMessage,
     };
     use crate::protos::network::{NetworkMessage, NetworkMessageType};
-    use crate::signing::hash::HashVerifier;
+    use crate::signing::{
+        hash::{HashSigner, HashVerifier},
+        Signer,
+    };
     use crate::storage::get_storage;
     use crate::transport::{
         ConnectError, Connection, DisconnectError, RecvError, SendError, Transport,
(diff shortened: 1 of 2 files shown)


libsplinter/src/admin/shared.rs
```diff
@@ -2066,7 +2069,220 @@ mod tests {
         if let Ok(_) =
             admin_shared.validate_circuit_vote(&vote, b"test_signer_a", &proposal, "node_a")
         {
-            panic!("Should have been invalid becasue the circuit hash does not match");
+            panic!("Should have been invalid because the circuit hash does not match");
+        }
+    }
+
+    #[test]
+    // test that the CircuitManagementPayload returns an error in case the signature is empty.
```
agunde406: technically the CircuitManagementPayload does not return this error

libsplinter/src/admin/shared.rs
```diff
@@ -514,24 +518,29 @@ impl AdminServiceShared {
...
-            Err(ServiceError::UnableToHandleMessage(_)) => (),
-            Err(err) => return Err(err),
-        };
-
         let header =
             protobuf::parse_from_bytes::<CircuitManagementPayload_Header>(payload.get_header())?;
+        self.validate_circuit_management_payload(&payload, &header)
+            .map_err(|err| ServiceError::UnableToHandleMessage(Box::new(err)))?;
+
+        self.verify_signature(&payload)?;
```
agunde406: Either split this out into its own commit or add a message to the commit.

shannynalayna: Split handling the unknown / unset actions into a separate commit and added a description of the signature verification to that existing commit.

PR #304, opened 2019-11-01
Gameroom resubscribes to existing gamerooms on startup

RyanLassigBanks:


examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -271,6 +271,14 @@ fn process_admin_event(
             })
         }
         AdminServiceEvent::CircuitReady(msg_proposal) => {
+            let conn = &*pool.get()?;
```
agunde406: In commit message should "replying" be "replaying"?

examples/gameroom/database/src/helpers/gameroom.rs
```diff
@@ -116,6 +116,22 @@ pub fn gameroom_service_is_active(conn: &PgConnection, circuit_id: &str) -> Quer
         .or_else(|err| if err == NotFound { Ok(false) } else { Err(err) })
 }
 
+pub fn fetch_active_gamerooms(
```
agunde406: This should be renamed to make it clear it is fetching gameroom proposals, not gamerooms

agunde406: and services.

examples/gameroom/database/src/helpers/gameroom.rs
```diff
@@ -104,6 +104,18 @@ pub fn update_gameroom_proposal_status(
         .map(|_| ())
 }
 
+pub fn gameroom_service_is_active(conn: &PgConnection, circuit_id: &str) -> QueryResult<bool> {
+    gameroom_service::table
```
peterschwarz: You could make this slightly more efficient by not querying the whole record

examples/gameroom/database/src/helpers/gameroom.rs
peterschwarz: This change could also be pulled into a separate commit, like the other query function in this PR

examples/gameroom/database/Cargo.toml
```diff
@@ -22,3 +22,4 @@ license = "Apache-2.0"
 [dependencies]
 diesel = { version = "1.0.0", features = ["postgres", "r2d2", "serde_json"] }
 serde_json = "1.0"
+log = "0.4"
```
peterschwarz: I don't see this being used in these changes.

examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -406,6 +413,48 @@ fn process_admin_event(
     }
 }
 
+fn resubscribe(
```
ltseeley: It seems like all of this is duplicated in the `CircuitReady` branch above. Can you rename this method to `subscribe` and call it from the new location as well as in the `CircuitReady` branch?

RyanLassigBanks: They're close, The WebSocket in `CircuitReady` has an `on_open` callback that creates the xo contract initially. `resubscribe` doesn't have this because it's assumed that the contract already exists when the application authorization handler attempts to resubscribe.

ltseeley: Since you're just returning the `WebSocketClient` here anyway, why not call this function from the `CircuitReady` handler, then just add the `on_open` callback to the returned client?

peterschwarz: They are also very different argument sources in these two cases, so it sill makes sense that they are separate activities.  Let's leave them as-is.

examples/gameroom/database/src/models.rs
```diff
@@ -194,3 +194,12 @@ pub struct NewXoGame {
     pub created_time: SystemTime,
     pub updated_time: SystemTime,
 }
+
+#[derive(Queryable, PartialEq, Debug)]
+pub struct ActiveGameRoom {
```
agunde406: ActiveGameRoom -> ActiveGameroom

PR #309, opened 2019-11-04
Remove hardcoded protocols for octet-stream submission

dplumb94:
Replaces request-promise with built in XMLHTTPRequest functionality.
This fixes a bug where the app could not submit batches or payloads over
TLS. This is because protocol had to be included in the URL for
request-promise, which was hard coded as http. Removing the dependency
allows us to just use a relative URL which will get handled by the proxy

Signed-off-by: Darian Plumb <dplumb@bitwise.io>

### DIFF
--- a/examples/gameroom/gameroom-app/src/store/api.ts
+++ b/examples/gameroom/gameroom-app/src/store/api.ts
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import axios from 'axios';
-import rp from 'request-promise';
 import {
   GameroomNotification,
   GameroomProposal,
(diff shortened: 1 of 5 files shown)


examples/gameroom/gameroom-app/src/store/api.ts
```diff
@@ -146,7 +146,7 @@ export async function listGames(circuitID: string): Promise<Game[]> {
 export async function submitPayload(payload: Uint8Array): Promise<void> {
   const options = {
     method: 'POST',
-    url: `http://${window.location.host}/api/submit`,
+    url: `${document.location.protocol}//${window.location.host}/api/submit`,
```
peterschwarz: Does this work if you just start the string with `/api/submit` or does it require a complete URL

examples/gameroom/gameroom-app/src/store/api.ts
```diff
@@ -144,43 +175,22 @@ export async function listGames(circuitID: string): Promise<Game[]> {
...
-    headers: {
-      'Content-Type': 'application/octet-stream',
-    },
-  };
-
-  return await rp(options).then((rawBody) => {
+  return await http('POST', `/gamerooms/${circuitID}/batches`, payload, (request: XMLHttpRequest) => {
+    request.setRequestHeader('Content-Type', 'application/octet-stream');
+  }).catch((err) => {
+    throw new Error(err);
```
peterschwarz: This is redundant.

examples/gameroom/gameroom-app/src/store/api.ts
```diff
@@ -144,43 +175,22 @@ export async function listGames(circuitID: string): Promise<Game[]> {
...
-    },
-  };
-
-  await rp(options).then((body) => {
-    return;
-  })
-  .catch((err) => {
-    console.error(err.message);
-    throw new Error('Failed to send request. Contact administrator for help.');
+  await http('POST', '/submit', payload, (request: XMLHttpRequest) => {
```
peterschwarz: Do you need to parse the JSON Response on this one?

dplumb94: If anything goes wrong at this level, I don't think it should get displayed to the user. I am console.erroring the error in `http()` for debugging.

examples/gameroom/gameroom-app/src/store/api.ts
dplumb94: This is required to send the error back to the component to invoke the toast alert

PR #312, opened 2019-11-05
Stop scabbard upon shutdown signal

eloaverona:
This PR add a method to AdminServiceShared to stop all running services, and adds a call to that method up admin service shutdown.

### DIFF
--- a/libsplinter/src/admin/mod.rs
+++ b/libsplinter/src/admin/mod.rs
@@ -188,6 +188,10 @@ impl Service for AdminService {
         // Shutdown event dealers and disconnect active websocket connections
         admin_service_shared.shutdown_event_dealers();
 
+        admin_service_shared
+            .stop_services()
+            .map_err(|err| ServiceStopError::Internal(Box::new(err)))?;
+
         info!("Admin service stopped and disconnected");
 
         Ok(())
--- a/libsplinter/src/admin/shared.rs
+++ b/libsplinter/src/admin/shared.rs
@@ -34,7 +34,7 @@ use crate::network::{
     auth::{AuthorizationInquisitor, PeerAuthorizationState},
     peer::PeerConnector,
 };
-use crate::orchestrator::{ServiceDefinition, ServiceOrchestrator};
+use crate::orchestrator::{ServiceDefinition, ServiceOrchestrator, ShutdownServiceError};
 use crate::protos::admin::{
     AdminMessage, AdminMessage_Type, Circuit, CircuitManagementPayload,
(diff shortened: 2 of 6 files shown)


libsplinter/src/admin/shared.rs
```diff
@@ -1070,6 +1070,30 @@ impl AdminServiceShared {
...
     }
 
+    /// Stops all running services
+    pub fn stop_services(&mut self) -> Result<(), AdminSharedError> {
+        let running_services = self
+            .running_services
+            .drain()
+            .collect::<Vec<ServiceDefinition>>();
+        let shutdown_errors = running_services
+            .into_iter()
```
ltseeley: Looks like this is converting an iterator -> vec -> back to iterator.

libsplinter/src/admin/shared.rs
```diff
@@ -1070,6 +1070,30 @@ impl AdminServiceShared {
...
+            .map(|service| {
+                debug!(
+                    "Stoping service {} in circuit {}",
+                    service.service_type, service.circuit
+                );
+                self.orchestrator.shutdown_service(&service)
+            })
+            .filter_map(|res| if let Err(err) = res { Some(err) } else { None })
```
ltseeley: Could probably do `.filter_map(Result::err)`

libsplinter/src/admin/mod.rs
```diff
@@ -188,6 +188,10 @@ impl Service for AdminService {
         // Shutdown event dealers and disconnect active websocket connections
         admin_service_shared.shutdown_event_dealers();
 
+        admin_service_shared
+            .stop_services()
```
ltseeley: Typo in commit message: up -> upon

libsplinter/src/service/scabbard/mod.rs
```diff
@@ -201,6 +201,7 @@ impl Service for Scabbard {
         &mut self,
         service_registry: &dyn ServiceNetworkRegistry,
     ) -> Result<(), ServiceStopError> {
+        debug!("Stopping scabbard");
```
ltseeley: I'd include the service ID here (for when there's more than once scabbard service running).

libsplinter/src/admin/shared.rs
eloaverona: updated

libsplinter/src/admin/mod.rs
eloaverona: updated

libsplinter/src/service/scabbard/mod.rs
eloaverona: updated

PR #314, opened 2019-11-05
Gameroomd websocket reconnect

eloaverona:
This PR implements:
- Gameroomd app auth handler and xo_state_delta subscribers websocket clients attempt to reconnect if they receive a "close" message from the ws server
- Gameroomd app auth handler and xo_state_delta subscribers websocket clients drop the connect and attempt to reconnect if they get no response or no messages from the server after an specified timeout.

### DIFF
--- a/examples/gameroom/daemon/src/authorization_handler/mod.rs
+++ b/examples/gameroom/daemon/src/authorization_handler/mod.rs
@@ -44,6 +44,15 @@ use crate::application_metadata::ApplicationMetadata;
 
 use self::sabre::setup_xo;
 
+/// default value if the client should attempt to reconnet if ws connection is lost
+const RECONNECT: bool = true;
+
+/// default limit for number of consecutives failed reconnection attempts
+const RECONNECT_LIMIT: u64 = 10;
+
+/// default timeout in seconds if no message is received from server
+const CONNECTION_TIMEOUT: u64 = 60;
+
 pub fn run(
     splinterd_url: String,
     node_id: String,
@@ -68,14 +77,25 @@ pub fn run(
         },
     );
 
+    ws.set_reconnect(RECONNECT);
+    ws.set_reconnect_limit(RECONNECT_LIMIT);
+    ws.set_timeout(CONNECTION_TIMEOUT);
+
     ws.on_error(move |err, ctx| {
         error!("An error occured while listening for admin events {}", err);
-        if let WebSocketError::ParserError { .. } = err {
-            debug!("Protocol error, closing connection");
-            Ok(())
-        } else {
-            debug!("Attempting to restart connection");
-            ctx.start_ws()
+        match err {
+            WebSocketError::ParserError { .. } => {
+                debug!("Protocol error, closing connection");
+                Ok(())
+            }
+            WebSocketError::ReconnectError(_) => {
+                debug!("Failed to reconnect. Closing WebSocket.");
+                Ok(())
+            }
+            _ => {
+                debug!("Attempting to restart connection");
+                ctx.start_ws()
+            }
         }
     });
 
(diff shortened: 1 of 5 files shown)


libsplinter/src/events/ws.rs
```diff
@@ -447,11 +471,22 @@ fn do_shutdown(
 pub struct Context<T: ParseBytes<T> + 'static> {
     igniter: Igniter,
     ws: WebSocketClient<T>,
+    reconnect_count: u64,
+    last_reconnect: SystemTime,
+    wait: Duration,
+    reactor_running: Arc<AtomicBool>,
```
RyanLassigBanks: I don't think `reactor_running` is necessary since `igniter` has the same variable. If `Context` needs to know if it's reactor is running it should call a method from `igniter`.

Additionally, I wonder if it would make more sense for the igniter to check if its reactor is still running an emit an error when context or a callback call `start_ws` or `send`. I don't think that needs to be addressed in this PR, just curious what people think.

eloaverona: yeah, I can use the reactor_running in the igniter, and will make that change. But I need the Context to know that the reactor is still running, otherwise, if the gameroom daemon is stopped while the context is waiting before attempt to reconnect to a ws server, the shutdown would be delayed until the wait time is over
https://github.com/Cargill/splinter/blob/cc31433dfb446221104452e0d0437a7e9779fdc1/libsplinter/src/events/ws.rs#L541-L554

eloaverona: updated

libsplinter/src/events/reactor.rs
```diff
@@ -29,11 +33,15 @@ use crate::events::{ReactorError, WebSocketError};
 pub struct Reactor {
     sender: Sender<ReactorMessage>,
     thread_handle: thread::JoinHandle<()>,
+    running: Arc<AtomicBool>,
```
peterschwarz: This change could be in a separate commit, with more explanation.

libsplinter/src/rest_api/events.rs
```diff
@@ -164,6 +165,10 @@ impl<T: Serialize + Debug + 'static> Actor for EventDealerWebSocket<T> {
         if let Some(recv) = self.recv.take() {
             debug!("Starting Event Websocket");
             ctx.add_stream(recv);
+            ctx.run_interval(Duration::from_secs(30), move |_, ctx| {
```
peterschwarz: magic number

libsplinter/src/events/ws.rs
```diff
@@ -338,6 +414,9 @@ impl<T: ParseBytes<T> + 'static> WebSocketClient<T> {
                                                 err
                                             );
                                         }
+                                        if let Err(err) = context.try_reconnect() {
```
peterschwarz: Should it try in this case?  Isn't this an explicit close?

eloaverona: Yes, this is called if the ws gets a close message from the server. I talked with @agunde406 about this, I guess the expectation was that even if the splinter daemon gracefully closes the connection the gameroomd should try to re-establish with the expectation that the server will return soon? The gameroom does work properly without those connections.

libsplinter/src/events/reactor.rs
eloaverona: updated

libsplinter/src/rest_api/events.rs
eloaverona: updated

examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -68,14 +68,25 @@ pub fn run(
         },
     );
 
+    ws.set_reconnect(true);
+    ws.set_reconnect_limit(10);
```
peterschwarz: Should these be using a configured value (as in the previous commit configures)?

eloaverona: Do you mean just using the defaults establish in the server side?

eloaverona: addressed

PR #316, opened 2019-11-06
Add notification REST API design

dnewh:
Adds the openapi specification for POSTing notifications, GETting all notifications for a user, GETting a specific notification for a user, and PATCHing a notification.

### DIFF
--- a/splinterd/api/static/openapi.yml
+++ b/splinterd/api/static/openapi.yml
@@ -327,7 +327,163 @@ paths:
         500:
           description: Internal service error
 
+  /biome/notifications:
+    post:
+      description: Publish a list of notifications
+      tags:
+        - Biome
+      requestBody:
+        $ref: '#/components/requestBodies/CreateNotifications'
+      responses:
+        200:
+          description: Notifications have been published
+        400:
+          description: Malformed notifications list
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /biome/users/{user_id}/notifications:
+    get:
+      description: Retrieve a list of notifications for a given user from the notifications database
+      tags:
+        - Biome
+      parameters:
+        - name: user_id
+          in: path
+          description: user to retrieve notifications for
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Notifications have been retrieved
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/NotificationProperties'
+        403:
+          description: User is forbidden to retrieve notifications for this user
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        404:
+          description: user {user} not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /biome/users/{user_id}/notifications/{notification_id}:
+    get:
+      description: Fetch a notification for a given user
+      tags:
+        - Biome
+      parameters:
+        - name: user_id
+          in: path
+          description: user to fetch a notification for
+          required: true
+          schema:
+            type: string
+        - name: notification_id
+          in: path
+          description: notification to fetch
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Notification has been retrieved
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NotificationProperties'
+        403:
+          description: User is forbidden to retrieve notifications for this user
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        404:
+          description: Notification {notification} for {user} not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+    patch:
+      description: Update a notification for a given user
+      tags:
+        - Biome
+      parameters:
+        - name: user_id
+          in: path
+          description: user to fetch a notification for
+          required: true
+          schema:
+            type: string
+        - name: notification_id
+          in: path
+          description: notification to fetch
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: Notification has been updated
+        403:
+          description: User is forbidden to update notifications for this user
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        404:
+          description: Notification {notification} for {user} not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
 components:
+  requestBodies:
+    CreateNotifications:
+      description: Create new notifications
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: '#/components/schemas/NotificationProperties'
+              - $ref: '#/components/schemas/NotificationRecipients'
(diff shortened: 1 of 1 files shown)


splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /user/users/{user}/notifications:
+    get:
+      description: Retrieve a list of notifications for a given user from the notifications database
+      parameters:
+        - name: user
```
eloaverona: Naming this `user` makes it very unclear what it means. Maybe `user-id` or `user-identifier`?

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/Error'
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /user/users/{user}/notifications:
```
eloaverona: Why `/user/users`?  `/user/{user}/notifications` seems a little cleaner to me

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+              schema:
+                type: array
+                $ref: '#/components/schemas/Notification'
+        403:
+          description: User not authorized to access this endpoint
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+    post:
```
eloaverona: Should a notification be posted to a particular user? Like POST `/user/{user}/notification` instead of the general notification endpoint. Or maybe both options should exist.

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+      description: Send a list of notifications to the user module
```
eloaverona: `user module` is an internal thing, I don't think it should be mentioned in this doc. I don't think it would is clear for the consumers of this doc what `user module` is.

splinterd/api/static/openapi.yml
dnewh: /user -> user module
.../users -> users resource

I agree /users/{user}/... is cleaner but it doesn't follow the <module>/<resource>/<id> pattern  previously established in this file.

splinterd/api/static/openapi.yml
dnewh: I don't think so. A notification can be sent to a list of recipients

splinterd/api/static/openapi.yml
dnewh: updated

splinterd/api/static/openapi.yml
dnewh: updated

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+        500:
+          description: Internal service error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /user/users/{user_id}/notifications:
+    get:
+      description: Retrieve a list of notifications for a given user from the notifications database
```
dplumb94: I don't think that the user-specific notification should include the full list of recipients.

splinterd/api/static/openapi.yml
```diff
@@ -469,3 +608,34 @@ components:
         prev: /nodes?offset=0&limit=10
         next: /nodes?offset=20&limit=10
         last: /nodes?offset=40&limit=10
+
+    Notification:
```
dplumb94: How do we know if the notification is read or not? If that functionality is part of the user resource, that should be added in this design.

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+            type: string
+      responses:
+        200:
+          description: Notifications have been retrieved
+          content:
+            application/json:
+              schema:
+                type: array
+                $ref: '#/components/schemas/Notification'
+        403:
```
dplumb94: What determines of the user is authorized?

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,145 @@ paths:
...
+          description: User not authorized to retrieve notifications for this user
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        404:
```
dplumb94: Add 500 error and any other errors that could result from these calls

splinterd/api/static/openapi.yml
dplumb94: Could we call it userModule instead? Or something...

splinterd/api/static/openapi.yml
dnewh: Is there any reason that it shouldn't?

splinterd/api/static/openapi.yml
dnewh: If they A.) Have some admin permissions or B.) have the same user ID ad the requested resource. Not entirely sure on how that gets added to the request. Suggestions?

splinterd/api/static/openapi.yml
dnewh: updated

splinterd/api/static/openapi.yml
dnewh: updated

splinterd/api/static/openapi.yml
dplumb94: Previously we talked about a bearer token

splinterd/api/static/openapi.yml
```diff
@@ -327,6 +327,163 @@ paths:
...
+          description: notifications for {user} not found
```
dplumb94: I think this is actually that the user itself is not found, since a list request would just return `[]` if there are no resources found

splinterd/api/static/openapi.yml
dnewh: So I guess "unauthorized" might be misleading in that description. 403 is forbidden. So a user could be authorized (i.e. logged in) but forbidden from accessing that resource (i.e have incorrect permissions).

splinterd/api/static/openapi.yml
```diff
@@ -469,3 +626,36 @@ components:
         prev: /nodes?offset=0&limit=10
         next: /nodes?offset=20&limit=10
         last: /nodes?offset=40&limit=10
+
+    Notification:
+      type: object
+      properties:
+        hash:
+          type: string
+          example: 1c9b512cef6d6a9064d0cf0eb4b9a957af98bfcec3eb4be2bfb047cf73051cfcca8fa1c142ee12c419d36c55c391b7df38751dae64a1238dfe1e319e9ff22113
```
dplumb94: line wrapping (I think you can use folded style: https://symfony.com/doc/current/components/yaml/yaml_format.html#strings)

splinterd/api/static/openapi.yml
dplumb94: It seems problematic to expose all of the uids to all requesters even if they arent authorized to view all notifications. That information is also not useful to Canopy in general when displaying notifications.

splinterd/api/static/openapi.yml
dplumb94: In any case, we should add the token documentation to this spec and apply it to the impacted routes.

splinterd/api/static/openapi.yml
dnewh: /accounts/users...?
/users/accounts...?

splinterd/api/static/openapi.yml
dnewh: that adds a space for every line break. The other key strings in this file simply exceed the wrap limit

splinterd/api/static/openapi.yml
dnewh: updated

splinterd/api/static/openapi.yml
dnewh: I don't believe we would put that in this spec. That would be part of the headers of the request, none of which are specified in any of the existing endpoints described in this file.

splinterd/api/static/openapi.yml
dnewh: updated

PR #328, opened 2019-11-11
Add features for 'admin-service' and 'scabbard-service'

vaporos:
This also adds 'stable' and 'experimental' in order to run the builds.

admin-service and scabbard-service are off be default as they may move outside libsplinter at some point (as services/ libraries like the health service).

### DIFF
--- a/tests/test-splinter.yaml
+++ b/tests/test-splinter.yaml
@@ -26,10 +26,18 @@ services:
       - ../:/project/splinter
     command: |
         bash -c "
+            echo \"Running workspace tests...\" && \
             cargo test && \
-            echo \"Running optional feature tests...\" && \
-            (cd libsplinter && cargo test --features zmq-transport) &&
-            cargo test signing::sawtooth --manifest-path libsplinter/Cargo.toml \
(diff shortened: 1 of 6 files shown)


tests/test-splinter.yaml
```diff
@@ -26,10 +26,18 @@ services:
...
     command: |
         bash -c "
+            echo \"Running workspace tests...\" && \
             cargo test && \
-            echo \"Running optional feature tests...\" && \
-            (cd libsplinter && cargo test --features zmq-transport) &&
-            cargo test signing::sawtooth --manifest-path libsplinter/Cargo.toml \
-                --features \"sawtooth-signing-compat\"
+            echo \"Running features=default tests...\" && \
+            cargo test \
```
agunde406: This will run all the tests, even if they are not affected by the feature (all splinterd tests, when we are only changing libsplinter for example) Is this desirable?

vaporos:
Closing due to travis double-build.

tests/test-splinter.yaml
agunde406: I was actually wrong about this, ignore

PR #329, opened 2019-11-11
Add basic Canopy notification implementation

dplumb94:
Adds a redux store with a `readNotification` action and some mock data. Once the user module is implemented, this will get replaced with async actions that will call splinterd REST endpoints.

Adds a NotificationList component, NotificationListItem component, and mplements some basic styles for the new components as well as the layout. Once we can import the design system styles, these styles will likely be replaced.

To test:

1. `cd canopy/app`
2. `yarn install`
3. `yarn start`
4. Click the notification bell
5. Click the notifications to read them

Note: reloading the page (including navigating to the sapling) will reload the mock notifications, clearing the read status. This won't happen when the notifications are being fetched/pushed from a database.


canopy/app/src/index.js
```diff
@@ -16,7 +16,25 @@
...
 import './index.css';
-import App from './App';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const store = createStore(
+  rootReducer,
+
+  // This enables the use of redux devtools in the browser
+  // eslint-disable-next-line no-underscore-dangle
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
zapplebee: This should probably be behind a macro so it wouldn't get compiled into a production build.

canopy/app/src/reducers/notifications.js
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { READ_NOTIFICATION } from 'actions';
+
+const mockNotifications = [
+  {
```
zapplebee: Not sure if we want hard coded, even example notifications in here.

canopy/app/src/App.scss
```diff
@@ -20,8 +20,45 @@
...
   .view {
+    position: fixed;
+    left: 10rem;
     flex: 0 1 100%;
     min-height: 100vh;
   }
 }
+
+.display-flex {
+  display: flex;
```
zapplebee: Single declarations shouldn't be made into classes.

There is no reason not to include these as style declarations in the necessary components.

canopy/app/src/components/notifications/NotificationListItem.js
```diff
@@ -0,0 +1,56 @@
...
+    <div
+      role="button"
+      tabIndex={0}
+      onClick={() => dispatch(readNotification(notification.id))}
+      onKeyPress={e => {
+        if (e.key === 'Enter') {
+          dispatch(readNotification(notification.id));
+        }
+      }}
+      className={`notification-list-item ${
```
zapplebee: Consider the dependency [classnames](https://www.npmjs.com/package/classnames) for clearer composition of conditional classes

canopy/app/src/components/notifications/NotificationListItem.js
```diff
@@ -0,0 +1,56 @@
...
+import 'components/notifications/NotificationListItem.scss';
+
+function NotificationListItem({ notification }) {
+  const dispatch = useDispatch();
+
+  return (
+    <div
+      role="button"
+      tabIndex={0}
+      onClick={() => dispatch(readNotification(notification.id))}
```
zapplebee: This could be deduped to make the implementation easier to change.

zapplebee:
Also, want to hear the justification for using Redux here.

We can certainly pull it in, but [React's Context](https://reactjs.org/docs/context.html) may provide the needed functionality.

PR #330, opened 2019-11-12
Event dealer updates

ltseeley:


libsplinter/src/rest_api/events.rs
```diff
@@ -61,17 +61,16 @@ impl<T: Serialize + Debug + Clone + 'static> EventDealer<T> {
     }
 
     /// Send message to all created WebSockets
-    pub fn dispatch(&mut self, msg: T) -> Result<(), EventDealerError> {
+    pub fn dispatch(&mut self, msg: T) {
         self.senders.retain(|sender| {
             if let Err(err) = sender.unbounded_send(MessageWrapper::Message(msg.clone())) {
-                warn!("Dropping sender due to error: {}", err);
+                error!("Dropping sender due to error: {}", err);
```
peterschwarz: I'm not sure if these should be logged as an error in the long run: they may create a bunch of error noise at shutdown.

libsplinter/src/rest_api/events.rs
```diff
@@ -60,14 +61,14 @@ impl<T: Serialize + Debug + Clone + 'static> EventDealer<T> {
...
+    /// Send event to all created WebSockets.
+    pub fn dispatch(&mut self, event: T) {
         self.senders.retain(|sender| {
-            if let Err(err) = sender.unbounded_send(MessageWrapper::Message(msg.clone())) {
+            if let Err(err) = sender.unbounded_send(MessageWrapper::Message(event.clone())) {
                 error!("Dropping sender due to error: {}", err);
                 false
             } else {
-                trace!("Message sent: {:?}", msg);
+                trace!("Message sent: {:?}", event);
```
peterschwarz: This log message will be duplicated for every sender, but without any information of who received - we should only log this once.

libsplinter/src/rest_api/events.rs
ltseeley: Reverted to warn

libsplinter/src/rest_api/events.rs
ltseeley: Moved log statement to the end of the method

PR #335, opened 2019-11-12
Add option to save generated certs to a directory

agunde406:
This is a temporary feature request until certificate generation is moved to the splinter cli


agunde406:
This will be replaced by a different solution

PR #338, opened 2019-11-13
Store scabbard transaction receipts

ltseeley:
Adds the `TransactionReceiptStore` from the `sawtooth` crate to
`ScabbardState` for storing transaction receipts that are committed. The
receipt store provides an absolute history of all transactions that were
committed by the scabbard service, which can be used to "catch-up"
state-delta subscribers that may be behind.

### DIFF
--- a/libsplinter/src/service/scabbard/factory.rs
+++ b/libsplinter/src/service/scabbard/factory.rs
@@ -32,11 +32,15 @@ use super::{Scabbard, SERVICE_TYPE};
 
 const DEFAULT_STATE_DB_DIR: &str = "/var/lib/splinter";
 const DEFAULT_STATE_DB_SIZE: usize = 1028 * 1028 * 1028;
+const DEFAULT_RECEIPT_DB_DIR: &str = "/var/lib/splinter";
+const DEFAULT_RECEIPT_DB_SIZE: usize = 1 << 30; // 1024 ** 3
 
 pub struct ScabbardFactory {
     service_types: Vec<String>,
     state_db_dir: String,
     state_db_size: usize,
+    receipt_db_dir: String,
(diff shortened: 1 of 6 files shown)


libsplinter/src/service/scabbard/factory.rs
```diff
@@ -30,26 +30,34 @@ use crate::signing::SignatureVerifierFactory;
 
 use super::{Scabbard, SERVICE_TYPE};
 
-const DEFAULT_DB_DIR: &str = "/var/lib/splinter";
-const DEFAULT_DB_SIZE: usize = 1028 * 1028 * 1028;
+const DEFAULT_STATE_DB_DIR: &str = "/var/lib/splinter";
+const DEFAULT_STATE_DB_SIZE: usize = 1028 * 1028 * 1028;
```
jsmitchell: Why are these 1028? (I realize this predates this commit)

ltseeley: Just the precedent set by Sawtooth.

ltseeley: Uhhh yeah that should actually be 1024. No idea why it's 1028.

ltseeley: Fixed

splinterd/src/daemon.rs
```diff
@@ -289,6 +289,8 @@ impl SplinterDaemon {
                 })?;
         let orchestrator = ServiceOrchestrator::new(
             vec![Box::new(ScabbardFactory::new(
+                None,
```
peterschwarz: These options feel like they should probably be a single config object: reading this is pretty heavy on the `None`'s

PR #340, opened 2019-11-13
Scabbard state delta catch up

ltseeley:
Updates the scabbard service to catch-up a subscriber if it has missed
any state change events. When the subscriber sends the WebSocket
subscription request, it can specify the ID of the last state change
event it received; if this is specified, any state changes that occurred
after the specified one will be sent to the subscriber. If no ID is
specified with the subscription request, all state change events will be
sent (scabbard assumes that this is a new subscriber that will need to
be caught-up).

Also updates gameroomd to specify the last event it saw when it resubscribes.

### DIFF
--- a/libsplinter/src/service/scabbard/state.rs
+++ b/libsplinter/src/service/scabbard/state.rs
@@ -31,7 +31,7 @@ use transact::scheduler::{
 };
 use transact::state::{
     merkle::{MerkleRadixTree, MerkleState, INDEXES},
-    StateChange, Write,
+    StateChange as TransactStateChange, Write,
 };
 use transact::{
     execution::{adapter::static_adapter::StaticExecutionAdapter, executor::Executor},
@@ -99,7 +99,7 @@ impl ScabbardState {
                     err
                 ))
             })?;
-            let admin_keys_state_change = StateChange::Set {
+            let admin_keys_state_change = TransactStateChange::Set {
                 key: ADMINISTRATORS_SETTING_ADDRESS.into(),
                 value: admin_keys_setting_bytes,
             };
@@ -221,7 +221,7 @@ impl ScabbardState {
         // Save the results and compute the resulting state root
         let state_root = MerkleState::new(self.db.clone()).compute_state_id(
             &self.current_state_root,
-            &receipts_into_state_changes(&txn_results),
+            &receipts_into_transact_state_changes(&txn_results),
         )?;
         self.pending_changes = Some((signature.to_string(), txn_results));
         Ok(state_root)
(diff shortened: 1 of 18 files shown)


libsplinter/src/service/scabbard/state.rs
```diff
@@ -284,22 +275,67 @@ impl ScabbardState {
...
+                "failed to get transaction receipts from store: {}",
+                err
+            ))
+        })?;
+
+        let unseen_receipts = if let Some(txn_id) = last_txn_receipt {
+            all_receipts
+                .skip_while(|receipt| receipt.transaction_id != txn_id)
+                .skip(1)
+                .collect::<Vec<_>>()
```
jsmitchell: won't this potentially load an unbounded amount of stuff into memory?

ltseeley: Potentially, yes. Couple things to note, though: 1) the `TransactionReceiptStore.iter()` method is already doing that, because it returns an iterator over owned items. 2) there are plans to "age-out" old transaction receipts when they aren't necessary to reproduce the current state, so this will be bounded by the amount of addresses that are used by the smart contracts.

ltseeley:
To test this PR:
1) Startup gameroom, create a gameroom, and create a game
2) Shutdown bubba's gameroomd: `docker-compose -f examples/gameroom/docker-compose.yaml rm -f -s gameroomd-bubba`
3) Make a move for Alice using the acme UI
4) Restart bubba's gameroomd: `docker-compose -f examples/gameroom/docker-compose.yaml up -d  gameroomd-bubba`
5) Refresh bubba's UI and verify that Alice's move appears

libsplinter/src/service/scabbard/state.rs
peterschwarz: Improvements could be made by adding a `range_iter` method, that can be translated by the underlying store implementation into more efficient query.  Those changes would need to be made in `libsawtooth` first.

Either way, you could avoid the `collect` call entirely and just use the iterators.

peterschwarz: @jsmitchell Iterators are lazy.

ltseeley: I think we could use the iterators directly here. The ordered store implementations in libsawtooth do load all items into memory as they're currently implemented, but that could be changed in the future.

peterschwarz: That should probably be changed pretty quickly...

ltseeley: Updated the "Send past state deltas to scabbard subscriber" commit to use the iterators directly.

vaporos:
Let's fix the libsawtooth implementation to avoid the memory issue and consider writing some tests with ridiculously large receipt stores to make sure we have it correct.

peterschwarz:
That iterator is passed into a function that dumps them into a async channel.  We should look and see if we can send a Stream via the channel...

jsmitchell:
Approving as @peterschwarz has signed a blood oath to fix this in a imminent PR.

PR #342, opened 2019-11-14
splinterd add/update/delete node endpoints

ltseeley:
Add the following routes the splinterd REST API:
- POST /nodes for adding nodes to the local registry
- DELETE /nodes/{identity} for deleting nodes from the local registry
- PATCH /nodes/{identity} for updating existing nodes in the local registry

Also moves node registry implementations to libsplinter (will be put behind features(s) in the future), removes the NodeRegistryManager from splinterd, and changes the node registry's create_node to add_node to more accurately reflect the functionality.


ltseeley:
> This is missing the openapi documentation for the new routes

Added commit to update the OpenAPI spec

splinterd/src/daemon.rs
```diff
@@ -358,6 +358,7 @@ impl SplinterDaemon {
                 routes::get_status(node_id.clone(), service_endpoint.clone())
             }))
             .add_resource(routes::make_fetch_node_resource(node_registry.clone()))
+            .add_resource(routes::make_update_node_resource(node_registry.clone()))
```
agunde406: Typo in commit message "PATH /nodes/{identity}" -> "PATCH  /nodes/{identity}"

ltseeley: Fixed

ltseeley:
Added two commits to fix how REST API resources are defined and how the YamlNodeRegistry writes changes to its file.

libsplinter/src/rest_api/mod.rs
```diff
@@ -32,7 +32,7 @@
 //!     fn resources(&self) -> Vec<Resource> {
 //!         let name = self.name.clone();
 //!
-//!         vec![Resource::new(Method::Get, "/index", move |r, p| {
```
peterschwarz: I would suggest that we deprecate the old `new` and make the new way `build`

ltseeley: Done

PR #343, opened 2019-11-15
CanopyJS user implementation

zapplebee:
A Sapling can now use the CanopyJS functions to set and get a user from Canopy.

That user data is stored in a React Context and is exposed via the `window.$CANOPY` object.

The reason that the refactor of `promiseLoader` is here is that React was barking about setting one of its refs to a function from a cross origin script that was not explicitly stated on script load.

The deletion of the `canopyState` object is intended to bring as much of that state into the React application as possible. Some variable names have been updated to better reflect their intent because of this.

### DIFF
--- a/canopy/app/src/App.js
+++ b/canopy/app/src/App.js
@@ -14,70 +14,57 @@
  * limitations under the License.
  */
 
-import React, { useRef, useLayoutEffect, useState } from 'react';
+import React, { useRef, useState, useEffect } from 'react';
+import { useUserState, UserProvider } from 'UserStore';
 import { loadAllSaplings } from './loadSaplings';
 import 'App.scss';
 
-const canopyState = {
-  configSaplings: {},
-  bootstrapApp: null
-};
-
-function invokeConfigSaplings() {
-  const configSaplings = Object.values(canopyState.configSaplings);
-  if (configSaplings.length === 0) {
-    throw new Error('No Config Saplings registered');
-  }
-  configSaplings.forEach(bootstrapConfigSapling => {
-    bootstrapConfigSapling();
-  });
-}
-
-function invokeRegisteredApp(domNode) {
-  if (canopyState.registeredApp === null) {
-    throw new Error('No Sapling registered');
-  }
-  canopyState.bootstrapApp(domNode);
-}
-
-window.$CANOPY = {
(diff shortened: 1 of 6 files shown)


canopy/app/src/App.js
```diff
@@ -14,80 +14,74 @@
  * limitations under the License.
  */
 
-import React, { useRef, useLayoutEffect, useState } from 'react';
+import React, { useRef, useState, useEffect } from 'react';
 import { loadAllSaplings } from './loadSaplings';
+import { useUserState, UserProvider } from './UserStore';
```
dplumb94: You can just say ` import { useUserState, UserProvider } from 'UserStore'; ` here because of the absolute import config

zapplebee: ✅

canopy/canopyjs/src/index.spec.ts
```diff
@@ -58,4 +64,30 @@ describe('CanopyJS', () => {
       expect($CANOPY.registerConfigSapling.mock.calls[0][1]).toEqual(bootstrap);
     });
   });
+
+  describe('getUser()', () => {
+    it('it should call getUser from window object', async () => {
```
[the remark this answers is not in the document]
zapplebee: ✅

canopy/canopyjs/src/index.ts
```diff
@@ -13,6 +13,20 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+interface User {
+  userId: string;
+  displayName?: string;
```
eloaverona: In the REST API side, currently there is no concept of a display name. I imagine that would be included when we have user preference's implemented? So, I don't know if it makes sense to include this here at this time.

peterschwarz: It could be part of a the client codes transform from the rest api to this structure.  Presumably the display name would be in some sort of arbitrary properties.

zapplebee: So yes, `displayName` is an optional argument and it is not meant to be a 1-to-1 with the data that is exposed via the API. The structure of the user object on the front-end should be as flat as possible and this is meant to illustrate that. Perhaps a comment would be nice, but I don't see it as a necessity.

PR #344, opened 2019-11-15
Add certificate generation subcommand to splinter cli

agunde406:
This subcommand will generate certificates that
can be used for development. They should never be used
in production.

There are options to change where the certs and keys will
be generated. If this is not provided, an environment
variable, SPLINTER_CERT_DIR, will be checked. Other
wise the command will try to write the certs/keys to
/etc/splinter/certs/.

If the files already exist the command will
fail unless --force is provided. You can also use
--skip to see if the files exist or not. --quiet will not
print any information.

This also updates splinterd to check expected locations for the certs, instead of requiring that all are passed in the config or on the command line.

The current --generate-certs option to splitnerd is put behind a feature and deprecated. The feature will remain in default until after the next release.  It will then be removed from default but left as a feature until the following release. At this time the capability will be removed from splinterd all together, forcing the use of the splinter cli command.

### DIFF
--- a/cli/src/action/certs.rs
+++ b/cli/src/action/certs.rs
@@ -25,6 +25,7 @@ use std::os::linux::fs::MetadataExt;
 use std::os::unix::fs::MetadataExt;
 
 use clap::ArgMatches;
+use flexi_logger::ReconfigurationHandle;
 use openssl::asn1::Asn1Time;
 use openssl::bn::{BigNum, MsbOption};
 use openssl::error::ErrorStack;
@@ -51,7 +52,11 @@ const CA_CERT: &str = "generated_ca.pem";
 const CA_KEY: &str = "generated_ca.key";
 
 impl Action for CertGenAction {
-    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+    fn run<'a>(
+        &mut self,
+        arg_matches: Option<&ArgMatches<'a>>,
+        _logger_handle: &ReconfigurationHandle,
+    ) -> Result<(), CliError> {
         let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
 
         let common_name = args
--- a/cli/src/action/certs.rs
+++ b/cli/src/action/certs.rs
@@ -1,0 +1,734 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::env;
+use std::fs::{self, metadata, OpenOptions};
+use std::io;
+use std::io::Write;
+use std::os::unix::fs::OpenOptionsExt;
+use std::path::{Path, PathBuf};
+
+#[cfg(target_os = "linux")]
+use std::os::linux::fs::MetadataExt;
+#[cfg(not(target_os = "linux"))]
+use std::os::unix::fs::MetadataExt;
+
+use clap::ArgMatches;
+use openssl::asn1::Asn1Time;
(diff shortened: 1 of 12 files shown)


cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+        let server_key_path = private_cert_dir.join(SERVER_KEY);
+
+        // if check
+        if args.is_present("skip") {
+            if client_cert_path.exists()
+                && client_key_path.exists()
+                && server_cert_path.exists()
+                && server_key_path.exists()
+            {
+                println!("Certificates and keys exist");
```
vaporos: This should print out more information -- maybe something like "Certificate and key exists, skipping: {}, {}" on two separate lines.

Did we decide that these have to be done together?

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+            } else if client_cert_path.exists()
+                || client_key_path.exists()
+                || server_cert_path.exists()
+                || server_key_path.exists()
+            {
+                // this will always be an error condition
+                return Err(CliError::ActionError(
+                    "Some but not all certificates and keys already exist".into(),
```
vaporos: This should be more specific. If I was trying to debug this, I would need to know the paths that this tool was using and which existed and which did not. (Applies to similar messages in this file.)

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+                ));
+            } else {
+                println!("Certificates and keys do not exist");
```
vaporos: This should be more explicit about what does not exist (paths, etc.).

Do these really get turned off if you give -q? Maybe consider doing these with logging and configure the logger at the top-level, instead of passing a quiet boolean through the code.

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+        private_cert_path.clone(),
+        CLIENT_KEY,
+        &client_key.private_key_to_pem_pkcs8()?,
+        change_permissions,
+    )?;
+
+    if !quiet {
+        if forcing {
+            println!(
+                "overwote file: {:?}",
```
vaporos: overwote typo here and below

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    write_file(
+        private_cert_path.clone(),
+        CLIENT_KEY,
+        &client_key.private_key_to_pem_pkcs8()?,
+        change_permissions,
+    )?;
+
+    if !quiet {
+        if forcing {
```
vaporos: You could refactor this a bit and put a "File will be overwritten (forced): {}" in the calling function and remove force boolean from this function.

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+    };
+
+    let mut file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o640)
+        .open(temp_path.clone())?;
+    file.write_all(bytes)?;
+    fs::rename(temp_path, final_path)?;
```
vaporos: we could add a feature here which, if final_path exists at this point, renames it to final_path.bkN (where N increments until the file doesn't exist). We would want to log "renaming existing final_path to final_path.bkN".

cli/src/action/certs.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    if change_permissions {
```
vaporos: This should happen prior to the atomic rename above.

cli/src/action/certs.rs
agunde406: We did not make a decisions so I went with the first suggested solution that it should be all or nothing.

agunde406: will update so each pair can be created on its own

cli/src/action/certs.rs
agunde406: Created a story for this feature

cli/src/action/certs.rs
agunde406: done

cli/src/action/certs.rs
agunde406: done

cli/src/action/certs.rs
agunde406: done

cli/src/action/certs.rs
agunde406: fixed

cli/src/action/certs.rs
agunde406: fixed

PR #345, opened 2019-11-18
Add navigation sidebar

dplumb94:
Adds a sidebar for navigation, currently pulling in any user-specified saplings as tabs.

### DIFF
--- a/canopy/app/src/App.js
+++ b/canopy/app/src/App.js
@@ -15,8 +15,11 @@
  */
 
 import React, { useRef, useState, useEffect } from 'react';
+
 import { useUserState, UserProvider } from 'UserStore';
+import SideNav from 'components/navigation/SideNav';
 import { loadAllSaplings } from './loadSaplings';
+
 import 'App.scss';
 
 window.$CANOPY = {};
@@ -66,19 +69,21 @@ function App() {
     })();
   }, []);
 
+  const userSaplingRoutes = userSaplingManifests.map(
+    ({ displayName, namespace, icon }) => {
+      return {
+        path: `/${namespace}`,
+        displayName,
+        logo: icon
+      };
+    }
+  );
+
   return (
-    <>
-      <nav>
-        {userSaplingManifests.map(({ displayName, namespace }) => {
-          return (
-            <a href={`/${namespace}`} key={namespace}>
-              {displayName}
-            </a>
-          );
-        })}
-      </nav>
-      <div ref={saplingDomNode} />
-    </>
+    <div className="app">
+      <SideNav userSaplingRoutes={userSaplingRoutes} />
+      <div className="view" ref={saplingDomNode} />
+    </div>
   );
 }
 
--- a/canopy/app/src/components/navigation/SideNav.scss
+++ b/canopy/app/src/components/navigation/SideNav.scss
@@ -1,0 +1,98 @@
+/**
+ * Copyright 2019 Cargill Incorporated
+ *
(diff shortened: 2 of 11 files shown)


canopy/app/src/App.js
```diff
@@ -61,19 +64,21 @@ function App() {
...
-              {displayName}
-            </a>
-          );
-        })}
-      </nav>
-      <div ref={saplingDomNode} />
-    </>
+    <div className="app">
+      <SideNav userSaplingRoutes={userSaplingRoutes} />
+      <div className="view background-light" ref={saplingDomNode} />
```
dnewh: Maybe switch from using the background-light class to using the scss variable for consistency

canopy/app/src/components/navigation/SideNav.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.side-nav {
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  top: 0;
+  left: 0;
+  width: fit-content;
+  height: 100vh;
+  background: #ffffff;
```
zapplebee: The sidebar blips before rendering the Nav Items. You could fix this with a min-width.

`min-width: ($padding-s * 3) + 2rem + 9rem;`

This isn't the most elegant solution because the nav-item styles are in a different file.

Really, just adding those Nav Item definitions to the same files as the SideNav classes would clean this up. And it would be totally fine to do so because NavItems are always a child of SideNav.

If you did that, you could have something much easier to keep in sync with a SCSS map
```
$nav-tab-spacings: (
  padding: $padding-s,
  column-gap: $padding-s,
  icon-width: 2rem,
  label-width: 9rem
);
$total-nav-width: map-get($nav-tab-spacings, 'padding') * 2 +
  map-get($nav-tab-spacings, 'column-gap') +
  map-get($nav-tab-spacings, 'icon-width') +
  map-get($nav-tab-spacings, 'label-width');

.side-nav {
  display: flex;
  flex-direction: column;
  position: relative;
  top: 0;
  left: 0;
  width: fit-content;
  height: 100vh;
  background: #ffffff;
  min-width: $total-nav-width;

  hr {
    border: 0;
    width: 100%;
    background-color: $background-light;
    height: 2px;
  }
}

.brand {
  display: flex;
  flex-direction: column;
  align-items: center;
  text-decoration: none;
  color: darken($color-canopy, 20%);
  padding: $padding-s;
  color: text-primary;

  h5 {
    margin: 0;
  }
}

.nav-tab {
  display: grid;
  grid-template-columns:
    [icon-start] map-get(
      $nav-tab-spacings,
      'icon-width'
    ) [icon-end label-start] map-get($nav-tab-spacings, 'label-width')
    [label-end];
  column-gap: map-get($nav-tab-spacings, 'column-gap');
  align-items: center;
  padding: map-get($nav-tab-spacings, 'padding');
  text-overflow: ellipsis;
  overflow: hidden;
  text-decoration: none;
  color: $text-primary;
  transition: all 0.2s linear;

  .icon {
    height: map-get($nav-tab-spacings, 'icon-width');
    width: map-get($nav-tab-spacings, 'icon-width');
  }
```

zapplebee: Other than that, LGTM

dplumb94: Nice tip! Updated

zapplebee:
LGTM

PR #351, opened 2019-11-19
Allow cors from http-server for Canopy App

zapplebee:
Without CORS allowed, Canopy cannot load the scripts correctly. This was left out of the [previous PR](https://github.com/Cargill/splinter/pull/343).


zapplebee:
> This is just CORS in the development context, correct?

Yeah. The start script in `package.json` starts an http server for the Saplings to prove that they are separate from the built source of the the Canopy app.

PR #355, opened 2019-11-21
Clean-up authorization state after peer disconnect

peterschwarz:
This commit adds a DisconnectListener to the Network struct, which will be called when ever a peer is detected as disconnected.

The auth manager listens for disconnects, in order to clean up any auth state that may still exist, in order for connections to be re-connected.

### DIFF
--- a/libsplinter/src/network/auth/mod.rs
+++ b/libsplinter/src/network/auth/mod.rs
@@ -16,7 +16,10 @@ pub mod handlers;
 
 use std::collections::HashMap;
 use std::fmt;
-use std::sync::{Arc, Mutex};
+use std::sync::{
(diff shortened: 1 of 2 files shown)


libsplinter/src/network/auth/mod.rs
```diff
@@ -472,6 +501,59 @@ mod tests {
         );
     }
 
+    /// This test verifies that a connection that is authorized, if has disconnected, can begin the
+    /// authorization process over again.
+    /// authorized again.
```
agunde406: remove this line

peterschwarz: Fixed

PR #361, opened 2019-11-25
Subscribe to admin auth events with catch-up params.

peterschwarz:


libsplinter/src/admin/service/mailbox.rs
```diff
@@ -241,7 +237,72 @@ mod tests {
...
-                .iter()
+                .iter_since(SystemTime::UNIX_EPOCH)
+                .expect("Unable to create an iterator")
+                .map(|(_, evt)| evt)
+                .collect::<Vec<_>>(),
+        );
+    }
+
+    #[test]
+    fn test_iterate_empty() {
```
agunde406: Please add descriptions to these tests

libsplinter/src/events/reactor.rs
```diff
@@ -17,8 +17,9 @@ use std::sync::{
     Arc,
 };
 use std::thread;
+use std::time::Duration;
```
agunde406: typo in commit message: interupts -> interrupts

peterschwarz: Fixed

libsplinter/src/admin/service/mailbox.rs
peterschwarz: Added

examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -81,8 +81,21 @@ pub fn run(
...
 
+    let since_param = helpers::get_last_updated_proposal_time(&pool)?
+        .map(|time| {
+            format!("?since={}", {
+                time.duration_since(SystemTime::UNIX_EPOCH)
+                    .map(|duration| duration.as_millis())
+                    .unwrap_or(0)
+            })
+        })
+        .unwrap_or_else(|| String::from(""));
```
ltseeley: Maybe the parameter should simply be left off when the auth handler doesn't have a last updated time, rather than providing an empty string (which isn't an integer as the API spec requires) for the param?

examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -54,6 +54,20 @@ const RECONNECT_LIMIT: u64 = 10;
 /// default timeout in seconds if no message is received from server
 const CONNECTION_TIMEOUT: u64 = 60;
 
+#[derive(Deserialize, Debug, Clone)]
+struct Event {
```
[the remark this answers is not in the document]
peterschwarz: It's just for deserializing the received event - it's a local, temporary struct

libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -81,26 +81,47 @@ fn make_application_handler_registration_route<A: AdminCommands + Clone + 'stati
...
         };
 
+        let (skip, last_seen_timestamp) = query
+            .remove("since")
+            .map(|since_millis| {
+                // Since this is the last seen event, we will skip it in our since
+                // query
+                debug!("Catching up on events since {}", since_millis);
+                (
+                    1usize,
```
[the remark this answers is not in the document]
peterschwarz: Well, `1` if we didn't send the since value.  Though, I wonder if the default should be "start from now" instead of "catch me up from all of time"

libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -154,7 +154,9 @@ impl AdminEventDealers {
         event_type: &str,
         initial_events: Box<dyn Iterator<Item = JsonAdminEvent> + Send>,
     ) -> Result<(EventSender<JsonAdminEvent>, Response), ResponseError> {
-        let mut event_dealers = self.event_dealers_by_type.lock().unwrap();
+        let mut event_dealers = self.event_dealers_by_type.lock().map_err(|_| {
+            ResponseError::InternalError("Admin event dealer lock was poisoned".into())
```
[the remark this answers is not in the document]
peterschwarz: Hmm, I'm not sure we've been very consistent on it, actually.  I'll change it though.

examples/gameroom/daemon/src/authorization_handler/mod.rs
peterschwarz: That is how the parameter is being left off. That produces either:
```
{
    "?since={some number}"
} else {
    ""
}
```

peterschwarz: It could return the full path in either block, I suppose.

ltseeley: Oops, misread this. Thanks for clarifying.

libsplinter/src/admin/rest_api/mod.rs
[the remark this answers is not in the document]
ltseeley: Might be nice to have both options for the app auth handler; it's easy to specify the all-time scenarios by just providing `since=0`.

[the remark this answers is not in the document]
peterschwarz: maybe the parameter name is wrong - perhaps `last` to have the last timestamp of the event received, instead of since.

peterschwarz:
Renamed the `since` parameter to `last`

examples/gameroom/database/src/helpers/gameroom.rs
```diff
@@ -116,6 +116,15 @@ pub fn gameroom_service_is_active(conn: &PgConnection, circuit_id: &str) -> Quer
         .or_else(|err| if err == NotFound { Ok(false) } else { Err(err) })
 }
 
+pub fn get_last_updated_proposal_time(conn: &PgConnection) -> QueryResult<Option<SystemTime>> {
```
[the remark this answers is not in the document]
peterschwarz: Fixed

libsplinter/src/admin/rest_api/mod.rs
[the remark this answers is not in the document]
peterschwarz: Fixed

PR #362, opened 2019-11-26
Add Canopy transaction submission functionality

dnewh:
Adds transaction submission functionality to canopyjs.
This includes:
- Transaction/Sabre transaction builder implementations
- Batch builder implementation
- Ecdsa Secp256k1 Signer implementation
- http request utility

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -54,6 +54,7 @@
 /canopy/**/coverage
 /canopy/**/build
 /canopy**/dist
+/canopy/**/node_modules
 
 /target
 /Cargo.lock
--- a/canopy/canopyjs/package.json
+++ b/canopy/canopyjs/package.json
@@ -5,12 +5,15 @@
   "author": "Cargill Incorporated",
   "license": "Apache-2.0",
   "main": "dist/index.js",
-  "dependencies": {},
+  "dependencies": {
+    "protobufjs": "^6.8.8"
+  },
   "scripts": {
     "test": "jest",
     "build": "tsc -p ./tsconfig.json --strict --outDir dist",
     "format": "prettier --write \"**/*.+(ts|*rc|json|js)\"",
-    "lint": "eslint src/*"
+    "lint": "eslint src/**/**/**/*",
+    "generate-proto-files": "node scripts/compile_protobuf.js ./protos"
   },
   "devDependencies": {
     "@types/jest": "^24.0.19",
--- a/canopy/canopyjs/package.json
+++ b/canopy/canopyjs/package.json
@@ -6,7 +6,9 @@
   "license": "Apache-2.0",
   "main": "dist/index.js",
   "dependencies": {
-    "protobufjs": "^6.8.8"
+    "@types/secp256k1": "^3.5.0",
+    "protobufjs": "^6.8.8",
+    "secp256k1": "^3.7.1"
   },
   "scripts": {
     "test": "jest",
--- a/canopy/canopyjs/package.json
+++ b/canopy/canopyjs/package.json
@@ -11,6 +11,7 @@
     "secp256k1": "^3.7.1"
   },
   "scripts": {
+    "postinstall": "yarn generate-proto-files",
     "test": "jest",
     "build": "tsc -p ./tsconfig.json --strict --outDir dist",
     "format": "prettier --write \"**/*.+(ts|*rc|json|js)\"",
--- a/canopy/canopyjs/scripts/compile_protobuf.js
+++ b/canopy/canopyjs/scripts/compile_protobuf.js
@@ -1,0 +1,56 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+const fs = require('fs');
+const pbjs = require('protobufjs/cli/pbjs');
+const pbts = require('protobufjs/cli/pbts');
+const process = require('process');
+
+const path = require('path');
+
+const protoDir = process.argv[2];
+
+const files = fs.readdirSync(protoDir).map(f => path.resolve(protoDir, f));
+
+pbjs.main(
+  [
+    '-t',
+    'static-module',
+    '-w',
+    'commonjs',
+    '-o',
+    'src/compiled_protos.js',
+    ...files
+  ],
+  function handleJsCompileError(err) {
+    if (err) {
+      console.error(`Error compiling protobuf bundle: ${err}`);
+    } else {
+      console.log('Sucessfully created protobuf bundle');
+    }
+  }
+);
+
+pbts.main(
+  ['-o', 'src/compiled_protos.d.ts', 'src/compiled_protos.js'],
+  function handleJsCompileError(err) {
+    if (err) {
+      console.error(`Error compiling protobuf typings: ${err}`);
+    } else {
+      console.log('Sucessfully created protobuf typings');
+    }
+  }
+);
--- a/canopy/canopyjs/src/index.ts
+++ b/canopy/canopyjs/src/index.ts
@@ -13,7 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+export {
+  TransactionBuilder,
+  SabreTransactionBuilder
+} from './transactions/transaction';
 interface User {
   userId: string;
   displayName?: string;
--- a/canopy/canopyjs/src/index.ts
+++ b/canopy/canopyjs/src/index.ts
@@ -17,6 +17,7 @@ export {
   TransactionBuilder,
   SabreTransactionBuilder
 } from './transactions/transaction';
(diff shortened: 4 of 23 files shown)


canopy/canopyjs/src/transactions/submitter.ts
```diff
@@ -0,0 +1,81 @@
...
+    request.onload = () => {
+      if (request.status >= 200 && request.status < 300) {
+        resolve(request.response);
+      } else {
+        console.error(request);
+        if (request.status >= 400 && request.status < 500) {
+          reject('Failed to send request. Contact the administrator for help.');
+        } else {
+          reject(
+            'The Gameroom server has encountered an error. Please contact the administrator.'
```
zapplebee: Copy/paste artifact

canopy/canopyjs/src/transactions/submitter.ts
```diff
@@ -0,0 +1,81 @@
...
+interface BatchInfo {
+  id: string;
+  status: BatchStatus;
+}
+
+async function http(
+  method: string,
+  url: string,
+  data: Uint8Array,
+  headerFn: (request: XMLHttpRequest) => void
```
zapplebee: This should probably have a default implementation, no?

canopy/canopyjs/src/transactions/submitter.ts
```diff
@@ -0,0 +1,81 @@
...
+  error_message: string;
+  error_data: number[];
+}
+
+interface BatchInfo {
+  id: string;
+  status: BatchStatus;
+}
+
+async function http(
```
zapplebee: What is the benefit to writing our own http abstraction in place of using something like Fetch or Axios?

canopy/canopyjs/src/transactions/batch/errors/index.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
zapplebee: MissingFieldError can probably be defined here to reduce the import nesting.

canopy/canopyjs/src/protobuf/index.ts
```diff
@@ -0,0 +1,25 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import protobuf, { Root } from 'protobufjs';
+import protoJSON from '../compiled_protos.json';
+
+const root: Root = protobuf.Root.fromJSON(protoJSON);
+
+export default Object.keys(root)
```
zapplebee: Is this actually needed since the `compiled_protos.js` file gets generated?

canopy/canopyjs/scripts/compile_protobuf.js
```diff
@@ -0,0 +1,65 @@
...
+      console.log('Sucessfully created protobuf bundle');
+    }
+  }
+);
+
+pbts.main(
+  ['-o', 'src/compiled_protos.d.ts', 'src/compiled_protos.js'],
+  function(err) {
+    err
+      ? console.error(`Error compiling protobuf typings: ${err}`)
```
zapplebee: This error handling should be consistent. Sorry, I know this was a PR I had into your branch.

canopy/canopyjs/package.json
```diff
@@ -5,12 +5,17 @@
...
+    "protobufjs": "^6.8.8",
+    "secp256k1": "^3.7.1"
+  },
   "scripts": {
     "test": "jest",
     "build": "tsc -p ./tsconfig.json --strict --outDir dist",
     "format": "prettier --write \"**/*.+(ts|*rc|json|js)\"",
-    "lint": "eslint src/*"
+    "lint": "eslint src/**/**/**/*",
+    "generate-proto-files": "node scripts/compile_protobuf.js ./protos"
```
zapplebee: Should probably add this as a `postinstall` script

canopy/canopyjs/src/transactions/addressing.ts
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const crypto = require('crypto');
+
+// The namespace registry prefix for global state (00ec00)
+const NAMESPACE_REGISTRY_PREFIX = '00ec00';
+
+// The contract prefix for global state (00ec02)
+const CONTRACT_PREFIX = '00ec02';
```
zapplebee: These magic strings should probably come from some kind of config file instead of constants living with business logic, no?

canopy/canopyjs/src/transactions/submitter.ts
dnewh: We ran into some hard-to-debug issues using other libraries. E.g. Axios just doesn't handle octet-stream right at all.

canopy/canopyjs/src/transactions/submitter.ts
dnewh: What headers would you suggest adding by default?

canopy/canopyjs/src/transactions/batch/errors/index.ts
dnewh: This will cause lint errors later on if any other error types get added

canopy/canopyjs/src/transactions/submitter.ts
dnewh: updated

canopy/canopyjs/scripts/compile_protobuf.js
dnewh: updated

canopy/canopyjs/src/transactions/addressing.ts
peterschwarz: It's a very sabre-specific constant, so it doesn't make sense to be in a config file. Any changes to these values would probably also require code changes here.

canopy/canopyjs/src/protobuf/index.ts
dnewh: no, it is not. Updated

canopy/canopyjs/package.json
dnewh: added

.dockerignore
```diff
@@ -54,6 +54,7 @@
 /canopy/**/coverage
 /canopy/**/build
 /canopy**/dist
+/canopy/**/node_modules
```
dplumb94: Isn't this already accomplished by line 50?

canopy/canopyjs/scripts/compile_protobuf.js
```diff
@@ -0,0 +1,56 @@
...
+const pbts = require('protobufjs/cli/pbts');
+const process = require('process');
+
+const path = require('path');
+
+const protoDir = process.argv[2];
+
+const files = fs.readdirSync(protoDir).map(f => path.resolve(protoDir, f));
+
+pbjs.main(
```
dplumb94: How come this is a different script than what we used in Gameroom? It looks like in that one we are generating JSON descriptors and reading them into a TS module, but this is a different pattern.

canopy/canopyjs/tsconfig.json
```diff
@@ -2,7 +2,9 @@
   "compilerOptions": {
     "lib": ["es2015", "dom"],
     "esModuleInterop": true,
-    "declaration": true
+    "declaration": true,
+    "resolveJsonModule": true,
+    "allowJs": true
```
dplumb94: does this mean we lose types for the protobuf library?

canopy/canopyjs/src/transactions/signing.ts
```diff
@@ -0,0 +1,163 @@
+/**
```
dplumb94: Is this a rewrite of the Sawtooth SDK signing in typescript?

canopy/canopyjs/src/transactions/signing.ts
```diff
@@ -0,0 +1,163 @@
...
+    .map(b => b.toString(16).padStart(2, '0'))
+    .join('');
+}
+
+interface IContext {
+  sign: (message: Uint8Array, privateKey: PrivateKey) => string;
+  getPublicKey: (privateKey: PrivateKey) => PublicKey;
+}
+
+interface IPrivateKey {
```
dplumb94: I'd recommend splitting the interfaces and implementations out into separate files, see the rust sdk for example: [Rust SDK](https://github.com/hyperledger/sawtooth-sdk-rust/tree/master/src/signing).

Traits would be interfaces and structs would be classes.

canopy/canopyjs/src/transactions/signing.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+class Context implements IContext {
+  sign(message: Uint8Array, privateKey: PrivateKey): string {
+    const hash = crypto
+      .createHash('sha512')
+      .update(message)
+      .digest();
+
+    const result = secp256k1.sign(hash, Buffer.from(privateKey.asBytes()));
```
dplumb94: Shouldn't have any secp256k1 stuff in the abstract implementation. Looks like this is the same code as the `Secp256k1Context`?

canopy/canopyjs/src/index.ts
```diff
@@ -17,6 +17,7 @@ export {
   TransactionBuilder,
   SabreTransactionBuilder
 } from './transactions/transaction';
+export { BatchBuilder } from './transactions/batch';
```
dplumb94: Spacing looks slightly off here.

How come these aren't exported in the canopy object?

.dockerignore
dnewh: I thought so, but I thought I was getting node_modules when I built

canopy/canopyjs/tsconfig.json
zapplebee: No since the types file is generated.

PR #363, opened 2019-11-26
Add shared config to canopy

dnewh:
Adds shared config to canopy. This can be used for configuration such as the splinter endpoint that should be used by various saplings. This config is intended to be set and accessed in a similar way to the userSaplings and loginRegister configurations but it should contain more general specifications.


zapplebee:
LGTM

PR #367, opened 2019-12-04
Biome login route

eloaverona:
This PR implements a login route for biome. When it a client sends a request to the route with the payload:
```json
{ 
    "username": "my_username",
    "hashed_password": "D3912FA5639A37367C15814367069FE7798CFAF6"
}
```
 - The route handler verifies that username exists
 - Verifies that the password provided is correct
 - Issues a JWT token that the client can use to access other protected routes and the server will use to verify that the users have the permission to access such routes
- Returns the user_id and jwt token in the payload:
```json 
{
    "message": "Successful login",
    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjoiYzlhNTM0MjEtZTUzOS00NDMyLTkyYjktMTAzNTI5M2FkZWVlIiwiaXNzIjoic2VsZi1pc3N1ZWQiLCJleHAiOjE1NzU0MTk4Mjh9.97n-gSCjBTE7lcVNhkDt6WNL3-x_Z8VLz5fLfn8uCH8",
    "user_id": "c9a53421-e539-4432-92b9-1035293adeee"
}
```

To support that this PR implements:
- An API to pass configuration options for BiomeRestResources, such as how long the JWT access tokens should last and what the issuer field of the token payload should be set to.  
- A basic Biome API to manage server side secrets 
- A basic Biome API to issue JWT tokens
- A method to verify JWT tokens

### DIFF
--- a/libsplinter/src/biome/credentials/mod.rs
+++ b/libsplinter/src/biome/credentials/mod.rs
@@ -21,17 +21,23 @@ pub mod database;
 mod error;
 pub(in crate::biome) mod rest_resources;
 
-use bcrypt::{hash, DEFAULT_COST};
+use bcrypt::{hash, verify, DEFAULT_COST};
 
 use database::models::{NewUserCredentialsModel, UserCredentialsModel};
 
-pub use error::{CredentialsStoreError, UserCredentialsBuilderError};
+pub use error::{CredentialsStoreError, UserCredentialsBuilderError, UserCredentialsError};
 
 /// Represents crendentials used to authenticate a user
 pub struct UserCredentials {
     user_id: String,
     username: String,
     password: String,
+}
+
+impl UserCredentials {
+    pub fn verify_password(&self, password: &str) -> Result<bool, UserCredentialsError> {
+        Ok(verify(password, &self.password)?)
+    }
 }
 
 /// Builder for UsersCredential. It hashes the password upon build.
--- a/libsplinter/src/biome/credentials/mod.rs
+++ b/libsplinter/src/biome/credentials/mod.rs
@@ -21,11 +21,16 @@ pub mod database;
 mod error;
 pub(in crate::biome) mod rest_resources;
 
+use std::convert::TryFrom;
+
 use bcrypt::{hash, verify, DEFAULT_COST};
 
 use database::models::{NewUserCredentialsModel, UserCredentialsModel};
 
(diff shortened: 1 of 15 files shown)


libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -62,6 +64,8 @@ pub struct BiomeRestResourceManager {
     // This is only used if the biome-credentials feature is enabled
     #[allow(dead_code)]
     user_store: Arc<SplinterUserStore>,
+    #[allow(dead_code)]
```
peterschwarz: What is the need for this?

eloaverona: For now, that field is only used when the biome-credentials feature is enabled. So you get a lint warning when that feature is not enabled.

libsplinter/src/biome/credentials/mod.rs
```diff
@@ -21,11 +21,11 @@ pub mod database;
 mod error;
 pub(in crate::biome) mod rest_resources;
 
-use bcrypt::{hash, DEFAULT_COST};
```
RyanLassigBanks: The default cost for bycrpt is pretty high and can cause login to appear to hang for a while, you might want to consider lowering it or making it configurable.

eloaverona: I added a commit to make the hash cost configurable

libsplinter/src/biome/rest_api/mod.rs
eloaverona: I added a comment on the dead_code warning. I thought it would be better not conditionally compile that because that field is one I think other features will use once they are implemented. And I did not want to imply that that field should only be used for credentials feature

libsplinter/src/biome/secrets/auto_secret_manager.rs
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use rand::{distributions::Alphanumeric, Rng};
+
+use super::{SecretManager, SecretManagerError};
+
+const SECRET_LENGHT: usize = 64;
```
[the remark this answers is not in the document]
eloaverona: Fixed

PR #368, opened 2019-12-04
Node registry validation

ltseeley:
- Update the definition of a `Node` in the node registry to have required `endpoint` and `display_name` fields
- Update the YAML node registry to verify that all nodes' required fields are not empty
- Update the YAML node registry to verify that all node identities and endpoints are unique
- Change `NodeRegistry.update_node` to take a new node definition rather than just updated fields

### DIFF
--- a/libsplinter/src/node_registry/error.rs
+++ b/libsplinter/src/node_registry/error.rs
@@ -70,6 +70,7 @@ pub enum InvalidNodeError {
     EmptyEndpoint,
     EmptyIdentity,
     EmptyDisplayName,
+    InvalidIdentity(String, String), // (identity, message)
 }
 
 impl Error for InvalidNodeError {
@@ -80,6 +81,7 @@ impl Error for InvalidNodeError {
             InvalidNodeError::EmptyEndpoint => None,
             InvalidNodeError::EmptyIdentity => None,
             InvalidNodeError::EmptyDisplayName => None,
+            InvalidNodeError::InvalidIdentity(..) => None,
         }
     }
 }
@@ -97,6 +99,9 @@ impl fmt::Display for InvalidNodeError {
             InvalidNodeError::EmptyIdentity => write!(f, "node must have non-empty identity"),
             InvalidNodeError::EmptyDisplayName => {
                 write!(f, "node must have non-empty display_name")
+            }
+            InvalidNodeError::InvalidIdentity(identity, msg) => {
+                write!(f, "identity {} is invalid: {}", identity, msg)
             }
         }
     }
--- a/libsplinter/src/node_registry/error.rs
+++ b/libsplinter/src/node_registry/error.rs
@@ -19,9 +19,7 @@ use std::fmt;
 pub enum NodeRegistryError {
     /// This error is returned when a node is not found.
     NotFoundError(String),
-    /// This error is returned when the user attempts to create a node with an identity that
-    /// already exists.
-    DuplicateNodeError(String),
+    InvalidNode(InvalidNodeError),
     /// This error is returned when an internal error occurred
     InternalError(Box<dyn Error + Send>),
     /// This error is returned when the user cannot create a node in the registry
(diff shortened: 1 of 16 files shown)


libsplinter/src/node_registry/error.rs
```diff
@@ -75,6 +75,7 @@ pub enum InvalidNodeError {
     EmptyEndpoint,
     EmptyIdentity,
     EmptyDisplayName,
+    InvalidIdentity(String, String), // (identity, msg)
```
peterschwarz: Should this be more general? Like `InvalidUpdate`?

ltseeley: This is general in the context of the `InvalidNodeError`. An `InvalidUpdate` variant would not be appropriate since this error type is used for non-update actions as well.

libsplinter/src/node_registry/mod.rs
```diff
@@ -137,12 +132,8 @@ where
         (**self).add_node(node)
     }
 
-    fn update_node(
-        &self,
-        identity: &str,
-        updates: HashMap<String, String>,
-    ) -> Result<(), NodeRegistryError> {
-        (**self).update_node(identity, updates)
+    fn update_node(&self, identity: &str, node: Node) -> Result<(), NodeRegistryError> {
```
peterschwarz: If you're passing the node to this API, do you still need to pass identity?  Then you wouldn't need the error for identity mismatch.

libsplinter/src/node_registry/mod.rs
```diff
@@ -21,9 +21,9 @@ use std::collections::HashMap;
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
 pub struct Node {
-    /// The Splinter identity of the node.
```
peterschwarz: As these are public fields, you should keep and update the doc comments, not remove them.

ltseeley: Fixed

libsplinter/src/node_registry/mod.rs
ltseeley: Updated to add/update nodes in the registry using a single `add_or_update_node` method. The REST endpoints have been updated accordingly, with the update endpoint returning an error if the identity of the node is changed.

libsplinter/src/node_registry/mod.rs
```diff
@@ -123,29 +123,13 @@ pub trait NodeRegistryReader: Send + Sync {
...
-    ///
-    ///  * `identity` - The Splinter identity of the node.
-    ///  * `updates` - A map containing the updated properties.
-    ///
-    fn update_node(
-        &self,
-        identity: &str,
-        updates: HashMap<String, String>,
-    ) -> Result<(), NodeRegistryError>;
+    fn add_or_update_node(&self, node: Node) -> Result<(), NodeRegistryError>;
```
peterschwarz: Why not just `set_node` or `insert_node` (`insert` being the name of this sort of method on `HashMap`, which does the equivalent)

libsplinter/src/node_registry/yaml/mod.rs
```diff
@@ -117,11 +116,14 @@ impl NodeRegistryReader for YamlNodeRegistry {
...
 
 impl NodeRegistryWriter for YamlNodeRegistry {
-    fn add_node(&self, node: Node) -> Result<(), NodeRegistryError> {
+    fn add_or_update_node(&self, node: Node) -> Result<(), NodeRegistryError> {
         let mut nodes = self
             .get_cached_nodes()
             .map_err(|err| NodeRegistryError::InternalError(Box::new(err)))?;
 
+        // If a node with the same identity already exists, remove it
+        nodes.retain(|existing_node| existing_node.identity != node.identity);
```
peterschwarz: Shouldn't this be done after the validation?

libsplinter/src/node_registry/mod.rs
ltseeley: `insert_node` is much better; updated to that.

libsplinter/src/node_registry/yaml/mod.rs
ltseeley: Good catch; fixed.

ltseeley: Actually, no; when checking for uniqueness of the node's endpoint, we don't want to be comparing the node against itself. Removing the node from this vec has no consequences if the method returns early, because it's just a cache that only gets written at the end of the method.

peterschwarz: You can at least validate that the incoming node is valid before removing the old node from the cache.

ltseeley: Moved the field validation before removing the node from the cache.

PR #371, opened 2019-12-05
Connection Manager enhancements

RyanLassigBanks:
* Replace crossbeam with standard library channels
* Remove locks around connection state
* Move the heartbeat monitor into its own module
* Add remove connection to connection manager

### DIFF
--- a/libsplinter/src/network/connection_manager/heartbeat_monitor.rs
+++ b/libsplinter/src/network/connection_manager/heartbeat_monitor.rs
@@ -1,0 +1,100 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    mpsc::SyncSender,
+    Arc,
+};
+use std::thread;
+use std::time::Duration;
+
+use crate::network::connection_manager::{error::ConnectionManagerError, messages::CmMessage};
+
+pub struct HeartbeatMonitor {
+    interval: u64,
+    join_handle: Option<thread::JoinHandle<()>>,
+    shutdown_handle: Option<HbShutdownHandle>,
+}
+
+impl HeartbeatMonitor {
+    pub fn new(interval: u64) -> Self {
+        Self {
+            interval,
+            join_handle: None,
+            shutdown_handle: None,
+        }
+    }
+
+    pub fn start(&mut self, cm_sender: SyncSender<CmMessage>) -> Result<(), ConnectionManagerError> {
+        if self.join_handle.is_some() {
+            return Ok(());
+        }
+
+        let running = Arc::new(AtomicBool::new(true));
+
+        let running_clone = running.clone();
+        let interval = self.interval;
+        let join_handle = thread::Builder::new()
+            .name("Heartbeat Monitor".into())
+            .spawn(move || {
+                info!("Starting heartbeat manager");
+
+                while running_clone.load(Ordering::SeqCst) {
+                    thread::sleep(Duration::from_secs(interval));
+                    if let Err(err) = cm_sender.send(CmMessage::SendHeartbeats) {
+                        error!(
+                            "Connection manager has disconnected before 
+                            shutting down heartbeat monitor {:?}",
+                            err
+                        );
+                        break;
+                    }
+                }
+            })?;
+
+        self.join_handle = Some(join_handle);
+        self.shutdown_handle = Some(HbShutdownHandle { running });
+
+        Ok(())
+    }
+
+    pub fn shutdown_handle(&self) -> Option<HbShutdownHandle> {
+        self.shutdown_handle.clone()
+    }
+
+    pub fn await_shutdown(self) {
+        let join_handle = if let Some(jh) = self.join_handle {
+            jh
+        } else {
+            return;
+        };
+
+        if let Err(err) = join_handle.join() {
+            error!("Failed to shutdown heartbeat monitor gracefully: {:?}", err);
+        }
+    }
+}
+
+#[derive(Clone)]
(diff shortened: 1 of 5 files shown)


libsplinter/src/network/connection_manager/heartbeat_monitor.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright 2019 Cargill Incorporated
```
vaporos: Any reason not to call this "heartbeat" for a shorter module name (without an underscore)?

libsplinter/src/network/connection_manager/messages.rs
```diff
@@ -21,7 +21,7 @@ pub enum CmMessage {
     Subscribe(String, Sender<Vec<CmNotification>>),
     UnSubscribe(String),
     Request(CmRequest),
-    HeartbeatNotifications(Vec<CmNotification>),
+    SendHeartbeats,
```
vaporos: Consider calling this "HeartbeatPulse". Which, I guess, might make the current HeartbeatMonitor a HeartbeatPacemaker or HeartbeatPulseGenerator.

Conceptually, this would make HeartbeatMonitor more passive -- it's not requesting heartbeats as much as providing the tempo/pulse at which they should be considered.

RyanLassigBanks: I think Pacemaker actually makes a lot of sense, but I think HeartbeatPacemaker is tad redundant. Instead of "HeartbeatPacemaker" what if we just called the struct "PaceMaker"?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -38,17 +38,16 @@ const DEFAULT_HEARTBEAT_INTERVAL: u64 = 10;
 
 pub struct ConnectionManager {
     hb_monitor: HeartbeatMonitor,
-    connection_state: ConnectionState,
+    connection_state: Option<ConnectionState>,
```
agunde406: Can you add the reasoning to the commit message

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -270,7 +270,7 @@ pub struct Connector {
 
 impl Connector {
     pub fn request_connection(&self, endpoint: &str) -> Result<CmResponse, ConnectionManagerError> {
-        let (sender, recv) = bounded(1);
+        let (sender, recv) = channel();
```
peterschwarz: I think `sync_channel` would be the more correct change, as that has a bound.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -477,6 +409,56 @@ fn notify_subscribers(
     }
 }
 
+fn send_heartbeats(
+    state: &mut ConnectionState,
+    subscribers: &mut HashMap<String, Sender<Vec<CmNotification>>>,
+) {
+    let heartbeat_message = match create_heartbeat() {
```
peterschwarz: I'm actually not convinced this can actually ever fail, so this may be a good place to have a panic.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -564,7 +564,7 @@ pub mod tests {
     #[test]
     fn test_heartbeat_notifications_raw_tcp() {
         let mut transport = Box::new(RawTransport::default());
-        let mut listener = transport.listen("tcp://localhost:8080").unwrap();
+        let mut listener = transport.listen("tcp://localhost:3030").unwrap();
```
peterschwarz: You could use `localhost:0` and it will generate a random port.  Take a look at the tests in `libsplinter/src/transport/raw.rs`

RyanLassigBanks: I'm not in love with the idea of having the port be random each run. Also this seems to cause several tests to fail and hang in OSX.

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: Changed over to sync_channel

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: made create_heartbeat panic

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -405,17 +398,14 @@ fn send_heartbeats(
     notify_subscribers(subscribers, notifications);
 }
 
-fn create_heartbeat() -> Result<Vec<u8>, ConnectionManagerError> {
-    let heartbeat = NetworkHeartbeat::new().write_to_bytes().map_err(|_| {
-        ConnectionManagerError::HeartbeatError("cannot create NetworkHeartbeat message".to_string())
-    })?;
+fn create_heartbeat() -> Vec<u8> {
+    let heartbeat = NetworkHeartbeat::new().write_to_bytes().unwrap();
```
vaporos: Why is panicking here desirable?

RyanLassigBanks: I believe @peterschwarz mentioned this a few times, but `write_to_bytes` only fails if the system completely runs out of memory.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -62,7 +61,7 @@ impl ConnectionManager {
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
         let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
-        let mut state = self.connection_state.clone();
+        let mut state = self.connection_state.take().unwrap();
```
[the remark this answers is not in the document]
RyanLassigBanks: Updated

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -62,7 +61,13 @@ impl ConnectionManager {
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
         let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
-        let mut state = self.connection_state.clone();
+        let mut state = if let Some(state) = self.connection_state.take() {
```
ltseeley: This `if let ... else ... ` could be simplified by using `self.connection_state.take().ok_or_else(|| Err(...))?`

libsplinter/src/network/connection_manager/mod.rs
ltseeley: This deviates from the way we handle these errors everywhere else in splinter, so I'm not sure I 100% agree with this change from a consistency perspective, but I won't block the PR because of that. However, this should be changed to an `expect` to provide an error message.

PR #372, opened 2019-12-05
Add Shared Config to CanopyJS

zapplebee:
While previously, shared config was made available to the Canopy app, it was not invoked.

This change brings the shared config into the Canopy app and exposes it to Saplings via CanopyJS

### DIFF
--- a/canopy/canopyjs/src/index.spec.ts
+++ b/canopy/canopyjs/src/index.spec.ts
@@ -27,7 +27,10 @@ const minimalUser = { userId: 'MINIMAL' };
   registerConfigSapling: jest.fn(),
   registerApp: jest.fn(),
   getUser: jest.fn(() => completeUser),
-  setUser: jest.fn()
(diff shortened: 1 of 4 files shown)


zapplebee:
> typo in commit message, "condif" -> "config"

✔️

canopy/canopyjs/src/index.spec.ts
```diff
@@ -90,4 +94,11 @@ describe('CanopyJS', () => {
       expect($CANOPY.setUser.mock.calls[0][0]).toEqual(minimalUser);
     });
   });
+
+  describe('getSharedConfig', () => {
+    it('should call getSharedConfig form window object', async () => {
```
jsmitchell: should "form" be "from"?

zapplebee: > should "form" be "from"?
✔️

PR #373, opened 2019-12-05
Node registry list simplification

peterschwarz:
This PR pushes the parsing of the predicate operations and the limit/offset parameters to the REST API layer.  This simplifies the API for the Node registry and removes the need for implementers of the traits to parse the predicate operations.

### DIFF
--- a/libsplinter/src/node_registry/mod.rs
+++ b/libsplinter/src/node_registry/mod.rs
@@ -28,24 +28,71 @@ pub struct Node {
     pub metadata: HashMap<String, String>,
 }
 
+/// A predicate on a key/value pair in a Node's metadata table.
+///
+/// Each variant is an operator, and supplies a tuple representing a key/value pair. It is applied
+/// by the comparison operator on the value found at the given key (the first item in the tuple)
+/// against the predicate's value (the second item in the tuple).
+///
+/// If the item is missing in a node's metadata table, the predicate returns false.
+#[derive(Clone)]
+pub enum MetadataPredicate {
+    /// Applies the `==` operator.
+    Eq(String, String),
+    /// Applies the `!=` operator.
+    Ne(String, String),
+    /// Applies the `>` operator.
+    Gt(String, String),
+    /// Applies the `>=` operator.
+    Ge(String, String),
+    /// Applies the `<` operator.
+    Lt(String, String),
+    /// Applies the `<=` operator.
+    Le(String, String),
+}
+
+impl MetadataPredicate {
+    /// Apply this predicate against a given node.
+    pub fn apply(&self, node: &Node) -> bool {
+        match self {
+            MetadataPredicate::Eq(key, val) => {
+                node.metadata.get(key).map(|v| v == val).unwrap_or(false)
+            }
+            MetadataPredicate::Ne(key, val) => {
+                node.metadata.get(key).map(|v| v != val).unwrap_or(false)
+            }
+            MetadataPredicate::Gt(key, val) => {
+                node.metadata.get(key).map(|v| v > val).unwrap_or(false)
+            }
+            MetadataPredicate::Ge(key, val) => {
+                node.metadata.get(key).map(|v| v >= val).unwrap_or(false)
+            }
+            MetadataPredicate::Lt(key, val) => {
+                node.metadata.get(key).map(|v| v < val).unwrap_or(false)
+            }
+            MetadataPredicate::Le(key, val) => {
+                node.metadata.get(key).map(|v| v <= val).unwrap_or(false)
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


libsplinter/src/node_registry/mod.rs
```diff
@@ -28,24 +28,65 @@ pub struct Node {
     pub metadata: HashMap<String, String>,
 }
 
+/// A predicate on a key/value pair in a Node's metadata table.
+///
+/// Each variant is an operator, and supplies a tuple representing a key/ value pair. It is applied
+/// by the comparison operator on the valued found at the given key (the first item in the tuple)
```
ltseeley: s/valued/value/

libsplinter/src/node_registry/mod.rs
```diff
@@ -28,24 +28,65 @@ pub struct Node {
     pub metadata: HashMap<String, String>,
 }
 
+/// A predicate on a key/value pair in a Node's metadata table.
+///
+/// Each variant is an operator, and supplies a tuple representing a key/ value pair. It is applied
```
ltseeley: Unnecessary space after /

libsplinter/src/node_registry/mod.rs
```diff
@@ -28,24 +28,65 @@ pub struct Node {
...
+/// by the comparison operator on the valued found at the given key (the first item in the tuple)
+/// against the predicate's value  (the second item in the tuple).
```
ltseeley: Extra space after "value"

libsplinter/src/node_registry/mod.rs
```diff
@@ -28,24 +28,65 @@ pub struct Node {
...
+///
+/// If the item is missing in a node's metadata table, the predicate returns false.
+pub enum MetadataPredicate {
+    /// Applies the `==` operator for
```
ltseeley: This comment should probably be removed

libsplinter/src/node_registry/mod.rs
```diff
@@ -28,24 +28,65 @@ pub struct Node {
...
+    Ne(String, String),
+    Gt(String, String),
+    Ge(String, String),
+    Lt(String, String),
+    Le(String, String),
+}
+
+impl MetadataPredicate {
+    /// Apply this predicate against a given node.
+    pub fn matches(&self, node: &Node) -> bool {
```
ltseeley: Would more sense semantically for this method to be part of the `Node` struct, so you'd have `node.matches(predicate)` rather than `predicate.matches(node)`.

peterschwarz: Possibly, but this method is more of a convenience for in-memory usage.  Other implementations will use the enum to generate SQL statements, for example.

peterschwarz: Perhaps the name of the function could be changed from `matches` to `apply`

libsplinter/src/node_registry/mod.rs
peterschwarz: Fixed.

libsplinter/src/node_registry/mod.rs
peterschwarz: Fixed

libsplinter/src/node_registry/mod.rs
peterschwarz: Fixed

libsplinter/src/node_registry/mod.rs
peterschwarz: Fixed.

libsplinter/src/node_registry/mod.rs
peterschwarz: Forgot to actually complete the doc strings for those.  Corrected.

peterschwarz:
Actually yes. This count essentially requires doing the query twice, and provides one way to indicate if there are more items. 

I think we should move away from that count and merely return a next link, if there are more items

agunde406:
Sounds fine to me. Then the gameroom test needs to be updated for that.

peterschwarz:
The alternative is to add a `count` method to the reader interface

peterschwarz:
@agunde406 added a count method.

libsplinter/src/node_registry/mod.rs
```diff
@@ -79,23 +79,26 @@ impl MetadataPredicate {
 
 /// Provides Node Registry read capabilities.
 pub trait NodeRegistryReader: Send + Sync {
-    /// Returns a list of nodes.
+    /// Returns an iterator over the nodes in the registry.
```
agunde406: Squash typo: fixup_iter after the signature should be removed.

peterschwarz: Fixed.

PR #377, opened 2019-12-11
Connection Manager Example

RyanLassigBanks:
Adds `cm-repl`, an example daemon using the connection manger and refactors notifications API

### Testing

1) Build with docker
```
docker-compose -f examples/cm-repl/docker-compose.yaml
```
2) Exec into one of the node containers
```
docker exec -it cm-repl_node-0_1 bash
```
3) Run connection manager client commands
```
cmc connection add tcp://node-0:3040
cmc connection list
cmc connection remove tcp://node-0:3040
```

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -20,7 +20,7 @@ use std::collections::HashMap;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     mpsc::{sync_channel, Receiver, SyncSender},
-    Arc, Mutex, MutexGuard,
+    Arc,
 };
 use std::thread;
 use std::time::Duration;
@@ -39,7 +39,7 @@ const CHANNEL_CAPACITY: usize = 15;
 
 pub struct ConnectionManager {
     hb_monitor: HeartbeatMonitor,
-    connection_state: Arc<Mutex<ConnectionState>>,
+    connection_state: ConnectionState,
     join_handle: Option<thread::JoinHandle<()>>,
     sender: Option<SyncSender<CmMessage>>,
     shutdown_handle: Option<ShutdownHandle>,
@@ -47,9 +47,9 @@ pub struct ConnectionManager {
 
 impl ConnectionManager {
     pub fn new(mesh: Mesh, transport: Box<dyn Transport + Send>) -> Self {
-        let connection_state = Arc::new(Mutex::new(ConnectionState::new(mesh, transport)));
+        let connection_state = ConnectionState::new(mesh, transport);
         let hb_monitor =
-            HeartbeatMonitor::new(connection_state.clone(), DEFAULT_HEARTBEAT_INTERVAL);
+            HeartbeatMonitor::new(DEFAULT_HEARTBEAT_INTERVAL);
 
         Self {
             hb_monitor,
@@ -62,7 +62,7 @@ impl ConnectionManager {
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
         let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
-        let connection_state = self.connection_state.clone();
+        let mut state = self.connection_state.clone();
 
         let join_handle = thread::Builder::new()
             .name("Connection Manager".into())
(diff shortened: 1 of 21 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -60,7 +61,7 @@ impl ConnectionManager {
     }
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
-        let (sender, recv) = unbounded();
+        let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
```
peterschwarz: Why the replacement of `unbounded` with `sync_channel`?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -62,7 +62,7 @@ impl ConnectionManager {
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
         let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
-        let connection_state = self.connection_state.clone();
+        let mut state = self.connection_state.clone();
```
peterschwarz: Is this state only used by the thread created below?  If so, it doesn't need to be a field.  It should just owned by the thread.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -571,7 +571,7 @@ pub mod tests {
     #[test]
     fn test_heartbeat_notifications_raw_tcp() {
         let mut transport = Box::new(RawTransport::default());
-        let mut listener = transport.listen("tcp://localhost:8080").unwrap();
+        let mut listener = transport.listen("tcp://localhost:3030").unwrap();
```
peterschwarz: These could be `:0` which will assign a random port (and not conflict).  See the transport tests as an example

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -62,7 +61,13 @@ impl ConnectionManager {
 
     pub fn start(&mut self) -> Result<Connector, ConnectionManagerError> {
         let (sender, recv) = sync_channel(CHANNEL_CAPACITY);
-        let mut state = self.connection_state.clone();
+        let mut state = if let Some(state) = self.connection_state.take() {
```
peterschwarz: See my previous comment about this state.  The fact that `self.join_handle.is_some()` already tells us that the connection manager has already been started.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -35,7 +35,7 @@ const DEFAULT_HEARTBEAT_INTERVAL: u64 = 10;
 const CHANNEL_CAPACITY: usize = 15;
 
 pub struct ConnectionManager {
-    hb_monitor: HeartbeatMonitor,
+    pacemaker: Pacemaker,
```
peterschwarz: 👍

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -350,11 +350,11 @@ fn handle_request(req: CmRequest, state: &mut ConnectionState) {
 }
 
 fn notify_subscribers(
-    subscribers: &mut HashMap<String, SyncSender<Vec<CmNotification>>>,
-    notifications: Vec<CmNotification>,
+    subscribers: &mut HashMap<String, SyncSender<CmNotification>>,
+    notification: CmNotification,
 ) {
     for (id, sender) in subscribers.clone() {
```
peterschwarz: Couldn't say which is more efficient here: using `clone`, and remove if it fails or using `retain`, like we have in other situations.  Probably the later, since it only creates a new vector, and doesn't clone the contents.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -211,10 +211,11 @@ pub struct NotificationHandler {
 }
 
 impl NotificationHandler {
-    pub fn listen(&self) -> Result<CmNotification, ConnectionManagerError> {
-        match self.recv.recv() {
-            Ok(notifications) => Ok(notifications),
-            Err(_) => Err(ConnectionManagerError::SendMessageError(
+    pub fn try_next(&self) -> Result<Option<CmNotification>, ConnectionManagerError> {
```
peterschwarz: What's the justification for this?  There's not information in the commit message explaining why this is the case.  

It seems like it may just introduce busy-wait situation for the user.  Shouldn't there still be the blocking version of this method?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -231,6 +231,20 @@ impl NotificationHandler {
...
 }
 
+impl Iterator for NotificationHandler {
+    type Item = CmNotification;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.recv.recv() {
+            Ok(notification) => Some(notification),
+            Err(_) => {
+                error!("connection manager no longer running");
```
peterschwarz: Is this really an error?  This message will be displayed on a clean shutdown.  Probably should be debug message.

peterschwarz:
Could you add more to the commit messages? There is very little context around each change. 

Likewise, this PR seems like it's covering a number of separate topics, which will make it very hard to merge in a timely fashion.  I suggest you break it up into several smallery PR's

PR #379, opened 2019-12-12
Add sapling style loader

dnewh:
Adds a style loader to load external sapling styles into the canopy
DOM.

### DIFF
--- a/canopy/app/src/loadSaplings.js
+++ b/canopy/app/src/loadSaplings.js
@@ -15,6 +15,7 @@
  */
 
 import promiseLoader from 'promiseLoader';
+import styleLoader from 'styleLoader';
 import { getUserSaplings, getConfigSaplings } from './getSaplings.macro';
 
 async function loadCurrentSapling(userSaplingsResponse) {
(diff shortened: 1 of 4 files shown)


canopy/app/src/loadSaplings.js
```diff
@@ -74,11 +90,13 @@ export async function loadAllSaplings() {
     getConfigSaplings()
   ]);
 
-  const configSaplingsAreLoaded = await loadConfigSaplings(
-    configSaplingResponse
-  );
+  const configSaplingsAreLoaded =
+    (await loadConfigSaplings(configSaplingResponse)) &&
```
zapplebee: This doesn't maximize concurrency.

You could `await Promise.all([loadConfigSaplings(configSaplingResponse), loadSaplingStyles(configSaplingResponse)])

Otherwise they'll happen one after the other with await.

dnewh: updated

canopy/app/src/loadSaplings.js
```diff
@@ -48,6 +49,21 @@ async function loadCurrentSapling(userSaplingsResponse) {
   return false;
 }
 
+async function loadSaplingStyles(saplingResponse) {
+  const saplingStyleFiles = saplingResponse
+    .map(s => s.styleFiles)
```
peterschwarz: Could you give this and the following arguments a better name? Its very difficult to figure out what we're mapping over or flatMapping

dnewh: updated

PR #381, opened 2019-12-13
Notifications api

RyanLassigBanks:
Updates notifications api for connection manager.

### DIFF
--- a/libsplinter/src/network/connection_manager/messages.rs
+++ b/libsplinter/src/network/connection_manager/messages.rs
@@ -18,7 +18,7 @@ use crate::network::connection_manager::error::ConnectionManagerError;
 
 pub enum CmMessage {
     Shutdown,
-    Subscribe(String, SyncSender<Vec<CmNotification>>),
+    Subscribe(String, SyncSender<CmNotification>),
     UnSubscribe(String),
     Request(CmRequest),
     SendHeartbeats,
--- a/libsplinter/src/network/connection_manager/messages.rs
+++ b/libsplinter/src/network/connection_manager/messages.rs
@@ -18,7 +18,7 @@ use crate::network::connection_manager::error::ConnectionManagerError;
 
 pub enum CmMessage {
     Shutdown,
-    Subscribe(SyncSender<CmNotification>),
+    Subscribe(SyncSender<ConnectionManagerNotification>),
     Request(CmRequest),
     SendHeartbeats,
 }
@@ -56,7 +56,7 @@ pub enum CmResponseStatus {
 /// Messages that will be dispatched to all
 /// subscription handlers
 #[derive(Debug, PartialEq, Clone)]
-pub enum CmNotification {
+pub enum ConnectionManagerNotification {
     FatalError {
         error: ConnectionManagerError,
         message: String,
--- a/libsplinter/src/network/connection_manager/messages.rs
+++ b/libsplinter/src/network/connection_manager/messages.rs
@@ -18,8 +18,7 @@ use crate::network::connection_manager::error::ConnectionManagerError;
 
 pub enum CmMessage {
     Shutdown,
-    Subscribe(String, SyncSender<CmNotification>),
-    UnSubscribe(String),
+    Subscribe(SyncSender<CmNotification>),
     Request(CmRequest),
     SendHeartbeats,
 }
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -203,11 +203,11 @@ impl ShutdownHandle {
 pub struct NotificationHandler {
     id: String,
     sender: SyncSender<CmMessage>,
-    recv: Receiver<Vec<CmNotification>>,
+    recv: Receiver<CmNotification>,
 }
 
 impl NotificationHandler {
-    pub fn listen(&self) -> Result<Vec<CmNotification>, ConnectionManagerError> {
+    pub fn listen(&self) -> Result<CmNotification, ConnectionManagerError> {
         match self.recv.recv() {
             Ok(notifications) => Ok(notifications),
             Err(_) => Err(ConnectionManagerError::SendMessageError(
@@ -346,11 +346,11 @@ fn handle_request(req: CmRequest, state: &mut ConnectionState) {
 }
 
 fn notify_subscribers(
-    subscribers: &mut HashMap<String, SyncSender<Vec<CmNotification>>>,
-    notifications: Vec<CmNotification>,
+    subscribers: &mut HashMap<String, SyncSender<CmNotification>>,
+    notification: CmNotification,
 ) {
     for (id, sender) in subscribers.clone() {
-        if sender.send(notifications.clone()).is_err() {
+        if let Err(_) = sender.send(notification.clone()) {
             warn!("subscriber has dropped its connection to connection manager");
             subscribers.remove(&id);
         }
@@ -359,7 +359,7 @@ fn notify_subscribers(
 
 fn send_heartbeats(
     state: &mut ConnectionState,
-    subscribers: &mut HashMap<String, SyncSender<Vec<CmNotification>>>,
+    subscribers: &mut HashMap<String, SyncSender<CmNotification>>,
 ) {
     let heartbeat_message = match create_heartbeat() {
         Ok(h) => h,
@@ -368,7 +368,6 @@ fn send_heartbeats(
             return;
         }
     };
-    let mut notifications = Vec::new();
 
     for (endpoint, metadata) in state.connection_metadata() {
         info!("Sending heartbeat to {}", endpoint);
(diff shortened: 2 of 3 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -207,10 +207,11 @@ pub struct NotificationHandler {
 }
 
 impl NotificationHandler {
-    pub fn listen(&self) -> Result<CmNotification, ConnectionManagerError> {
-        match self.recv.recv() {
-            Ok(notifications) => Ok(notifications),
-            Err(_) => Err(ConnectionManagerError::SendMessageError(
+    pub fn try_next(&self) -> Result<Option<CmNotification>, ConnectionManagerError> {
+        match self.recv.try_recv() {
```
peterschwarz: What's the justification for this? There's not information in the commit message explaining why this is the case.

It seems like it may just introduce busy-wait situation for the user. Shouldn't there still be the blocking version of this method?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -227,6 +227,20 @@ impl NotificationHandler {
...
 }
 
+impl Iterator for NotificationHandler {
+    type Item = CmNotification;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.recv.recv() {
+            Ok(notification) => Some(notification),
+            Err(_) => {
+                error!("connection manager no longer running");
```
peterschwarz: Is this really an error? This message will be displayed on a clean shutdown. Probably should be debug message.

libsplinter/src/network/connection_manager/messages.rs
```diff
@@ -18,8 +18,7 @@ use crate::network::connection_manager::error::ConnectionManagerError;
 
 pub enum CmMessage {
     Shutdown,
-    Subscribe(String, SyncSender<CmNotification>),
```
peterschwarz: Could you add some context to the commit message? I'm not sure I understand why this change is being made.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -227,6 +227,20 @@ impl NotificationHandler {
     }
 }
 
+impl Iterator for NotificationHandler {
```
peterschwarz: I think the more appropriate trait for this use case is to implement `IntoIterator`.  See https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#impl-IntoIterator.  It should be pretty simple, as you can just return the Receiver.

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: Updated commit message

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -227,6 +227,20 @@ impl NotificationHandler {
...
+    type Item = CmNotification;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.recv.recv() {
+            Ok(notification) => Some(notification),
+            Err(_) => {
+                warn!("connection manager no longer running");
```
vaporos: Need a more accurate error message here, and we shouldn't throw away the root cause.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -686,4 +700,50 @@ pub mod tests {
...
+    /// creates a thread that send HeartbeatSent notifications to
+    /// the notifier.
+    ///
+    /// Asserts:
+    ///
+    /// The notifications sent are received by the Notifier
+    /// correctly
+    ///
+    /// That the total number of notifications sent equals 5
+    fn test_nofication_handler_iterator() {
```
vaporos: Typo - nofication

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -361,29 +340,27 @@ fn handle_request(req: CmRequest, state: &mut ConnectionState) {
...
 fn notify_subscribers(
-    subscribers: &mut HashMap<String, SyncSender<CmNotification>>,
+    subscribers: &mut Vec<SyncSender<CmNotification>>,
     notification: CmNotification,
 ) {
-    for (id, sender) in subscribers.clone() {
-        if let Err(_) = sender.send(notification.clone()) {
+    subscribers.retain(|sender| {
+        if sender.send(notification.clone()).is_err() {
             warn!("subscriber has dropped its connection to connection manager");
```
vaporos: Do we need some sort of id for this to be useful? How would we use this message when problem solving?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -227,6 +227,20 @@ impl NotificationHandler {
     }
 }
 
+impl Iterator for NotificationHandler {
+    type Item = CmNotification;
```
vaporos: This should probably become ConnectionManagerNotification or just Notification as part of the public API (Cm being for private implementation).

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: I don't think this is the correct spot to log this as an error condition.  The only error returned is that the channel disconnected.  That could happen due to a clean shutdown or an error condition of the sending end, neither of which can be properly debugged by a message here.  This should be a debug message (as previously suggested), so that we know the receiver has exited, but any error condition should be caught elsewhere.

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: This is now hidden behind a feature

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: Could you elaborate a bit? I'm sure why this would be more appropriate.

libsplinter/src/network/connection_manager/messages.rs
RyanLassigBanks: Updated commit message. Unsubscribe is unnecessary because you'd be using notifier like an iterator and calling unsubscribe on an iterator makes not sense, since it'll stop iterating when the sender goes out of scope.

libsplinter/src/network/connection_manager/mod.rs
RyanLassigBanks: Hmm maybe this shouldn't be `warn!` since it's completely valid for the subscribe to drop out of scope.

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: I guess it's a question of whether or not a NotificationHandler should _be_ an iterator or can be transformed _into_ an iterator.

RyanLassigBanks: I think it makes more sense as an Iterator, since it doesn't have much purpose otherwise, and having it implement `IntoIterator` instead just seems to add an extra step (calling `notifier.next()` vs calling `notifier.into_iter().next()`.

peterschwarz: I'm not sure I agree with that usage, though.  By hiding the blocking version behind `Iterator::next`, it's not exactly documented.  It should be its own method on Notifier, and the iterator provide just calls it.

RyanLassigBanks: Ok. So you're saying don't remove `listen`? Because what you're describing is what `listen` was. Should we even have an iterator then?

peterschwarz: We can still have an iterator, but `listen` should be part of the public API.  Iterator is still a nice convenience, so it can be both.

RyanLassigBanks: Changed `try_next` to `try_listen`, and add listen back.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -216,6 +218,19 @@ impl NotificationHandler {
         }
     }
 
+    #[cfg(feature = "connection-manager-notifier-try-next")]
```
vaporos: The name of this feature doesn't match the function (try_listen) and try_listen doesn't make as much as much sense as calling it try_next.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -216,17 +209,34 @@ impl NotificationHandler {
...
     }
 }
 
+impl Iterator for Notifier {
+    type Item = ConnectionManagerNotification;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.recv.recv() {
+            Ok(notification) => Some(notification),
+            Err(_) => {
```
vaporos: I think the Notifier should explicitly recv a end-of-iteration marker so that if we get an err on receive it is unexpected and means there was a coding error.

RyanLassigBanks: I'm not entirely sure how this would work, or if it's necessary. Again, `recv` can only "fail" if the connection manger has gone away either by shutting down gracefully or by panicking. Regardless, from the perspective of the iterator, it's done iterating. Adding an explicit marker seems redundant since the "marker" is the sender end going out of scope.

It's worth noting there is a `FatalError` notification that gets sent out if the connection manager is forced to shutdown unexpectedly, so the subscriber would be notified if something happened to connection manager and would be able to act accordingly.

vaporos: Ok, suggest removing the debug! then (completely uninteresting from a debugging standpoint, as you can't tell if it is normal operation or not) and adding a comment on that line reading "This is expected if the other side shuts down before this end." (re-used from Transact code).

PR #382, opened 2019-12-13
Organize features

ltseeley:
Standardize `[features]` sections of `Cargo.toml`s and update splinter tests to test all features.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -78,7 +78,6 @@ experimental = [
     "biome-notifications",
     "connection-manager",
     "database",
-    "ursa-compat",
     "zmq-transport",
 ]
 
@@ -89,8 +88,13 @@ connection-manager = []
 database = ["diesel", "diesel_migrations"]
 events = ["hyper", "tokio", "awc"]
 sawtooth-signing-compat = ["sawtooth-sdk"]
-ursa-compat = ["ursa"]
 zmq-transport = ["zmq"]
+
+# The following features are broken and should not be used.
+
+# The `ursa` dependency is currently broken, and will need to be fixed before
+# the `ursa-compat` feature can be added back to `experimental`.
+ursa-compat = ["ursa"]
 
 [package.metadata.docs.rs]
 features = [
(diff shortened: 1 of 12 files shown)


splinterd/Cargo.toml
```diff
@@ -45,16 +45,18 @@ toml = "0.4.8"
 [features]
 default = ["generate-certs"]
 
+stable = ["default"]
+
 experimental = [
+    "health",
```
vaporos: The order should probably be config-builder, config-toml, circuit-read, health.

ltseeley: Oops, fixed it so health is in the correct spot.

services/health/Cargo.toml
```diff
@@ -12,4 +12,6 @@ splinter = { path = "../../libsplinter" }
 [features]
 default = []
 
+stable = ["default"]
```
peterschwarz: Do we know how this sort of arrangement will work with `--no-default-features --features stable`?

libsplinter/Cargo.toml
```diff
@@ -66,23 +66,32 @@ glob = "0.2"
...
-sawtooth-signing-compat = ["sawtooth-sdk"]
-events = ["hyper", "tokio", "awc"]
 biome = ["database", "jsonwebtoken", "rand"]
 biome-credentials = ["biome", "database", "bcrypt"]
 biome-notifications = ["biome", "database"]
+connection-manager = []
 database = ["diesel", "diesel_migrations"]
+events = ["hyper", "tokio", "awc"]
+sawtooth-signing-compat = ["sawtooth-sdk"]
+# The `ursa` dependency is currently broken, and will need to be fixed before
```
vaporos: Suggest moving this to its own section/list, and making it clear that it is a list of broken features.

services/health/Cargo.toml
ltseeley: Works fine in my tests

libsplinter/Cargo.toml
ltseeley: Updated

PR #385, opened 2019-12-16
Add component build to design system

dnewh:
Adds commonJS, UMD, and ES component bundles to the design system build. This allows the components to be imported and used from the module.

### DIFF
--- a/canopy/design-system/package.json
+++ b/canopy/design-system/package.json
@@ -2,14 +2,24 @@
   "name": "canopy-design-system",
   "version": "0.1.0",
   "private": true,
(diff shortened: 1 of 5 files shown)


canopy/design-system/package.json
```diff
@@ -2,14 +2,24 @@
   "name": "canopy-design-system",
   "version": "0.1.0",
   "private": true,
+  "main": "lib/components.min.js",
```
dplumb94: Can you add some info to the readme on how to use these scripts?

dnewh: updated

PR #387, opened 2019-12-17
Move splinterd REST API routes

ltseeley:
Moves most of the REST API routes from splinterd into libsplinter so they are available to future daemons other than splinterd.

Also puts the REST API code in libsplinter behind a new `rest-api` feature makes its dependencies optional. This removes the need to compile in actix when the REST API isn't needed.


ltseeley:
Note: this PR does not move the OpenAPI specs for the REST endpoints into libsplinter; those remain in splinterd for now. Moving those will require a tool for assembling a single spec from multiple endpoint specs, which splinter doesn't currently have. A story has been backlogged to cover this.

PR #390, opened 2019-12-18
Add node registry commands to splinter CLI

ltseeley:
Add the `nodes` subcommand to the splinter CLI that has its own
subcommands:
- `list` for listing all nodes in the registry
- `get` for getting a specific node from the registry
- `add` for adding nodes to the registry from a file
- `update` for updating nodes in the registry from a file
- `remove` for removing nodes from the registry

The `list` feature does not currently allow for filtering nodes.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -23,9 +23,11 @@ experimental = [
     "database-migrate-biome-credentials",
     "database-migrate-biome-notifications",
     "database-migrate-biome-users",
+    "nodes",
 ]
 
 health = ["reqwest", "serde_json"]
+nodes = ["reqwest", "serde_json"]
 
 database = ["splinter/database", "diesel"]
 database-migrate-biome-users = ["splinter/biome", "database"]
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -30,6 +30,9 @@ use log::Record;
 const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
+#[cfg(feature = "health")]
+const DEFAULT_SPLINTER_NODE_URL: &str = "http://localhost:8085";
+
 // log format for cli that will only show the log message
 pub fn log_format(
     w: &mut dyn std::io::Write,
@@ -108,6 +111,7 @@ fn run() -> Result<(), CliError> {
                             Arg::with_name("url")
                                 .short("U")
                                 .takes_value(true)
+                                .default_value(DEFAULT_SPLINTER_NODE_URL)
                                 .help("URL of node"),
                         ),
                 ),
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -30,7 +30,7 @@ use log::Record;
 const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-#[cfg(feature = "health")]
+#[cfg(any(feature = "health", feature = "nodes"))]
 const DEFAULT_SPLINTER_NODE_URL: &str = "http://localhost:8085";
 
 // log format for cli that will only show the log message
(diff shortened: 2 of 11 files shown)


cli/src/main.rs
```diff
@@ -106,12 +109,103 @@ fn run() -> Result<(), CliError> {
...
+                            Arg::with_name(nodes::FORMAT_ARG)
+                                .long("format")
+                                .takes_value(true)
+                                .default_value(nodes::YAML)
+                                .possible_values(nodes::SUPPORTED_FORMATS)
+                                .help("Format nodes will be displayed in"),
+                        ),
+                )
+                .subcommand(
+                    SubCommand::with_name("get")
```
vaporos: We should call this 'show'.

cli/src/main.rs
```diff
@@ -106,12 +109,103 @@ fn run() -> Result<(), CliError> {
...
         );
     }
 
+    #[cfg(feature = "node-registry")]
+    {
+        use action::nodes;
+        use clap::{Arg, SubCommand};
+
+        app = app.subcommand(
+            SubCommand::with_name("nodes")
```
vaporos: Should this be non-plural 'node' instead? (Or maybe 'registry'.) There is potentially another single-word for 'node registry' as a concept.

cli/src/main.rs
```diff
@@ -106,12 +109,103 @@ fn run() -> Result<(), CliError> {
...
+                            Arg::with_name(nodes::FORMAT_ARG)
+                                .long("format")
+                                .takes_value(true)
+                                .default_value(nodes::YAML)
+                                .possible_values(nodes::SUPPORTED_FORMATS)
+                                .help("Format nodes will be displayed in"),
+                        ),
+                )
+                .subcommand(
+                    SubCommand::with_name("add")
```
vaporos: Could we get away from the file-based approach here? Or, better, support both using a file ("import") and a arg-based approach "add". If we added "import" we should also be able to "export".

cli/src/main.rs
```diff
@@ -106,12 +109,103 @@ fn run() -> Result<(), CliError> {
...
+                        .about("Add nodes to the registry, read from a JSON or YAML file")
+                        .arg(
+                            Arg::with_name(nodes::FILE_ARG)
+                                .takes_value(true)
+                                .required(true)
+                                .help("Path to JSON or YAML file with node definition(s)"),
+                        ),
+                )
+                .subcommand(
+                    SubCommand::with_name("update")
```
vaporos: If we had "import" then this should just become part of that import command.

cli/src/main.rs
```diff
@@ -106,12 +109,103 @@ fn run() -> Result<(), CliError> {
...
+                    Arg::with_name(nodes::URL_ARG)
+                        .long("url")
+                        .short("U")
+                        .global(true)
+                        .takes_value(true)
+                        .default_value(DEFAULT_SPLINTER_NODE_URL)
+                        .help("Splinter node's endpoint"),
+                )
+                .subcommand(
+                    SubCommand::with_name("list")
```
vaporos: List should show a table-based view by default and support csv output.

cli/src/main.rs
ltseeley: What would the arg-based "add" command look like? I'm not sure how useful that would be; it seems like it would be pretty clunky because of the number of arguments required (ID, endpoint, display name, metadata). Maybe @rbuysse and @desktophero would have some input on this.

cli/src/main.rs
ltseeley: How would arbitrary metadata be represented in a table or csv?

cli/src/main.rs
ltseeley: I think "nodes" is better than "node" because it's more clear. "registry" might be okay, but I don't know if it's descriptive enough (what if we add another kind of registry?). What about "node-registry"?

cli/Cargo.toml
```diff
@@ -17,9 +17,10 @@ default = []
 
 stable = ["default"]
 
-experimental = ["health"]
+experimental = ["health", "node-registry"]
```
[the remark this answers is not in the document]
ltseeley: Changed the feature name to match the subcommand name (`nodes`, for now at least).

cli/src/main.rs
ltseeley: Updated to `show`

libsplinter/src/node_registry/mod.rs
```diff
@@ -169,6 +169,13 @@ pub trait NodeRegistryWriter: Send + Sync {
     ///  * `identity` - The Splinter identity of the node.
     ///
     fn delete_node(&self, identity: &str) -> Result<(), NodeRegistryError>;
+
+    /// Replaces all nodes in the registry with the given list of nodes.
+    ///
+    /// # Arguments
+    ///
+    ///  * `nodes` - The list of nodes that will comprise the registry.
+    fn replace_all(&self, nodes: Vec<Node>) -> Result<(), NodeRegistryError>;
```
peterschwarz: This should be an `&mut Iterator<Item=Node>`

libsplinter/src/node_registry/rest_api.rs
```diff
@@ -379,6 +383,54 @@ where
...
+    Box::new(
+        payload
+            .from_err::<Error>()
+            .fold(web::BytesMut::new(), move |mut body, chunk| {
+                body.extend_from_slice(&chunk);
+                Ok::<_, Error>(body)
+            })
+            .into_future()
+            .and_then(
+                move |body| match serde_json::from_slice::<Vec<Node>>(&body) {
```
peterschwarz: Is there a way to stream these items, as opposed to loading them into a vector?

ltseeley:
Closing in favor of smaller PRs for each subcommand

ltseeley:
Reopening for visibility

cli/src/main.rs
desktophero: "add" from the command line would get messy, quick. Honestly, a preferred option from me would be to have an `apply` verb in the CLI that accepts an argument as a file path. 

If the object does not exist, `apply` based on the file.
If the object does exist, `apply` based on the file

desktophero:
Not sure if comments are still open so leaving this here.

In addition to my comment [here, detailing `apply`](https://github.com/Cargill/splinter/pull/390#discussion_r366009579), some additional feedback

I was then looking at the get/show verb
I prefer `get` and then add additional args to manage the output
for example:
```
# full list, tab view, basic output
cli get <object_type>

# specific item, tab view, basic output
cli get <object_type> <object_id>

# manage output, YAML format, verbose output
cli get <object_type> <object_id> -o YAML
```

ltseeley:
This is very stale; closing

PR #392, opened 2019-12-19
Unified node registry

peterschwarz:


libsplinter/src/node_registry/unified.rs
```diff
@@ -0,0 +1,375 @@
...
+// limitations under the License.
+
+//! Unified NodeRegistry implementations.
+//!
+//! This module provides a unified node registry which combines the node data from one or more
+//! read-only node registries with one local read-write node registry.  The data is merged from the
+//! local source into values from the read-only sources, allowing for an overlay or extension to
+//! existing node meta data.
+//!
+//! This module is behind the `"node-registry-unified"` feature, and is considered experimental.
```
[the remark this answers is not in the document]
peterschwarz: It should be noted for every thing included in the doc that is not enabled by default.

libsplinter/src/node_registry/unified.rs
```diff
@@ -0,0 +1,359 @@
...
+
+use super::{
+    MetadataPredicate, Node, NodeRegistryError, NodeRegistryReader, NodeRegistryWriter,
+    RwNodeRegistry,
+};
+
+/// Unifies a set of read-only node registries with a local, read-write node registry.
+///
+/// Nodes read from the unified registry utilize the read-only sources to fetch node definitions
+/// and then overlay any metadata values based on the local source.
```
ltseeley: Not sure this description is accurate anymore.

libsplinter/src/node_registry/unified.rs
```diff
@@ -0,0 +1,359 @@
...
+            Some(node!("node1", "meta_b" => "b value")),
+            iterator.next()
+        );
+
+        let predicates = vec![MetadataPredicate::ne("meta_b", "b value")];
+        let mut iterator = unified
+            .list_nodes(&predicates)
+            .expect("Unable to list nodes");
+
+        assert_eq!(None, iterator.next());
```
ltseeley: Probably not relevant to this PR, but should node2 actually be returned here? Does the ne predicate mean "filter out all nodes that don't have this metadata or have the metadata and it doesn't match" or "filter out all nodes that have this metadata and it doesn't match"? I would assume the latter if I was using this. Maybe we need a "has metadata field" predicate?

libsplinter/src/node_registry/unified.rs
peterschwarz: Updated

libsplinter/src/node_registry/unified.rs
peterschwarz: Added a commit to fix this, and updated the test to cover the scenario.

PR #393, opened 2019-12-19
Add database migration command to Splinter CLI

dplumb94:
Adds a new subcommand, `$ splinter database migrate`, to the Splinter CLI. Also adds Rust features that correspond to the libsplinter Rust features for database-requiring modules.

To test I used this compose file:

```
# test-migrations.yaml

version: "3.7"

services:
  splinter-cli:
    build:
      context: .
      dockerfile: cli/Dockerfile-installed-${DISTRO}
      args:
        - CARGO_ARGS=${CARGO_ARGS}
        - REPO_VERSION=${REPO_VERSION}
    image: ${REGISTRY}${NAMESPACE}splinter-cli:${ISOLATION_ID}
    container_name: splinter-cli
    command: bash -c "tail -f /dev/null"

  splinterd-db:
    image: postgres
    environment:
      POSTGRES_USER: admin
      POSTGRES_PASSWORD: admin
      POSTGRES_DB: splinterd
    expose:
      - 5432
    ports:
      - "5432:5432"
    restart: unless-stopped

  adminer:
    image: adminer
    container_name: splinterd-db-adminer
    restart: always
    ports:
      - "8080:8080"
```
First set some cargo args for the features you want enabled:
`$ export 'CARGO_ARGS=-- --features database'`

Start up the compose file:
`$ docker-compose -f test-migrations up`

Exec into the CLI and run the migrations:
`$ docker exec -it splinter-cli bash`
`# splinter database migrate -C postgres://admin:admin@splinterd-db:5432/splinterd`

You should see the message: "Successfully applied setup migrations". None of the other migrations were run because we only had the `database` feature enabled.

You can inspect the database using adminer (instructions at the bottom) or another method of your choosing to see that the ` __diesel_schema_migrations` table was created.

To run other migrations, down your docker-compose network and set some more CLI args:
`$ export 'CARGO_ARGS=-- --features database-migrate-biome-credentials'`
`$ docker-compose -f test-migrations.yaml build splinter-cli`
 (This would build the cli with `database`, `database-migrate-biome-users`, and `database-migrate-biome-credentials` features)

Reup your compose network, exec into the CLI and run the migrations again:
`$ docker-compose -f test-migrations up`
`$ docker exec -it splinter-cli bash`
`# splinter database migrate -C postgres://admin:admin@splinterd-db:5432/splinterd`

Now you should see:
"""
Successfully applied setup migrations
Successfully applied biome users migrations
Successfully applied biome credentials migrations
"""

Adminer:
[...]

PR #394, opened 2019-12-19
Add Biome users API routes

dnewh:
Adds the biome and biome-credentials API routes to splinterd behind the corresponding experimental features. This also, enables connecting to a database behind the database experimental feature.

This was tested with the changes to migrations introduced in PR #393

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -45,6 +45,9 @@ postgres = [
     "splinter/postgres",
 ]
 
+# Aliases - used to map libsplinter feature names to their CLI counterparts
+biome-credentials = ["database-migrate-biome-credentials"]
+
 [package]
 name = "splinter-cli"
 version = "0.3.9"
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -31,6 +31,8 @@ services:
       dockerfile: splinterd/Dockerfile-installed-${DISTRO}
       args:
         - REPO_VERSION=${REPO_VERSION}
+        - CARGO_ARGS=${CARGO_ARGS}
+        - SPLINTERD_ARGS=${SPLINTERD_ARGS}
     entrypoint: |
       bash -c "
         splinter cert generate --skip && \
@@ -38,6 +40,7 @@ services:
             -c ./configs/splinterd-node-0-docker.toml \
             --registry-file ./node_registries/nodes.yaml \
             --insecure \
+            $SPLINTERD_ARGS \
             -vv
       "
 
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -56,6 +56,7 @@ circuit-read = ["splinter/circuit-read"]
 proposal-read = ["splinter/proposal-read"]
 config-builder = []
 config-toml = ["config-builder"]
+database = ["splinter/database"]
 generate-certs = []
 
 [package.metadata.deb]
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -45,6 +45,8 @@ default = []
 stable = ["default"]
 
 experimental = [
+    "biome",
+    "biome-credentials",
     "circuit-read",
     "config-builder",
     "config-toml",
@@ -52,6 +54,8 @@ experimental = [
     "proposal-read"
 ]
 
+biome = ["splinter/biome", "database"]
+biome-credentials = ["splinter/biome-credentials", "biome"]
 circuit-read = ["splinter/circuit-read"]
 proposal-read = ["splinter/proposal-read"]
 config-builder = []
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -21,6 +21,8 @@ use crossbeam_channel;
 #[cfg(feature = "health")]
 use health::HealthService;
 use splinter::admin::service::{admin_service_id, AdminService};
+#[cfg(feature = "biome")]
+use splinter::biome::rest_api::{BiomeRestResourceManager, BiomeRestResourceManagerBuilder};
 use splinter::circuit::directory::CircuitDirectory;
 use splinter::circuit::handlers::{
     AdminDirectMessageHandler, CircuitDirectMessageHandler, CircuitErrorHandler,
@@ -29,6 +31,8 @@ use splinter::circuit::handlers::{
 #[cfg(feature = "circuit-read")]
 use splinter::circuit::rest_api::CircuitResourceProvider;
 use splinter::circuit::SplinterState;
+#[cfg(feature = "biome")]
+use splinter::database::{self, ConnectionPool};
 use splinter::keys::{
     insecure::AllowAllKeyPermissionManager, rest_api::KeyRegistryManager,
     storage::StorageKeyRegistry,
@@ -384,6 +388,12 @@ impl SplinterDaemon {
             rest_api_builder = rest_api_builder.add_resources(circuit_resource.resources());
         }
 
+        #[cfg(feature = "biome")]
+        {
+            let biome_resources = build_biome_routes(&self.db_url)?;
+            rest_api_builder = rest_api_builder.add_resources(biome_resources.resources());
+        }
+
         let (rest_api_shutdown_handle, rest_api_join_handle) = rest_api_builder.build()?.run()?;
 
         let (admin_shutdown_handle, service_processor_join_handle) =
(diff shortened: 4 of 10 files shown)


splinterd/Cargo.toml
```diff
@@ -45,15 +45,19 @@ default = ["generate-certs"]
 stable = ["default"]
 
 experimental = [
-    "circuit-read",
+    "biome",
```
vaporos: Looks like "circuit-read" was removed here, perhaps accidentally.

splinterd/src/daemon.rs
```diff
@@ -382,6 +388,24 @@ impl SplinterDaemon {
...
+        {
+            info!("Adding biome routes");
+            let connection_pool: ConnectionPool =
+                database::create_connection_pool(&self.db_url).unwrap();
+            let mut biome_rest_provider_builder: BiomeRestResourceManagerBuilder =
+                Default::default();
+            biome_rest_provider_builder =
+                biome_rest_provider_builder.with_user_store(connection_pool.clone());
+            #[cfg(feature = "biome-credentials")]
+            {
```
vaporos: This shouldn't need the block (it is a single statement).

splinterd/src/daemon.rs
```diff
@@ -382,6 +388,24 @@ impl SplinterDaemon {
             rest_api_builder = rest_api_builder.add_resources(circuit_resource.resources());
         }
 
+        #[cfg(feature = "biome")]
+        {
```
vaporos: When we ever remove this feature, we would have to change the indent of the implementation (or leave a block indented without purpose). Maybe consider making this a function as a result, and calling the function. The call to the function and the function itself can both be guarded by the feature.

splinterd/src/daemon.rs
```diff
@@ -646,6 +671,11 @@ impl SplinterDaemonBuilder {
         self
     }
 
+    pub fn with_db_url(mut self, value: String) -> Self {
```
vaporos: This should be guarded by the "database" feature (as well as the struct's field).

splinterd/src/main.rs
```diff
@@ -148,6 +148,8 @@ fn main() {
           "If set tls should accept all peer certificates")
         (@arg bind: --("bind") +takes_value
           "Connection endpoint for REST API")
+        (@arg database: --("db") +takes_value
```
vaporos: Here we use '--db' but in the configuration file it is 'database'. Suggest we use '--database'/database or '--db-url'/db_url so we have consistency.

splinterd/src/main.rs
```diff
@@ -293,6 +295,13 @@ fn main() {
         .or_else(|| Some("127.0.0.1:8080".to_string()))
         .expect("Must provide a url for REST API endpoint");
 
+    let db_url = matches
```
vaporos: This and the arg stuff should all be behind the 'database' feature flag.

docker-compose.yaml
```diff
@@ -31,12 +31,14 @@ services:
...
       args:
         - REPO_VERSION=${REPO_VERSION}
+        - CARGO_ARGS=${CARGO_ARGS}
     entrypoint: |
       bash -c "
         splinterd \
             --generate-certs \
             -c ./configs/splinterd-node-0-docker.toml \
             --registry-file ./node_registries/nodes.yaml \
+            --db postgres://splinter_admin:splinter_test@splinter-db:5432/splinter \
```
vaporos: This becomes problematic if behind a feature flag. We might need to add a SPLINTERD_ARGS env variable so we can pass this through.

splinterd/src/daemon.rs
dnewh: This is a limitation with Rust. Specifically, it throws an error saying attributes cannot be used on expressions.

splinterd/Cargo.toml
dnewh: Accident, updated.

splinterd/src/daemon.rs
dnewh: updated

splinterd/src/main.rs
dnewh: updated

docker-compose.yaml
dnewh: added SPLINTERD_ARGS

splinterd/src/main.rs
dnewh: updated to --database

splinterd/src/daemon.rs
dnewh: Update with the add_biome_routes function. It's a little gross with the memory swapping but it works

splinterd/Cargo.toml
```diff
@@ -45,15 +45,20 @@ default = []
...
     "config-toml",
     "health",
 ]
 
-circuit-read = ["splinter/circuit-read"]
+biome = ["splinter/biome", "database"]
+biome-credentials = ["splinter/biome-credentials", "biome"]
 config-builder = []
 config-toml = ["config-builder"]
+circuit-read = ["splinter/circuit-read"]
```
peterschwarz: Looks like this was moved - it should be before `config-*`

dnewh: rebase issue. Fixed

cli/Cargo.toml
```diff
@@ -26,7 +26,7 @@ experimental = [
 ]
 
 health = ["reqwest", "serde_json"]
-
+biome-credentials = ["database-migrate-biome-credentials"]
```
vaporos: This should be placed in a separate list below these features, with a comment above it such as "# Aliases - these aliases exist to map libsplinter feature names to their CLI equivalents" (or similar).

docker-compose.yaml
```diff
@@ -31,13 +31,16 @@ services:
       dockerfile: splinterd/Dockerfile-installed-${DISTRO}
       args:
         - REPO_VERSION=${REPO_VERSION}
+        - CARGO_ARGS=${CARGO_ARGS}
```
vaporos: The commit message "Add example splinter db url to splinter-node-0 image." doesn't seem to exactly reflect what is in this commit.

dnewh: updated

cli/Cargo.toml
dnewh: updated

splinterd/src/main.rs
```diff
@@ -147,6 +147,8 @@ fn main() {
           "If set tls should accept all peer certificates")
         (@arg bind: --("bind") +takes_value
           "Connection endpoint for REST API")
+        (@arg database: --("database") +takes_value
```
peterschwarz: This will be included if the `"database"` feature is off. It should match the pattern below, like the `generate-certs` block.

splinterd/src/main.rs
```diff
@@ -302,11 +318,20 @@ fn main() {
         .map(String::from)
         .or_else(|| config.registry_file());
 
+    // Allow unused mut for experimental features
+    #[allow(unused_mut)]
+    let mut feature_fields = "".to_string();
```
peterschwarz: This is debug output, so it should only be done within the `debug!` macro's call.

You should replace `feature_fields` in the debug log statement with something like:
```rust
        ...
        heartbeat_interval,
        if cfg!(feature = "database") {
            format!("{}, db_url: {}", feature_fields, db_url)
        } else {
            String::new()
        }
    );
```

dnewh: updated

splinterd/src/daemon.rs
```diff
@@ -593,6 +605,28 @@ fn start_health_service(
     })?
 }
 
+#[cfg(feature = "biome")]
+fn build_biome_routes(db_url: &str) -> BiomeRestResourceManager {
+    info!("Adding biome routes");
+    let connection_pool: ConnectionPool = database::create_connection_pool(db_url)
+        .map_err(|err| database::DatabaseError::ConnectionError(Box::new(err)))
+        .unwrap();
```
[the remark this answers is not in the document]
dnewh: handled

splinterd/src/main.rs
```diff
@@ -292,6 +300,14 @@ fn main() {
         .or_else(|| Some("127.0.0.1:8080".to_string()))
         .expect("Must provide a url for REST API endpoint");
 
+    #[cfg(feature = "database")]
+    let db_url = matches
+        .value_of("database")
+        .map(String::from)
+        .or_else(|| config.database())
+        .or_else(|| Some("127.0.0.1:5432".to_string()))
+        .expect("Must provide a url for database");
```
vaporos: CLIs should not panic with user-caused errors.

splinterd/src/main.rs
```diff
@@ -164,6 +164,14 @@ fn main() {
             .takes_value(true),
     );
 
+    #[cfg(feature = "database")]
+    let app = app.arg(
+        Arg::with_name("database")
+            .long("database")
+            .long_help("DB connection url, postgres://db_admin:db_password@127.0.0.1:5432/db")
```
vaporos: Suggest removing the example URL here; need to keep these short to cut down on wrapping.

dnewh: updated to remove example url

splinterd/src/main.rs
vaporos: This will be handled in a separate PR.

cli/Cargo.toml
```diff
@@ -45,6 +45,9 @@ postgres = [
     "splinter/postgres",
 ]
 
+# Aliases - used to map libsplinter feature names to their CLI counterparts
+biome-credentials = ["database-migrate-biome-credentials"]
```
peterschwarz: This should be in a separate commit, since it relates only to features of the CLI

dnewh: Updated

PR #395, opened 2019-12-19
Add Matrix trait

RyanLassigBanks:
Add matrix experimental feature and integrates it into connection manager

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -30,24 +30,32 @@ pub use messages::{
 use pacemaker::Pacemaker;
 use protobuf::Message;
 
-use crate::mesh::{Envelope, Mesh};
+use crate::matrix::{MatrixLifeCycle, MatrixSender};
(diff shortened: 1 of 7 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -360,8 +381,8 @@ fn notify_subscribers(
     });
 }
 
-fn send_heartbeats(
-    state: &mut ConnectionState,
+fn send_heartbeats<T: MatrixLifeCycle, U: MatrixSender>(
+    state: &mut ConnectionState<T, U>,
```
peterschwarz: Given that this is private and only uses the `MatrixSender`, you can change the parameters to take a `&dyn MatrixSender` and you won't need the generics or the whole state

RyanLassigBanks: It would still need state to access connection metadata though..

peterschwarz: Ah, I didn't see the later use of state

PR #396, opened 2019-12-20
Connection manager CLI example

RyanLassigBanks:
Adds `cm-repl`, an example daemon using the connection manger and refactors notifications API

### Testing

1) Build with docker
```
docker-compose -f examples/cm-repl/docker-compose.yaml
```
2) Exec into one of the node containers
```
docker exec -it cm-repl_node-0_1 bash
```
3) Run connection manager client commands
```
cmc connection add tcp://node-0:3040
cmc connection list
cmc connection remove tcp://node-0:3040
```

### DIFF
--- a/examples/cm-repl/cli/src/actions.rs
+++ b/examples/cm-repl/cli/src/actions.rs
@@ -1,0 +1,66 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use reqwest;
+
+pub fn add_connection(url: &str, addr: &str) {
+    let response = reqwest::Client::new()
+        .post(&format!("{}/connections/create", url))
+        .json(&json!({
+            "url": addr.to_string()
+        }))
+        .send();
+
+    match response {
+        Ok(mut r) => {
+            println!("{}", r.text().unwrap());
+        }
+        Err(err) => {
+            error!("{}", err);
+        }
+    };
+}
+
+pub fn remove_connection(url: &str, addr: &str) {
+    let response = reqwest::Client::new()
+        .delete(&format!("{}/connections/delete", url))
+        .json(&json!({
+            "url": addr.to_string()
+        }))
+        .send();
+
+    match response {
+        Ok(mut r) => {
+            println!("{}", r.text().unwrap());
+        }
+        Err(err) => {
+            error!("{}", err);
+        }
+    };
+}
+
+pub fn list_connections(url: &str) {
+    let response = reqwest::Client::new()
+        .get(&format!("{}/connections/fetch", url))
+        .send();
+
+    match response {
+        Ok(mut r) => {
+            println!("{}", r.text().unwrap());
+        }
+        Err(err) => {
+            error!("{}", err);
+        }
+    };
+}
--- a/examples/cm-repl/cli/src/main.rs
+++ b/examples/cm-repl/cli/src/main.rs
@@ -1,0 +1,88 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[macro_use]
+extern crate serde_json;
+#[macro_use]
+extern crate log;
+
+mod actions;
+
+use clap::{clap_app, crate_version};
+use flexi_logger::{DeferredNow, LogSpecBuilder, Logger};
+use log::Record;
+
+use actions::{add_connection, list_connections, remove_connection};
+
+pub fn log_format(
+    w: &mut dyn std::io::Write,
+    _now: &mut DeferredNow,
+    record: &Record,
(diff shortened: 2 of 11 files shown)


examples/cm-repl/daemon/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "cm-repl-daemon"
+version = "0.1.0"
+authors = ["Cargill Incorporated"]
+edition = "2018"
+
+[dependencies]
+splinter = { path = "../../../libsplinter", features = ["connection-manager", "matrix"]}
```
peterschwarz: Alphabetize these

Cargo.toml
```diff
@@ -24,5 +24,6 @@ members = [
     "examples/private_xo",
     "examples/gameroom/database",
     "examples/gameroom/daemon",
+    "examples/cm-repl/daemon",
```
peterschwarz: Could you give some more context in this commit message?  Is this a daemon process that provides the functionality to a REPL or the REPL itself?  Normally I would think of a CLI tool, so it's not quite clear.

examples/cm-repl/cli/src/actions.rs
```diff
@@ -0,0 +1,66 @@
...
+            "url": addr.to_string()
+        }))
+        .send();
+
+    match response {
+        Ok(mut r) => {
+            println!("{}", r.text().unwrap());
+        }
+        Err(err) => {
+            error!("{:?}", err);
```
peterschwarz: We should probably print these using `{}` (i.e. `Display`) not `Debug`

examples/cm-repl/cli/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+        (@setting SubcommandRequiredElseHelp)
+        (@subcommand connection =>
+            (@setting SubcommandRequiredElseHelp)
+            (about: "Add, remove, and list connections")
+            (@subcommand add =>
+                (about: "Add a new connection")
+                (@arg address: +takes_value +required "Node address you want to add as a connection")
+            )
+            (@subcommand remove =>
+                (about: "remove connection")
```
chenette: Inconsistent: Start `about` content with a capital letter.

examples/cm-repl/cli/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+    record: &Record,
+) -> Result<(), std::io::Error> {
+    write!(w, "{}", record.args(),)
+}
+
+fn main() {
+    let matches = clap_app!(cmc =>
+        (version: crate_version!())
+        (about: "Connection manager client (cmc)")
+        (@arg url: +takes_value "Rest api address")
```
chenette: REST API

Also, can you clarify which REST API is specified here? For example, is this the endpoint for the scabbard REST API?  The Splinter REST API?  Something else?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 mod error;
-mod messages;
+pub mod messages;
```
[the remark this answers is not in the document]
RyanLassigBanks: >If this module is now public, we should remove the Cm prefix

Do you think this should be call `Message`, `ConnectionManagerMessage`, or something else? Personally I'd advocate for `Message`.

>We should also consider being consistent on whether or not we pluralize the module

The convention we seem to have adopted for splinter is to `error.rs` for error modules and `messages.rs` for message mondules. So the names are consistent with other modules in project atm. FWIW I think this discussion is out of this PRs scope.

examples/cm-repl/daemon/Dockerfile
```diff
@@ -0,0 +1,58 @@
+# Copyright 2019 Cargill Incorporated
```
rbuysse: Should be 2018-2020

examples/cm-repl/daemon/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut cm = ConnectionManager::new(
+        mesh.get_life_cycle(),
+        mesh.get_sender(),
+        Box::new(transport),
+    );
+
+    let mesh_clone = mesh.clone();
+    let _ = thread::spawn(move || {
+        for connection_result in listener.incoming() {
+            info!("Recieved connection");
```
peterschwarz: typo "Recieved" -> Received

examples/cm-repl/daemon/src/rest_api.rs
```diff
@@ -0,0 +1,136 @@
...
+                    body.extend_from_slice(&chunk);
+                    Ok::<_, Error>(body)
+                })
+                .and_then(move |body| match serde_json::from_slice::<Payload>(&body) {
+                    Ok(payload) => match connector.request_connection(&payload.url) {
+                        Ok(CmResponse::AddConnection {
+                            status,
+                            error_message,
+                        }) if status == CmResponseStatus::Error => {
+                            HttpResponse::InternalServerError().body(format!("{:?}", error_message))
```
peterschwarz: This, and the other messages should be formatted using display.

Cargo.toml
peterschwarz: The intent of this request was to write up a description of the server.  The line you put in the description would have worked just as fine for a summary line.  The message should include more information about the repl server, what it does, as of this commit etc.

peterschwarz: Likewise, the cli commit could use a similar description.

libsplinter/src/network/connection_manager/mod.rs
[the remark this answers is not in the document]
peterschwarz: Since this PR was originally written, I think the changes required is that module shouldn't be public, and the connector should be tranforming those from the internal message to the appropriate `Result<T, ConnectionManagerError>`.  That probably goes beyond the scope of this PR.

Cargo.toml
RyanLassigBanks: Updated commit messages, let me know if they're to your liking, I don't want phone anything in :)

examples/cm-repl/daemon/Dockerfile
RyanLassigBanks: updated

examples/cm-repl/daemon/src/main.rs
RyanLassigBanks: fixed

examples/cm-repl/daemon/src/rest_api.rs
RyanLassigBanks: This would require implementing display for the structs in `message.rs` which I think is out of scope for this PR.

Cargo.toml
peterschwarz: These are great!

RyanLassigBanks:
No longer relevent

PR #359, opened 2019-11-25
Remove generate-certs from default features

agunde406:
This PR remove generate-certs as a default features and updates the docker-compose files to use the splinter-cli to generate the required certs and keys for the splinter daemon. 

Also changes the name of generate-key-registry volume to game_shared and updates the gameroom README to use the new path.

PR #397, opened 2019-12-20
Remove generate-certs from default features

rbuysse:
Refactor of https://github.com/Cargill/splinter/pull/359. Instead of using shared volumes to create and share certificates, this PR adds the splinter cli to the splinterd image so cert creation can happen locally.


vaporos:
As a future enhancement on this, we should make it so we can restart the containers (adding the appropriate options to cert generate).

PR #398, opened 2019-12-20
List connection message

RyanLassigBanks:
Adds list connection message type, which is returns a list of endpoints
the connection manager is connected to. This is meant to be used for
debugging purposes.


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -168,6 +168,10 @@ impl Connector {
         }
     }
 
+    pub fn list_connections(&self) -> Result<CmResponse, ConnectionManagerError> {
```
peterschwarz: Why doesn't this return `Vec<String>` and translate any errors of the response into `ConnectionManagerError`.  That way `CmResponse` remains internal

RyanLassigBanks: updated

PR #399, opened 2019-12-23
Add --skip flag to compose files where certs are generated

rbuysse:


arsulegai:
Interesting to see `--skip` against something like `--force` for the cert generation.

PR #400, opened 2019-12-26
Localize the Key Registries

peterschwarz:
This PR changes the way the key registry is utilized by removing the assumption that the key registry is shared between the nodes.  Keys are now considered a local concern of a splinter node.

When a circuit proposal or vote is submitted, the signer is first checked to see if they are allowed to make that change on the nodes behalf.  If not, the request is rejected.   Likewise, if a node receives a proposal or vote via consensus, and it pertains to itself, the signer is checked.  If the signer is unknown or does not have permissions, the change is marked as invalid and fails consensus.

The docker-compose file for the Gameroom example has been updated to generate key registries for each node individually.  Note that to retrieve these private keys, the user must follow slightly different instructions:

```
$ docker-compose -f examples/gameroom/docker-compose.yaml exec splinterd-node-acme bash
root@<container-id>:/# cat /var/lib/splinter/alice.priv; echo ""
<the private key value>
root@<container-id>:/#
```

and likewise for bubba and bob's key.

### DIFF
--- a/examples/gameroom/README.md
+++ b/examples/gameroom/README.md
@@ -31,17 +31,18 @@ and [Docker Compose](https://docs.docker.com/compose).
      docker-compose -f examples/gameroom/docker-compose.yaml up
      ```
 1. To extract private keys to use in the web application, run bash using the
-   `generate-key-registry` image and read the private key.  For example, to get
-   Alice's private key:
+   associated splinter node image and read the private key.  For example, to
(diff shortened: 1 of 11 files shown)


peterschwarz:
This PR is on hold for some discussions related to authorization and verification for admin keys that can make this type of system work in a trusted manner.

peterschwarz:
Closing, as this is no longer the right approach

PR #404, opened 2019-12-27
Publish docker images for pushed tags

rbuysse:


rbuysse:
From the travis documentation:

> If you need to run multiple commands, write a executable wrapper script that runs them all. The argument to script: in the script deployment provider needs to be a single command.

https://docs.travis-ci.com/user/deployment/script/

vaporos:
You may want to set -e at the beginning of the script.

PR #407, opened 2020-01-02
Run no-default-features during cron builds

rbuysse:

### DIFF
--- a/tests/test-splinter-cron.yaml
+++ b/tests/test-splinter-cron.yaml
@@ -1,0 +1,33 @@
+# Copyright 2019 Cargill Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


rbuysse:
Example build from my fork: https://travis-ci.org/rbuysse/splinter/builds/631914109

tests/test-splinter-cron.yaml
```diff
@@ -0,0 +1,33 @@
...
+  unit-test-splinter:
+    build:
+      context: ../
+      dockerfile: tests/Dockerfile
+    image: test-splinter:${ISOLATION_ID}
+    volumes:
+      - ../:/project/splinter
+    command: |
+        bash -c "
+            echo \"Running no-default-features tests...\" && \
```
ltseeley: Is there a specific reason we aren't doing this as part of PR builds, or is it just that we're already running the unit tests 3 times for PRs?

rbuysse: @ltseeley Yeah, we don't expect this to break often, and we're always on a quest to reduce build times.

If it turns out that it fails frequently, we can move it into to PRs.

PR #408, opened 2020-01-03
Hide key-management code behind features

dnewh:


libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -72,6 +73,7 @@ pub struct BiomeRestResourceManager {
     user_store: Arc<SplinterUserStore>,
     // Disable lint warning, for now this is only used if the biome-credentials feature is enabled
     #[allow(dead_code)]
+    #[cfg(feature = "biome-key-management")]
```
vaporos: You could likely remove the #[allow(dead_code)] above. This new feature doesn't jive with the comment above.

dnewh: updated

PR #410, opened 2020-01-06
Implement splinter CLI circuit create command

peterschwarz:
To test, start the game room example normally.

Create a `circuit.yaml` with content like the following:

```yaml
circuit_id: my-circuit
roster:
  - service_id: my-scabbard-a
    service_type: scabbard
    allowed_nodes:
      - acme-node-000
  - service_id: my-scabbard-b
    service_type: scabbard
    allowed_nodes:
      - bubba-node-000
members:
  - node_id: acme-node-000
    endpoint: tls://splinterd-node-acme:8044
  - node_id: bubba-node-000
    endpoint: tls://splinterd-node-bubba:8044
authorization_type: Trust
durability: NoDurability
circuit_management_type: my-type
```

Extract the keys from one of the splinter nodes:

```
$ docker cp splinterd-node-acme:/key_registry_shared/ <some local dir>
```

Run the command:

```
$ cargo run --manifest-path cli/Cargo.toml --features circuit -- circuit create -k <some local dir>/key_registry_shared/alice.priv -U http://localhost:8088 circuit.yaml
```


cli/src/action/circuit/api.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright 2019 Cargill Incorporated
```
jsmitchell: 2020

cli/src/action/circuit/payload.rs
```diff
@@ -0,0 +1,98 @@
+// Copyright 2019 Cargill Incorporated
```
jsmitchell: 2020

cli/src/action/circuit/payload.rs
```diff
@@ -0,0 +1,98 @@
...
+use super::signing;
+
+/// A circuit action that has a type and can be converted into a protobuf-serializable struct.
+pub trait CircuitAction<M: Message> {
+    fn into_proto(self) -> Result<M, CliError>;
+
+    fn action_type(&self) -> Action;
+}
+
+/// Applies an circuit payload action to the given CircuitManagementPayload.
```
jsmitchell: s/Applies an/Applies a/

cli/src/action/circuit/payload.rs
```diff
@@ -0,0 +1,98 @@
...
+    action: A,
+) -> Result<Vec<u8>, CliError>
+where
+    M: Message + ApplyToEnvelope,
+    A: CircuitAction<M>,
+{
+    let action_type = action.action_type();
+    let action_proto = action.into_proto()?;
+    let serialized_action = action_proto.write_to_bytes().map_err(|err| {
+        CliError::ActionError(format!("Unable to serialize action to to bytes: {}", err))
```
jsmitchell: s/to to/to/

cli/src/action/circuit/signing.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2019 Cargill Incorporated
```
jsmitchell: 2020

peterschwarz: It _was_ 2019 when I wrote it...

cli/src/action/circuit/api.rs
peterschwarz: fixed

cli/src/action/circuit/payload.rs
peterschwarz: fixed

cli/src/action/circuit/payload.rs
peterschwarz: fixed

cli/src/action/circuit/payload.rs
peterschwarz: fixed

cli/src/action/circuit/signing.rs
peterschwarz: fixed

jsmitchell:
Verified this works correctly in test.

cli/src/action/circuit/api.rs
```diff
@@ -0,0 +1,85 @@
...
+    url: &'a str,
+}
+
+impl<'a> SplinterRestClient<'a> {
+    /// Constructs a new client for a Splinter node at the given URL.
+    pub fn new(url: &'a str) -> Self {
+        Self { url }
+    }
+
+    /// Fetches the node id of this client's Splinter node.
```
chenette: "node ID" (per the Splinter capitalization guidelines at https://docs.google.com/document/d/1_-VwbB_YXdNMQJjcTIm5nOrhMECa9LF3lSgm7P_AhaU/edit#heading=h.ui1pysii1b8c )

cli/src/action/circuit/payload.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    fn action_type(&self) -> Action;
+}
+
+/// Applies a circuit payload action to the given CircuitManagementPayload.
+pub trait ApplyToEnvelope {
+    fn apply(self, circuit_management_payload: &mut CircuitManagementPayload);
+}
+
+/// Makes a signed, circuit management payload to be submitted to the splinter rest api.
```
chenette: Capitalization: "Splinter REST API"

cli/src/action/circuit/api.rs
peterschwarz: fixed

cli/src/action/circuit/payload.rs
peterschwarz: Fixed

cli/src/action/circuit/signing.rs
```diff
@@ -0,0 +1,87 @@
...
+        .map(|pub_key| pub_key.as_slice().to_vec())
+        .map_err(|err| {
+            CliError::ActionError(format!(
+                "Unable get public key from provided private key: {}",
+                err
+            ))
+        })
+}
+
+pub fn sign_bytes(private_key: &str, bytes: &[u8]) -> Result<Vec<u8>, CliError> {
```
vaporos: Should create a Signer and pass it from around main() instead of a private_key. Remove all singing wrapper code. The only signing code should be what we need to read the key and get a signer. No thread-global variable.

cli/src/action/circuit/mod.rs
```diff
@@ -28,7 +28,9 @@ impl Action for CircuitCreateAction {
     fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
         let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
         let url = args.value_of("url").unwrap_or("http://localhost:8085");
-        let key = args.value_of("private_key_file").unwrap_or("splinter");
+        let key = args
+            .value_of("private_key_file")
+            .unwrap_or("./splinter.priv");
```
vaporos: This might be a security issue. Instead of defaulting to the cwd, hard-code the /etc directory location.

peterschwarz: But this is a user key, so this default is based on the default generated when using keygen on a user basis, not the admin private key.

cli/src/action/circuit/signing.rs
peterschwarz: Dropped the signing commit and code.  Replaced with the use of a combination of sawtooth-sdk and splinter::signing modules

cli/src/action/circuit/mod.rs
vaporos: Ok, I did miss that - use a path int he user's home directory similar to how it works with the sawtooth tools.

peterschwarz: I will fix that in both places

peterschwarz: I think this should be addressed in a separate PR, as it requires the `keygen` subcommand to be modified.  Looking at it again, it looks to be under `splinter admin keygen` which doesn't seem right for a user key.  We should consider moving this to `splinter keygen`, with the user's home directory as the default target, as in the Sawtooth tools.

vaporos: Agreed.

PR #411, opened 2020-01-06
Biome user reorg

vaporos:
This is the beginning of reorganizing all of the database code to leave room for implementing different backends such as sqlite.


libsplinter/src/biome/user/mod.rs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-mod users;
+//! Users are central entity in the biome module. They represent a real person who uses a splinter
```
jsmitchell: s/are central/are the central/

libsplinter/src/biome/user/store/diesel/operations/add_user.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019 Cargill Incorporated
```
jsmitchell: 2020

vaporos: This was intentional. I'm going to separately propose that we replace all the copyright header lines at the same time using the same (2019-2020) copyright.

libsplinter/src/biome/user/mod.rs
chenette: To avoid the number mismatch ("users = the entity"), I'd suggest: "A user is the central entity in the biome module.  It represents ..."

Also, "Splinter" (cap S).

vaporos: Fixed - This was preexisting text, so I added a commit to fix it instead of updating the existing commit which just moves the text as-is.

libsplinter/Cargo.toml
```diff
@@ -87,9 +88,10 @@ experimental = [
 ]
 
 biome = ["database", "jsonwebtoken", "rand"]
-biome-credentials = ["biome", "database", "bcrypt"]
+biome-credentials = ["biome", "biome-user", "database", "bcrypt"]
 biome-key-management = ["biome", "database"]
```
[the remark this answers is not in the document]
vaporos: I did some compile testing, and it doesn't appear so from a purely code perspective. I did find another issue with just having 'biome' alone was broken, so I added a commit for that.

libsplinter/src/biome/user/mod.rs
vaporos: Fixed 'Splinter'

cli/src/action/database.rs
```diff
@@ -23,7 +23,7 @@ use splinter::biome::credentials::database::run_migrations as run_biome_credenti
 #[cfg(feature = "database-migrate-biome-notifications")]
 use splinter::biome::notifications::database::run_migrations as run_biome_notifications_migrations;
 #[cfg(feature = "database-migrate-biome-users")]
-use splinter::biome::users::database::run_migrations as run_biome_users_migrations;
+use splinter::biome::user::database::run_migrations as run_biome_user_migrations;
```
eloaverona: There's some import statements referencing `users` at 
`libsplinter/src/biome/credentials/rest_resources.rs` and `libsplinter/src/biome/rest_api/mod.rs`

vaporos: Fixed two instances of this.

PR #412, opened 2020-01-07
Add compose file that uses published images

rbuysse:
This will track master and is meant to be used along with the repository in
situations where building from scratch is not feasible.


examples/gameroom/docker-compose-dockerhub.yaml
```diff
@@ -0,0 +1,210 @@
+# Copyright 2019 Cargill Incorporated
```
jsmitchell: 2020

PR #413, opened 2020-01-08
Key management routes to biome rest api

eloaverona:
This PR adds a POST, GET and PATCH `/biome/users/{user_id}/keys` to the key management feature in biome.

### DIFF
--- a/libsplinter/src/rest_api/mod.rs
+++ b/libsplinter/src/rest_api/mod.rs
@@ -193,8 +193,7 @@ impl Resource {
             ) -> Box<dyn Future<Item = HttpResponse, Error = ActixError>>
             + Send
             + Sync
-            + 'static
-            + Clone,
+            + 'static,
     {
         Self::build(route).add_method(method, handle)
     }
@@ -214,8 +213,7 @@ impl Resource {
             ) -> Box<dyn Future<Item = HttpResponse, Error = ActixError>>
             + Send
             + Sync
-            + 'static
-            + Clone,
+            + 'static,
     {
         self.methods.push((method, Arc::new(Box::new(handle))));
         self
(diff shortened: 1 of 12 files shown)


libsplinter/src/biome/key_management/database/postgres/migrations/2019-12-12-200956_create-keys/up.sql
```diff
@@ -18,5 +18,6 @@ CREATE TABLE IF NOT EXISTS keys (
     encrypted_private_key TEXT NOT NULL,
     user_id               TEXT NOT NULL,
     display_name          TEXT NOT NULL,
-    PRIMARY KEY(public_key, user_id)
```
peterschwarz: As this migration may be part of a release, this change should probably be in a separate migration.

libsplinter/src/rest_api/mod.rs
```diff
@@ -380,6 +380,29 @@ pub fn percent_encode_filter_query(input: &str) -> String {
     percent_encoding::utf8_percent_encode(input, QUERY_ENCODE_SET).to_string()
 }
 
+pub fn get_header(header_key: &str, request: &HttpRequest) -> Result<String, RequestError> {
```
peterschwarz: Would suggest naming this `require_header`, as it fails if the header is missing.

libsplinter/src/biome/key_management/database/postgres/migrations/2019-12-12-200956_create-keys/up.sql
eloaverona: Addressed

libsplinter/src/rest_api/mod.rs
eloaverona: Addressed

eloaverona:
@peterschwarz I think the change to the model introduced by Shawn should be done in a separate PR, it sounds to me out of scope for this PR.

PR #419, opened 2020-01-10
User keygen

RyanLassigBanks:
This PR:

1) Has `admin keygen` generate keys in /etc/.splinter/keys by default
2) Add a command to cli called `keygen` that generates keys in $HOME/.splinter/keys

### DIFF
--- a/cli/src/action/admin.rs
+++ b/cli/src/action/admin.rs
@@ -37,9 +37,9 @@ use super::{chown, Action};
 const DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
 const STATE_DIR_ENV: &str = "SPLINTER_STATE_DIR";
 
-pub struct KeyGenAction;
+pub struct AdminKeyGenAction;
 
-impl Action for KeyGenAction {
+impl Action for AdminKeyGenAction {
     fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
         let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
 
--- a/cli/src/action/keygen.rs
+++ b/cli/src/action/keygen.rs
@@ -129,8 +129,7 @@ pub fn create_key_pair(
             .open(private_key_path.as_path())
             .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
 
-        private_key_file
-            .write(private_key.as_hex().as_bytes())
+        writeln!(&private_key_file, "{}", private_key.as_hex())
             .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
     }
 
@@ -148,8 +147,7 @@ pub fn create_key_pair(
             .open(public_key_path.as_path())
             .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
 
-        public_key_file
-            .write(public_key.as_hex().as_bytes())
+        writeln!(&public_key_file, "{}", public_key.as_hex())
             .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
     }
     if change_permissions {
--- a/cli/src/action/keygen.rs
+++ b/cli/src/action/keygen.rs
@@ -1,0 +1,161 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fs::{create_dir_all, metadata, OpenOptions};
+use std::io::prelude::*;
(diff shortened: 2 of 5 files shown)


cli/src/action/admin.rs
```diff
@@ -36,6 +36,7 @@ use crate::error::CliError;
 use super::{chown, Action};
 
 const DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
+const DEFAULT_KEYS_DIR: &str = "/etc/.splinter/keys/";
```
peterschwarz: should be `/etc/splinter/keys`

cli/src/action/admin.rs
```diff
@@ -61,10 +62,14 @@ impl Action for KeyGenAction {
         let key_name = args.value_of("key_name").unwrap_or("splinter");
         let key_dir = args
             .value_of("key_dir")
-            .or(Some("."))
+            .or(Some(DEFAULT_KEYS_DIR))
             .map(Path::new)
             .unwrap();
 
+        create_dir_all(key_dir).map_err(|err| {
```
peterschwarz: This directory should be created at installation time, not by this command.

cli/src/action/keygen.rs
```diff
@@ -0,0 +1,172 @@
...
+        let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
+
+        let key_name = args.value_of("key_name").unwrap_or("splinter");
+        let key_dir = args
+            .value_of("key_dir")
+            .map(PathBuf::from)
+            .or_else(|| {
+                dirs::home_dir().map(|mut p| {
+                    println!("path {:?}", p);
+                    p.push(".sawtooth/keys");
```
peterschwarz: sawtooth -> splinter

cli/src/action/keygen.rs
```diff
@@ -0,0 +1,172 @@
...
+            logger_handle.parse_new_spec("error");
+        }
+
+        Ok(())
+    }
+
+    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+        let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
+
+        let key_name = args.value_of("key_name").unwrap_or("splinter");
```
peterschwarz: The default name should possibly be the user name

cli/src/main.rs
```diff
@@ -55,7 +55,7 @@ fn run() -> Result<(), CliError> {
                 (about: "Generates secp256k1 keys to use when signing circuit proposals")
                 (@arg key_name: +takes_value "Name of the key to create; defaults to \"splinter\"")
                 (@arg key_dir: -d --("key-dir") +takes_value
-                 "Name of the directory in which to create the keys; defaults to current working directory")
+                 "Name of the directory in which to create the keys; defaults to /etc/.sawtooth")
```
jsmitchell: This should be /etc/splinter/keys, not /etc/.sawtooth?

cli/src/main.rs
```diff
@@ -91,6 +91,24 @@ fn run() -> Result<(), CliError> {
         )
     );
 
+    #[cfg(feature = "keygen")]
+    {
+        use clap::{Arg, SubCommand};
+
+        app = app.subcommand(
+            SubCommand::with_name("keygen")
+            .about("Generates secp256k1 keys to use when signing circuit proposals in user home directory")
```
vaporos: May be used for signing other things in the future, so the 'signing circuit proposals' is likely to become outdated.

cli/src/action/keygen.rs
```diff
@@ -145,8 +145,7 @@ pub fn create_key_pair(
             .open(private_key_path.as_path())
             .map_err(|err| CliError::EnvironmentError(format!("{}", err)))?;
 
-        private_key_file
-            .write(private_key.as_hex().as_bytes())
+        writeln!(&private_key_file, "{}", private_key.as_hex())
```
peterschwarz: This needs to be `&mut`

RyanLassigBanks: Not sure if that's the case. The build did fail because I messed up a rebase but it should be fixed now.

cli/src/main.rs
```diff
@@ -89,6 +89,38 @@ fn run() -> Result<(), CliError> {
...
+                .arg(
+                    Arg::with_name("quiet")
+                        .short("q")
+                        .long("quiet")
+                        .help("Do not display output"),
+                )
+                .arg(
+                    Arg::with_name("admin")
+                        .long("admin")
+                        .help("Generate admin keys in /etc/splinter/keys"),
```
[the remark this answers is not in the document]
RyanLassigBanks: Did we ever come to a decision on whether or not the name of the flag needs to be changed as well. If we're not changed the flag name I think referring to it as an admin key is fine.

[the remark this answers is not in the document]
chenette: I'm okay with keeping the "admin" flag and "admin keys" terminology for now. Discussions on a better term can happen later.

cli/src/main.rs
```diff
@@ -89,6 +89,38 @@ fn run() -> Result<(), CliError> {
         )
     );
 
+    #[cfg(feature = "keygen")]
+    {
+        use clap::{Arg, SubCommand};
+
+        app = app.subcommand(
+            SubCommand::with_name("keygen")
+                .about("Generates secp256k1 keys user home directory")
```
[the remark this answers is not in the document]
chenette: The changed version look good. Thanks!

chenette:
The changed help content looks good to me.  That's all I looked at, so I didn't click "Approve", but I've resolved all my earlier comments.

PR #422, opened 2020-01-14
Create ScabbardClient and scabbard CLI

ltseeley:


cli/Cargo.toml
```diff
@@ -54,7 +54,11 @@ license = "Apache-2.0"
 
 [[bin]]
 name = "splinter"
-path = "src/main.rs"
+path = "src/splinter/main.rs"
+
+[[bin]]
+name = "scabbard"
+path = "src/scabbard/main.rs"
```
ltseeley: I'm not sure how much sense it makes to package both CLI's together; maybe they belong in their own respective crates.

libsplinter/src/service/scabbard/client/key.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+
+use sawtooth_sdk::signing::secp256k1::Secp256k1PrivateKey;
+
+use super::Error;
+
+/// Load a private key from the local filesystem.
+pub fn load_signing_key_from_file(filename: &str) -> Result<Secp256k1PrivateKey, Error> {
```
ltseeley: Currently this (and by extension the scabbard CLI) just loads the key from a file. I'm not sure if this should load from a known location in the user's directory as well to match up with the keygen tools.

libsplinter/src/service/scabbard/client/key.rs
```diff
@@ -0,0 +1,93 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fs::File;
+use std::io::{BufRead, BufReader};
+
+use sawtooth_sdk::signing::secp256k1::Secp256k1PrivateKey;
```
ltseeley: The scabbard client should probably support the generic signing API in libsplinter itself rather than relying explicitly on the sawtooth SDK.

libsplinter/src/service/scabbard/client/error.rs
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error::Error as ErrorTrait;
+use std::fmt;
+
+#[derive(Debug)]
+pub struct Error(pub String);
```
ltseeley: It's not clear to me yet if it would be useful to enumerate this error.

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+            url: url.into(),
+            signing_key,
+            unsent_payloads: vec![],
+        }
+    }
+
+    /// Submit all unsent transactions to the scabbard service.
+    pub fn submit(mut self) -> Result<(), Error> {
+        let batch_list = batch_payloads(&mut self.unsent_payloads.drain(..), &self.signing_key)?;
+        let _batch_link = submit_batch_list(&self.url, &batch_list)?;
```
ltseeley: This currently doesn't implement waiting for the batch to finish committing.

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+        Ok(())
+    }
+
+    /// Queue a transaction to upload a smart contract loaded from the given `.scar`. See
+    /// `load_scar` for more on how the `.scar` is loaded.
+    pub fn upload_contract_from_scar(self, scar: &str) -> Result<Self, Error> {
```
ltseeley: It may be useful to take a more nuanced approach to this where not only is the .scar loaded and the upload transaction added, but it also adds the related transactions for the contract (e.g. creating registries and permissions).

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// Queue a transaction to delete a contract registry.
+    pub fn delete_contract_registry(mut self, name: &str) -> Result<Self, Error> {
+        self.unsent_payloads
+            .push(contract_registry::create_contract_registry_delete_payload(
+                name,
+            )?);
+        Ok(self)
+    }
+
+    /// Queue a transaction to create a smart permssion, loaded from the specified
```
agunde406: permssion -> permission (repeated several times below

libsplinter/src/service/scabbard/client/submit.rs
```diff
@@ -0,0 +1,176 @@
...
+        .map_err(|err| Error(format!("received error status code: {}", err)))?;
+
+    let batch_link: Link = response
+        .json()
+        .map_err(|err| Error(format!("failed to parse response as batch link: {}", err)))?;
+
+    Ok(batch_link.link)
+}
+
+// pub fn wait_for_batch(url: &str, wait: u64) -> Result<StatusResponse, Error> {
```
peterschwarz: Commented code, if there is no batch waiting.

agunde406: commented out code, did you intend to finish this?

libsplinter/src/service/scabbard/client/submit.rs
```diff
@@ -0,0 +1,176 @@
...
+    message: String,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct StatusResponse {
+    data: Vec<BatchStatus>,
+    link: String,
+}
+
+// impl StatusResponse {
```
agunde406: Commented out coded, related to wait?

libsplinter/src/service/scabbard/client/mod.rs
peterschwarz: It probably should

libsplinter/src/service/scabbard/client/error.rs
RyanLassigBanks: I think the source pattern we've used in a few places could be useful.

```
struct Error {
   source: Box<dyn Error>,
   message: String
}
```

cli/src/splinter/action/circuit/api.rs
```diff
@@ -33,7 +33,7 @@ impl<'a> SplinterRestClient<'a> {
     pub fn fetch_node_id(&self) -> Result<String, CliError> {
         Client::new()
             .get(&format!("{}/status", self.url))
-            .send()
+            .submit()
```
agunde406: Can you explain this change? Othere places in this file now also use send.

cli/src/scabbard/action/execute.rs
```diff
@@ -0,0 +1,99 @@
...
+
+use clap::{App, Arg, ArgMatches, SubCommand};
+use splinter::service::scabbard::client::ScabbardClient;
+
+use super::CliError;
+
+pub const SUBCMD: &str = "exec";
+const ABOUT_STR: &str = "execute a Sabre contract";
+
+const CONTRACT_ARG: &str = "contract";
```
vaporos: The use of const for all strings seems unnecessary and it's actually hard to read the code here as a result.

libsplinter/src/service/scabbard/client/key.rs
RyanLassigBanks: Yeah, I wonder if it's useful in practice to let uses specify a specific file path vs saying all keys must live an specific directory.

cli/src/scabbard/action/execute.rs
RyanLassigBanks: I second this

cli/src/splinter/action/circuit/api.rs
ltseeley: Oops, think this was a mistake.

libsplinter/src/service/scabbard/client/key.rs
ltseeley: @vaporos are you suggesting that I take the approach you suggested (which is basically the same as the approach for loading .scar files)?

ltseeley:
This PR is no longer necessary due to #432 and #454

PR #424, opened 2020-01-14
Fix scabbard initial events

peterschwarz:
This PR fixes a couple of issues with scabbard events:

- Reject requests with an empty `last_seen_event` query parameter
- Properly handle empty transaction receipt stores.

### DIFF
--- a/libsplinter/src/service/scabbard/rest_api.rs
+++ b/libsplinter/src/service/scabbard/rest_api.rs
@@ -64,10 +64,18 @@ pub fn make_subscribe_endpoint() -> ServiceEndpoint {
 
             let last_seen_event_id = query.remove("last_seen_event");
 
-            if let Some(ref id) = last_seen_event_id {
-                debug!("Getting all state-delta events since {}", id);
-            } else {
-                debug!("Getting all state-delta events");
+            match last_seen_event_id {
+                Some(ref id) if id.trim().is_empty() => {
+                    return Box::new(
+                        HttpResponse::BadRequest()
+                            .json(json!({
+                                "message": "last_seen_event must not be empty",
+                            }))
+                            .into_future(),
+                    );
+                }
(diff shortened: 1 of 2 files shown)


libsplinter/src/service/scabbard/state.rs
```diff
@@ -408,49 +414,57 @@ impl Events {
...
+        }
     }
 }
 
 impl Iterator for Events {
     type Item = StateChangeEvent;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.cache.is_empty() {
+        if self.cache.is_empty() && self.query != EventQuery::Exhausted {
```
ltseeley: Can you add a unit test to verify this change?

libsplinter/src/service/scabbard/rest_api.rs
```diff
@@ -64,10 +64,12 @@ pub fn make_subscribe_endpoint() -> ServiceEndpoint {
 
             let last_seen_event_id = query.remove("last_seen_event");
 
-            if let Some(ref id) = last_seen_event_id {
-                debug!("Getting all state-delta events since {}", id);
-            } else {
-                debug!("Getting all state-delta events");
+            match last_seen_event_id {
+                Some(ref id) if id.trim().is_empty() => {
```
ltseeley: Assuming that the empty string will never be a valid event ID, it's probably fine to just treat it as if the parameter wasn't provided. I think this might be the desired behavior, especially since the response doesn't include any information about why the request was bad, and it would not be immediately obvious to the user.

peterschwarz: Updated to return a message

libsplinter/src/service/scabbard/state.rs
peterschwarz: Added.

PR #434, opened 2020-01-17
Log debug msg when splinterd config file not found

ltseeley:
Update the splinterd config file loading procedure to simply log `debug`
message if the config file is missing, since this is not considered an
error.

### DIFF
--- a/splinterd/src/main.rs
+++ b/splinterd/src/main.rs
@@ -68,30 +68,43 @@ const HEARTBEAT_DEFAULT: u64 = 30;
 
 #[cfg(not(feature = "config-toml"))]
 fn load_toml_config(config_file_path: &str) -> Config {
-    File::open(config_file_path)
-        .map_err(ConfigError::from)
-        .and_then(Config::from_file)
-        .unwrap_or_else(|err| {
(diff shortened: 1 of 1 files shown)


splinterd/src/main.rs
```diff
@@ -68,30 +68,43 @@ const HEARTBEAT_DEFAULT: u64 = 30;
...
-        .map_err(ConfigError::from)
-        .and_then(Config::from_file)
-        .unwrap_or_else(|err| {
-            warn!("Unable to load {}: {}", config_file_path, err);
-            Config::default()
-        })
+    match File::open(config_file_path) {
+        Ok(f) => Config::from_file(f),
+        Err(err) if err.kind() == io::ErrorKind::NotFound => {
+            debug!("Configuration file not found: {}", config_file_path);
```
peterschwarz: Should this still be warn?  I thought the concern about this message was the error message rather then the log level.

ltseeley: @vaporos suggested that it be `debug` unless the config file was explicitly provided as a command-line arg (which we could do by moving the `unwrap_or` of the argument match into these methods).

PR #436, opened 2020-01-19
update gameroom cli commands with new binary name

desktophero:
Signed-off-by: Jason <j_walker@cargill.com>

# `changes`

- Gameroom example docker-compose YAML doesn't start with the new `splinter-cli` name change in `0.3.8`

### DIFF
--- a/examples/gameroom/docker-compose.yaml
+++ b/examples/gameroom/docker-compose.yaml
@@ -37,7 +37,7 @@ services:
           bash -c "
             if [ ! -f /key_registry/keys.yaml ]
             then
-              splinter admin keyregistry \
+              splinter-cli admin keyregistry \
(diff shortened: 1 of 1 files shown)


desktophero:
The cli binary is still being built as `splinter-cli` which causes the gameroom example to not work.

https://github.com/Cargill/splinter/blob/master/cli/Cargo.toml#L52

agunde406:
This is the line the says what the binary is named https://github.com/Cargill/splinter/blob/master/cli/Cargo.toml#L64 Is it possible that your gameroom is being built with an old image?

desktophero:
I'll purge my local images and try again

desktophero:
Purged local Docker cache and wasn't able to reproduce. Closing. 

Thanks!

PR #437, opened 2020-01-20
Removes panics from Splinter daemon start up caused by user

agunde406:
Instead an error will be returned, logged and the program will be exited.

Also makes openssl optional as it is only required if generate-certs feature is enabled.

### DIFF
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+use std::error::Error;
+use std::fmt;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, RwLock};
 use std::thread;
@@ -132,7 +134,7 @@ impl SplinterDaemon {
         // Load initial state from the configured storage location and create the new
         // SplinterState from the retrieved circuit directory
         let storage = get_storage(&self.storage_location, CircuitDirectory::new)
-            .map_err(|err| StartError::StorageError(format!("Storage Error: {}", err)))?;
+            .map_err(StartError::StorageError)?;
 
         let circuit_directory = storage.read().clone();
         let state = Arc::new(RwLock::new(SplinterState::new(
@@ -336,7 +338,7 @@ impl SplinterDaemon {
 
         let key_registry = Box::new(
             StorageKeyRegistry::new(self.key_registry_location.clone())
-                .map_err(|err| StartError::StorageError(format!("Storage Error: {}", err)))?,
+                .map_err(|err| StartError::StorageError(format!("{}", err)))?,
         );
 
         let admin_service = AdminService::new(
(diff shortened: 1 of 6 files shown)


splinterd/src/daemon.rs
```diff
@@ -940,25 +971,25 @@ impl From<ConnectError> for StartError {
 
 impl From<ConnectionError> for StartError {
     fn from(connection_error: ConnectionError) -> Self {
-        StartError::NetworkError(format!("Network Error: {:?}", connection_error))
+        StartError::NetworkError(format!("{:?}", connection_error))
```
ltseeley: Not sure if it makes any difference, but this can be accomplished with `err.to_string()` as well.

splinterd/src/daemon.rs
```diff
@@ -914,9 +916,38 @@ pub enum StartError {
...
+        match self {
+            StartError::TransportError(msg) => write!(f, "transport returned an error: {}", msg),
+            StartError::NetworkError(msg) => write!(f, "network returned an error: {}", msg),
+            StartError::StorageError(msg) => write!(f, "unable to set up storage: {}", msg),
+            StartError::ProtocolError(msg) => {
+                write!(f, "unable to start due to a protocol error: {}", msg)
+            }
+            StartError::RestApiError(msg) => {
+                write!(f, "unable to start due to a rest api error: {}", msg)
+            }
```
ltseeley: The format of these two variants differs from the others by including "unable to start due to ..."

splinterd/src/main.rs
```diff
@@ -617,6 +619,42 @@ pub enum GetTransportError {
...
+            GetTransportError::IoError(err) => Some(err),
+        }
+    }
+}
+
+impl fmt::Display for GetTransportError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            GetTransportError::CertError(msg) => {
+                write!(f, "ran into an error retrieving certificate: {}", msg)
```
ltseeley: Any reason for using "ran into ..." instead of "unable to retrieve ..."?

splinterd/src/main.rs
```diff
@@ -617,6 +619,42 @@ pub enum GetTransportError {
...
+            }
+            GetTransportError::NotSupportedError(msg) => {
+                write!(f, "recieved transport type that is not supported: {}", msg)
+            }
+            GetTransportError::TlsTransportError(err) => {
+                write!(f, "unable to create TLS transport: {:?}", err)
+            }
+            #[cfg(feature = "generate-certs")]
+            GetTransportError::OpensslError(err) => {
+                write!(f, "unable to generate certificatesc: {}", err)
```
ltseeley: Looks like a typo ("certificatesc"). Also, is cert generation the only reason for an `OpensslError` could occur?

splinterd/src/main.rs
```diff
@@ -223,41 +230,35 @@ fn main() {
...
-        .expect("Must provide a unique node id");
+        .ok_or_else(|| UserError::MissingArgument("node_id".into()))?;
 
     let storage_type = matches
         .value_of("storage")
         .map(String::from)
         .or_else(|| config.storage())
-        .or_else(|| Some(String::from("yaml")))
-        .expect("No Storage Provided");
+        .unwrap_or_else(|| String::from("yaml"));
```
ltseeley: Can we add these defaults to the arg definitions in the clap app instead? That way they're documented in the CLI help as well.

agunde406: Shawn wants to remove all defaults from cli/clap.

splinterd/src/main.rs
agunde406: Its the only place that uses openssl in the splinter daemon code.

splinterd/src/main.rs
agunde406: nope, ill update for consistency

splinterd/src/main.rs
agunde406: They will eventually be moved into a DefaultConfig struct.

splinterd/src/main.rs
```diff
@@ -606,15 +616,126 @@ fn get_transport(
...
+}
+
+impl fmt::Display for GetTransportError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            GetTransportError::CertError(msg) => {
+                write!(f, "ran into an error retrieving certificate: {}", msg)
+            }
+            GetTransportError::NotSupportedError(msg) => {
+                write!(f, "recieved transport type that is not supported: {}", msg)
```
peterschwarz: recieved -> received

splinterd/src/main.rs
```diff
@@ -606,15 +616,126 @@ fn get_transport(
...
+            }
+            GetTransportError::TlsTransportError(err) => {
+                write!(f, "unable to create TLS transport: {:?}", err)
```
peterschwarz: Do these need to be debug output?

splinterd/src/daemon.rs
agunde406: updated

splinterd/src/daemon.rs
agunde406: fixed

splinterd/src/main.rs
agunde406: fixed

splinterd/src/main.rs
agunde406: It used to, not anymore. Fixed

splinterd/src/main.rs
```diff
@@ -609,6 +609,80 @@ fn get_transport(
...
+            }
+        }
+    }
+}
+
+impl fmt::Display for UserError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            UserError::TransportError(err) => write!(f, "unable to get transport: {}", err),
+            UserError::MissingArgument(msg) => write!(f, "missing requried argument: {}", msg),
```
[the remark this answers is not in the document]
agunde406: fixed

PR #438, opened 2020-01-20
Initial tests for the Config module.

shannynalayna:
Adds some initial tests for the Config module. This includes tests for the Config, ConfigBuilder and TomlConfig objects.

In order to run all tests (which are all from various features) run each of the following: 
- `cargo test` should result in only the first test being run 
- `cargo test --features config-builder` should run the first 3 tests
- `cargo test --features config-toml` should run the last 4 tests

### DIFF
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -69,6 +69,10 @@ impl ConfigBuilder {
         self
     }
 
+    // In testing for the ConfigBuilder module behind the config-builder feature, this method is
+    // not used in order to produce a None option in the resulting ConfigBuilder object. This
+    // method is required to compile and test the ConfigToml module behind the config-toml feature.
+    #[cfg(feature = "config-toml")]
     pub fn with_cert_dir(mut self, cert_dir: String) -> Self {
         self.cert_dir = Some(cert_dir);
         self
@@ -119,6 +123,10 @@ impl ConfigBuilder {
         self
     }
 
+    // In testing for the ConfigBuilder module behind the config-builder feature, this method is
+    // not used in order to produce a None option in the resulting ConfigBuilder object. This
+    // method is required to compile and test the ConfigToml module behind the config-toml feature.
+    #[cfg(feature = "config-toml")]
     pub fn with_bind(mut self, bind: String) -> Self {
         self.bind = Some(bind);
         self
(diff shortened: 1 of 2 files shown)


splinterd/src/config/mod.rs
```diff
@@ -134,3 +134,305 @@ impl Config {
...
+            registry_backend: None,
+            registry_file: None,
+            heartbeat_interval: None,
+        }
+    }
+
+    /// Creates a Config struct based on the values from the TEST_TOML2 file.
+    /// This function is only used in testing the TomlConfig module and therefore only used with
+    /// the `toml-config` feature.
+    #[allow(dead_code)]
```
vaporos: This should be protected with the toml-config feature then, instead of marked dead-code.

splinterd/src/config/mod.rs
```diff
@@ -134,3 +134,305 @@ impl Config {
...
+    static SERVER_KEY: &str = "certs/server.key";
+
+    /// Config values unique to the TEST_TOML1 file.
+    static SERVICE_ENDPOINT: &str = "127.0.0.1:8043";
+    static NETWORK_ENDPOINT: &str = "127.0.0.1:8044";
+    static NODE_ID: &str = "012";
+
+    /// Config values unique to the TEST_TOML2 file.
+    /// These values are only used in testing the TomlConfig module and therefore only used with
+    /// the `toml-config` feature.
```
vaporos: These should be guarded by the toml-config feature then instead of marked dead-code.

splinterd/src/config/builder.rs
```diff
@@ -69,6 +69,9 @@ impl ConfigBuilder {
         self
     }
 
+    /// In testing for the ConfigBuilder module behind the config-builder feature, this method is
+    /// not used in order to produce a None option in the resulting ConfigBuilder object.
+    #[allow(dead_code)]
```
vaporos: Should we guard these with a testing conditional instead of marking them with dead-code?

Also, these comments should probably not be doc strings.

splinterd/src/config/mod.rs

splinterd/src/config/mod.rs
shannynalayna: Addressed

splinterd/src/config/builder.rs
shannynalayna: Addressed.

splinterd/src/config/mod.rs
```diff
@@ -134,3 +134,305 @@ impl Config {
...
+        let built_config = config_builder.build();
+        // Construct an example Config object.
+        let example_config = construct_config_example();
+        // Compare the generated Config object against a manually constructed example Config object.
+        compare_configs(built_config, example_config);
+    }
+
+    #[cfg(feature = "config-toml")]
+    #[test]
+    /// This test verifies that a Config object, consructed from the TomlConfig module, contains
```
vaporos: Typo - consructed

shannynalayna: Addressed

shannynalayna:
Closing in favor of #456.

PR #439, opened 2020-01-21
Add enable-biome splinterd flag

peterschwarz:
Add the CLI flag --enable-biome to the splinter CLI.  The addition of
this flag relaxes the requirement that the database-url must be set, as
it is currently only required to satisfy the biome subsystem's needs.

It adds the following constraints:

- If the flag is set, the db-url is also required.
- If unset, the db-url is optional (and, in fact, has no effect, given
  that there is no other subsystem that uses it).

### DIFF
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -105,7 +105,9 @@ pub struct SplinterDaemon {
     node_id: String,
     rest_api_endpoint: String,
     #[cfg(feature = "database")]
-    db_url: String,
+    db_url: Option<String>,
+    #[cfg(feature = "biome")]
+    biome_enabled: bool,
     registry_config: RegistryConfig,
     storage_type: String,
 }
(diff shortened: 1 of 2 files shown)


dplumb94:
This would require a rework as soon as we have another module that needs database functionality (node registry or key registry for example). It also seems a bit odd that you have to compile with the biome feature _and_ set an enabled flag on startup to enable biome. What would happen if you compile with biome and set `db-url` but dont pass the `--enable-biome` flag?

What about `--enable-db`?

peterschwarz:
@dplumb94 The `enable-biome` flag is orthogonal to the database issue.  Once we implement other things that require a database, but are not optional (splinter state, for example), it would require re-strengthening the need for a database configuration parameter.  

Using a binary with biome compiled in (say from an official, general distribution), a user may still be interested in opting out of biome, but will want to use the database for other parts of the system

peterschwarz:
Rebased and removed an `expect`

PR #445, opened 2020-01-22
Default Config object

shannynalayna:
Adds an initial DefaultConfig object behind a `config-default` feature flag.

### DIFF
--- a/splinterd/src/config/default.rs
+++ b/splinterd/src/config/default.rs
@@ -118,3 +118,79 @@ impl PartialConfigBuilder for DefaultConfig {
         return partial_config.with_database(self.database);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use std::time::Duration;
+
+    /// Asserts config values based on the default values.
+    fn assert_default_values(config: PartialConfig) {
+        assert_eq!(config.storage(), Some(String::from("yaml")));
+        assert_eq!(config.transport(), Some(String::from("raw")));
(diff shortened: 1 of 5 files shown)


peterschwarz:
This is now built off of #456, so it will have to wait for review until that is merged.

splinterd/src/config/default.rs
```diff
@@ -0,0 +1,112 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::Path;
+
+use crate::config::{PartialConfig, PartialConfigBuilder};
+use crate::{
```
vaporos: Localize these here; it is the only place the defaults should exist after we are done with the config changes. (This probably means having them in both places temporarily.)

splinterd/src/config/mod.rs
```diff
@@ -194,4 +257,24 @@ mod tests {
         // Compare the generated PartialConfig object against the expected values.
         assert_config_values(built_config);
     }
+
+    #[cfg(feature = "config-default")]
```
vaporos: This should go in the config::default::test module.

splinterd/src/config/default.rs
shannynalayna: Addressed. I re-imported them in main from the config module to avoid compiler warnings, which required further cfg attributes, but (as you said in your comment) this should be temporary.

PR #446, opened 2020-01-22
Fix wait for batch status

peterschwarz:
Scabbard does not prefix the batch link with the scheme and host.  This causes the batches to always fail as the batch link cannot be queried. This corrects the client such that the link will have the base url added if missing.

Additionally:

- Adds more debug statements

### DIFF
--- a/libsplinter/src/service/scabbard/client/submit.rs
+++ b/libsplinter/src/service/scabbard/client/submit.rs
@@ -41,6 +41,7 @@ pub fn submit_batches(
 
     let body = batches.write_to_bytes()?;
 
+    debug!("Submitting batches via {}", url);
     let request = Client::new().post(url).body(body);
     let response = perform_request(request)?;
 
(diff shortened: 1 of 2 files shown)


libsplinter/src/service/scabbard/client/submit.rs
```diff
@@ -51,9 +52,14 @@ pub fn submit_batches(
     Ok(batch_link.link)
 }
 
-pub fn wait_for_batches(url: &str, wait: u64) -> Result<(), Error> {
-    let url = parse_http_url(&format!("{}&wait={}", url, wait))?;
+pub fn wait_for_batches(base_url: &str, batch_link: &str, wait: u64) -> Result<(), Error> {
+    let url = if batch_link.starts_with("http") || batch_link.starts_with("http") {
```
agunde406: are these supposed to both say `batch_link.starts_with("http")` ?

PR #453, opened 2020-01-27
Add commands to support circuit creation

eloaverona:
PR adds the commands bellow that will support creating circuits via cli
- `splinter node alias`:
   -  `add <node-alias> <node-endpoint>`: add new alias to an node endpoint
   -  `get <node-alias>`: see endpoint for alias
   -  `list`: list all aliases and endpoints
   -  `delete <node-alias>`: delete an alias

For more context on how these command can be used refer to [Shawn's comment](https://zenhub.cglcloud.com/app/workspaces/product-core-5d37298b0add7306b30efde2/issues/digitalsupplychain/product-core/1743#issuecomment-110397).

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -62,7 +62,7 @@ experimental = [
     "keygen",
 ]
 
-circuit = ["reqwest", "serde_json", "splinter/sawtooth-signing-compat"]
+circuit = ["reqwest", "serde_json", "splinter/sawtooth-signing-compat", "dirs"]
 health = ["reqwest", "serde_json"]
 keygen = ["dirs", "whoami"]
 
--- a/cli/src/action/mod.rs
+++ b/cli/src/action/mod.rs
@@ -22,6 +22,8 @@ pub mod database;
 pub mod health;
 #[cfg(feature = "keygen")]
 pub mod keygen;
+#[cfg(feature = "circuit")]
+pub mod node;
 
 use std::collections::HashMap;
 use std::ffi::CString;
--- a/cli/src/action/node.rs
+++ b/cli/src/action/node.rs
@@ -51,6 +51,30 @@ impl Action for AddNodeAliasAction {
     }
 }
 
+pub struct GetNodeAliasAction;
+
+impl Action for GetNodeAliasAction {
+    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+        let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
+        let alias = match args.value_of("alias") {
+            Some(alias) => alias,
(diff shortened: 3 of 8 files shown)


vaporos:
Can we split this PR into two (it is very large now) - this one could be just to node code (which looks smaller), with a follow-up being the circuit default.

cli/src/manager/node/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+mod error;
+
+use reqwest::Url;
+
+use crate::store::node::{FileBackedNodeStore, NodeStore};
+
+pub use crate::store::node::Node;
+pub use error::NodeManagerError;
+
+pub struct NodeManager {
```
vaporos: There is a lot of abstraction here, why do we need to wrap NodeStore in a NodeManager?

cli/src/store/yaml_store.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::fs::{self, File, OpenOptions};
+use std::path::PathBuf;
+
+use serde::{de::DeserializeOwned, Serialize};
+
+use super::YamlBackedStoreError;
+
+const DEFAULTS_PATH: &str = ".splinter";
+
+pub(super) trait YamlBackedStore<T: Serialize + DeserializeOwned> {
```
vaporos: This abstraction seems unnecessary. Just using serde directly by the file-backed stores would be a lot less code. The stores basically just need a load() and save() methods which should be a subset of this code to accomplish.

cli/src/manager/node/mod.rs
eloaverona: Not strictly necessary, I debated with myself about this, but the idea was that the CLI commands would no nothing of the backend being used. If there is no node manager or something of the like, the CLI will have to initialize an particular implementation of the NodeStore. 

The other part of it was that not the CLI nor the Store would know about data validation. The NodeStore would just know how to store the data, and load data, and any data validation is responsibility of the NodeManager.

eloaverona: The DefaultValueManager, might be a better example of any value this abstraction might offer

cli/src/store/yaml_store.rs
eloaverona: Yeah, I was also a little unsure about this. The idea was to avoid code duplication, as any store yaml backed store (in this PR both FileBackedNodeStore and FileBackedDefaultValueStore) would not have to implement the same logic again. And I wanted to avoid creating an `utils` like file with some loose methods that both stores use. 

As to the point that `the stores basically just need a load() and save() `, its easy to refactor the code to look more like that. I think the stores just directly use the `read_data_from_file` (like load()) and `write_to_file`(like save()). The other methods are working more as helper methods.

PR #456, opened 2020-01-27
Partial Config object

shannynalayna:
Replaces the current `Config` object with a `PartialConfig` and a `PartialConfigBuilder` which is the first step in a process of converting splinterd's use of the Config object which held options to a new `Config` object without options (thus cleaning up the `unwrap_or`s in the main function) 

This is just the first step in that process, which converts and removes the existing Config and ConfigBuilder objects, as these are going to be rewritten to match the description above ^. This work will be in a later PR. 

For testing: 

Run `cargo test` to run the first test, `cargo test --features config-builder` to run the first three tests and `cargo test --features config-toml` to run the last test.


splinterd/src/config/partial.rs
```diff
@@ -0,0 +1,338 @@
...
+#[cfg(not(feature = "config-toml"))]
+use std::io::Read;
+
+#[cfg(not(feature = "config-toml"))]
+use toml;
+
+#[cfg(not(feature = "config-toml"))]
+use crate::config::ConfigError;
+
+/// PartialConfig is an object to be used as an intermediary while collecting the config values
```
peterschwarz: Would suggest "PartialConfig is an intermediate representation of configuration values, used when combining several sources." as the opening sentence.

shannynalayna: That's definitely better wording, I changed the sentence to your suggestion.

splinterd/src/config/partial.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    #[cfg(feature = "config-builder")]
+    /// Adds a `storage` value to the PartialConfig object.
+    ///
+    /// # Arguments
+    ///
+    /// * `storage` - The type of storage that should be used to store circuit state.
+    ///
+    pub fn with_storage(mut self, storage: Option<String>) -> Self {
```
RyanLassigBanks: It seems weird to me that these would take an `Option<String>` instead of just `String`. Is the idea that you would want to overwrite a field in a config file with `None`?

shannynalayna: Well, this really just initializing the PartialConfig object with a None, possibly. Currently, these methods are being used to directly pass the value from the config object into the PartialConfig. So, it made sense to me to pass the entire thing as those config objects will all hold options (and match the PartialConfig field). Rather than unwrapping those objects, this makes it simpler to simply pass the value along.

shannynalayna: If you look here: https://github.com/Cargill/splinter/pull/456/files#diff-e81ee9df252e4ba07221e9155a7d6f9cR49-R67, you can see how these methods are being used. So, rather than unwrap those values which are automatically loaded into the TomlConfig object, it seemed cleaner to me to pass the entire option in

splinterd/src/config/partial.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    pub fn registry_file(&self) -> Option<String> {
+        self.registry_file.clone()
+    }
+
+    pub fn heartbeat_interval(&self) -> Option<u64> {
+        self.heartbeat_interval
+    }
+
+    #[cfg(feature = "config-builder")]
```
dplumb94: Is attributing each method with cfg an established style? I think it would be cleaner to wrap all of the functions for each feature in a block and then just apply the attribute to the block.

splinterd/src/config/partial.rs
```diff
@@ -0,0 +1,338 @@
...
+    database: Option<String>,
+    registry_backend: Option<String>,
+    registry_file: Option<String>,
+    heartbeat_interval: Option<u64>,
+}
+
+impl PartialConfig {
+    /// Creates a new PartialConfig object from a toml file. Available to use when the `config-toml`
+    /// feature flag is not in use.
+    #[cfg(not(feature = "config-toml"))]
```
vaporos: This should be in a toml module that implements the TomlPartialConfigBuilder, not part of PartialConfig (PartialConfig shouldn't know anything about implementations).

splinterd/src/config/partial.rs
vaporos: The disadvantage of that is that if you remove the cfg, then it would change the intent (larger future diff). So the current approach is actually good.

I assume these are here because otherwise we get compiler warnings.

splinterd/src/config/partial.rs
vaporos: Yes, None would be "unset this value". Seems good to me.

splinterd/src/main.rs
```diff
@@ -84,32 +84,28 @@ fn load_toml_config(config_file_path: &str) -> Config {
...
-        Config::default()
+        PartialConfig::default()
     })
 }
 
 #[cfg(feature = "config-toml")]
-fn load_toml_config(config_file_path: &str) -> Config {
-    let mut config_builder = ConfigBuilder::new();
-
+fn load_toml_config(config_file_path: &str) -> PartialConfig {
```
vaporos: Instead of returning a default, it may be better here to return None and the calling code could then avoid using it in the merge.

shannynalayna: I don't think this is a necessary change for this PR, because in the case that it would return None and the `toml-config` feature is enabled, the main function still expects a `config` object, which would necessitate creating a default config in the main function rather than here.

splinterd/src/config/partial.rs
shannynalayna: The compiler warnings assumption is correct.

splinterd/src/main.rs
vaporos: Ok, agreed, this probably makes more sense when we are collecting a list of PartialConfigs to merge into a Config.

PR #458, opened 2020-01-27
Initial register/login sapling

zapplebee:
Looking for comments. Until I can test this locally end-to-end, I want to make sure this stays a draft.

From the `canopy/saplings/register-login` directory, after installing dependencies one can run
`npm run deploy` to build the register/login sapling and it will automatically be added to the appropriate directory in `canopy/app`.

Specifically looking for feedback during this draft around:
1. The aforementioned `deploy` strategy
2. If the example is simple enough from an instructional perspective. My thinking is that our example saplings are meant to be demonstrative as to how to use CanopyJS.


dplumb94:
Implementation is in the Grid repo

PR #460, opened 2020-01-28
Initial Optional Protocol Header

peterschwarz:
This introduces the trait RequestGuard for the REST API Resource classes. It also adds a ProtocolVersionRangeGuard implementation of this trait.

The protocol guard looks for an optional header, `SplinterProtocolVersion`.  If this is provided, it will return a BadRequest if the version provide is out of range.

It adds this guard to the admin routes, with the remaining routes to be guarded in a future PR.

To test (use  with experimental):

`curl --verbose -s -H "SplinterProtocolVersion: 1" localhost:8088/admin/proposals`

The above should result in a 200 and empty list

Modify the value of the header, either `0`, or greater than `1` and the 400 error with a json error response will be returned.

### DIFF
--- a/libsplinter/src/admin/rest_api/mod.rs
+++ b/libsplinter/src/admin/rest_api/mod.rs
@@ -22,10 +22,11 @@ use std::time;
 
 use crate::actix_web::{web, HttpResponse};
 use crate::futures::{Future, IntoFuture};
+use crate::protocol;
 use crate::protos::admin::CircuitManagementPayload;
 use crate::rest_api::{
-    into_protobuf, new_websocket_event_sender, EventSender, Method, Request, Resource,
-    RestResourceProvider,
+    into_protobuf, new_websocket_event_sender, EventSender, Method, ProtocolVersionRangeGuard,
+    Request, Resource, RestResourceProvider,
 };
 use crate::service::ServiceError;
 
@@ -40,12 +41,10 @@ use super::service::{
 
 impl RestResourceProvider for AdminService {
     fn resources(&self) -> Vec<Resource> {
-        // Allowing unused_mut because resources must be mutable if feature proposal-read is enabled
-        #[allow(unused_mut)]
-        let mut resources = vec![
-            make_application_handler_registration_route(self.commands()),
-            make_submit_route(self.commands()),
-        ];
+        let mut resources = vec![];
+
+        resources.push(make_application_handler_registration_route(self.commands()));
+        resources.push(make_submit_route(self.commands()));
 
         #[cfg(feature = "proposal-read")]
         resources.push(make_fetch_proposal_resource(self.commands()));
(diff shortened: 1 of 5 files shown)


libsplinter/src/rest_api/mod.rs
```diff
@@ -330,6 +330,93 @@ impl RequestGuard for Box<dyn RequestGuard> {
...
+impl ProtocolVersionRangeGuard {
+    /// Constructs a new protocol version guard with the given minimum.
+    pub fn new(min: u32, max: u32) -> Self {
+        Self { min, max }
+    }
+}
+
+impl RequestGuard for ProtocolVersionRangeGuard {
+    fn evaluate(&self, req: &HttpRequest) -> Continuation {
+        if let Some(header_value) = req.headers().get("X-SplinterProtocolVersion") {
```
ltseeley: May not want to include the `X-` in the header name: https://stackoverflow.com/a/3561399

libsplinter/src/rest_api/mod.rs
```diff
@@ -330,6 +330,93 @@ impl RequestGuard for Box<dyn RequestGuard> {
...
+                    format!(
+                        "Invalid characters in X-SplinterProtocolVersion header: {}",
+                        err
+                    )
+                })
+                .and_then(|val_str| {
+                    val_str.parse::<u32>().map_err(|_| {
+                        format!("X-SplinterProtocolVersion must be a valid positive integer")
+                    })
+                }) {
```
ltseeley: This is a long statement to match on; would be easier to read if declared as a variable first.

libsplinter/src/rest_api/mod.rs
```diff
@@ -330,6 +330,93 @@ impl RequestGuard for Box<dyn RequestGuard> {
...
+                            )
+                        }))
+                        .into_future(),
+                ),
+                Ok(version) if version > self.max => Continuation::terminate(
+                    HttpResponse::BadRequest()
+                        .json(json!({
+                            "message": format!(
+                                "Client requires a newer protocol than provided: {} > {}",
+                                version,
```
ltseeley: This error message is actually a little difficult to make sense of at first; it might be more clear as "Client's requested protocol ({}) is not supported; supported protocols: {}" or something similar. This would also allow you to collapse these two match arms into one.

libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -22,10 +22,11 @@ use std::time;
 
 use crate::actix_web::{web, HttpResponse};
 use crate::futures::{Future, IntoFuture};
+use crate::protocol;
```
ltseeley: Have stories been added to add the guard to the other routes as well?

libsplinter/src/lib.rs
```diff
@@ -75,6 +75,8 @@ pub mod mesh;
 pub mod network;
 pub mod node_registry;
 pub mod orchestrator;
+#[cfg(any(feature = "rest-api", feature = "scabbard-client"))]
+mod protocol;
```
ltseeley: Would it make more sense for the protocol versions to be defined locally for the various REST APIs and clients? It seems odd for these to be centralized when the REST APIs and the clients themselves aren't.

libsplinter/src/admin/rest_api/mod.rs
peterschwarz: This is not the complete solution, with respect to guarding the routes. It is merely the mechanism to do so.  The task in question will be completed with a follow-on PR

libsplinter/src/lib.rs
peterschwarz: The location of these items has yet to be determined, but it is definitely something to be considered (though this approach was based on initial conversations about this feature).

libsplinter/src/rest_api/mod.rs
peterschwarz: Good catch, will update

peterschwarz: Good catch, will update

peterschwarz: Good catch, will update

peterschwarz: Good catch, will update

libsplinter/src/rest_api/mod.rs
peterschwarz: They signify very different problems - one is that your protocol is too old, and you should upgrade your client, the other is that your server is too old to work with the protocol in question, so you have to possibly talk to a sysadmin to fix the issue.

libsplinter/src/rest_api/mod.rs
peterschwarz: Fixed

libsplinter/src/rest_api/mod.rs
peterschwarz: Fixed

libsplinter/src/rest_api/mod.rs
```diff
@@ -219,11 +221,63 @@ impl Resource {
         self
     }
 
+    /// Adds a RequestGuard to the given Resource.
+    ///
+    /// This guard applies to all methods in defined for this resource.
```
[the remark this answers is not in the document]
peterschwarz: Fixed

libsplinter/src/rest_api/mod.rs
```diff
@@ -330,6 +330,94 @@ impl RequestGuard for Box<dyn RequestGuard> {
     }
 }
 
+/// Guards requests based on a minimum protocol version.
+///
+/// A protocol version is specified via the HTTP header `"SplinterProtocolVersion"`.  This header
```
ltseeley: Commit message still has `X-` header prefix

splinterd/api/static/openapi.yml
```diff
@@ -599,6 +618,14 @@ paths:
         500:
           description: Internal service error
 
+parameters:
+  protocol_version:
+    type: integer
+    description: The protocol version which the client can understand
+    in: header
+    name: X-SplinterProtocolVersion
```
ltseeley: Still has `X-` prefix

libsplinter/src/rest_api/mod.rs
peterschwarz: Fixed

splinterd/api/static/openapi.yml
peterschwarz: Fixed

PR #454, opened 2020-01-27
Scabbard CLI

ltseeley:
To test:
- Run the Grid on Splinter example and setup a circuit
- Download the xo smart contract (https://build.sawtooth.me/job/Sawtooth-Hyperledger/job/sawtooth-sdk-rust/job/master/lastSuccessfulBuild/artifact/build/scar/xo_0.4.1.scar)
- Add the xo smart contract's directory as a volume for the scabbard-cli service in docker-compose-installed.yaml (mapped to `/usr/share/scar`)
- Run the scabbard-cli image using `docker-compose -f docker-compose-installed.yaml run scabbard-cli bash`
- In the scabbard-cli container:
  - `mkdir -p /root/.splinter/keys`
  - `echo "<alpha-gridd-pub-key>" > /root/.splinter/keys/gridd.pub`
  - `echo "<alpha-gridd-priv-key>" > /root/.splinter/keys/gridd.priv`
  - `scabbard cr create sawtooth_xo --owner $(cat /root/.splinter/keys/gridd.pub) -k gridd -U 'http://splinterd-alpha:8085' --service-id 'my-grid-circuit::grid-scabbard-a'`
  - `scabbard upload /usr/share/scar/xo_0.4.1.scar -k gridd -U 'http://splinterd-alpha:8085' --service-id 'my-grid-circuit::grid-scabbard-a'`
  - `scabbard ns create 5b7349 --owner $(cat /root/.splinter/keys/gridd.pub) -k gridd -U 'http://splinterd-alpha:8085' --service-id 'my-grid-circuit::grid-scabbard-a'`
  - `scabbard perm 5b7349 sawtooth_xo -r -w -k gridd -U 'http://splinterd-alpha:8085' --service-id 'my-grid-circuit::grid-scabbard-a'`
  - `scabbard perm cad11d sawtooth_xo -r -k gridd -U 'http://splinterd-alpha:8085' --service-id 'my-grid-circuit::grid-scabbard-a'`

There is currently no way to verify the success of these actions besides the proposals being committed (seen in the logs).

PR #462, opened 2020-01-28
Initialize scabbard crate

ltseeley:
Initialize the `services/scabbard` crate that will contain all code
related to the scabbard service.

The single commit in this PR is broken out from #454 to work around an issue with the workspace Cargo.toml.


ltseeley:
Closing because this doesn't do what was intended

PR #463, opened 2020-01-28
Update design system for Grid theme

dnewh:
This updates the Canopy design system with some functional updates as well as stylistic ones. This PR also includes changes to the Canopy UI to implement the changes in the design system.


canopy/app/src/images/grid.svg
```diff
@@ -0,0 +1,16 @@
+<svg width="80" height="80" viewBox="0 0 80 80" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dplumb94: I don't think the grid logo should be in here. The grid UI should supply it in the Grid repo.

dnewh: removed

canopy/canopyjs/src/components/navigation/SideNav.js
```diff
@@ -0,0 +1,77 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import classnames from 'classnames';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { CanopyContext } from '../../CanopyContext';
```
dplumb94: Instead of this, you can just `import { useUserSaplings } from '../../CanopyContext'` and then in the component `const userSaplings = useUserSaplings()`. This way you can remove that `CanopyContext.Consumer` wrapper.

dnewh: updated

PR #468, opened 2020-01-30
Scabbard get state

ltseeley:
Add the `get_state_at_address` and `get_state_with_prefix` methods to
scabbard to enable getting the value at a specific address in state, and
all values under a given address prefix in state (respectively).

Add REST API endpoints for getting state from a scabbard service.

### DIFF
--- a/libsplinter/src/service/scabbard/rest_api.rs
+++ b/libsplinter/src/service/scabbard/rest_api.rs
@@ -227,6 +227,110 @@ pub fn make_get_batch_status_endpoint() -> ServiceEndpoint {
     }
 }
 
+#[cfg(feature = "scabbard-get-state")]
+pub fn make_get_state_at_address_endpoint() -> ServiceEndpoint {
+    ServiceEndpoint {
+        service_type: SERVICE_TYPE.into(),
+        route: "/state/{address}".into(),
+        method: Method::Get,
+        handler: Arc::new(move |request, _, service| {
+            let scabbard = match service.as_any().downcast_ref::<Scabbard>() {
+                Some(s) => s,
+                None => {
+                    return Box::new(
+                        HttpResponse::InternalServerError()
+                            .json(json!({
+                                "message": "service is not Scabbard"
+                            }))
+                            .into_future(),
+                    )
+                }
+            };
+
+            let address = request
+                .match_info()
+                .get("address")
+                .expect("address should not be none");
+
+            Box::new(match scabbard.get_state_at_address(address) {
+                Ok(Some(value)) => HttpResponse::Ok().json(value).into_future(),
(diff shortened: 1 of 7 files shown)


libsplinter/src/service/scabbard/rest_api.rs
```diff
@@ -227,6 +227,102 @@ pub fn make_get_batch_status_endpoint() -> ServiceEndpoint {
...
+                    let res = state_iter
+                        .map(|res| {
+                            res.map(|(address, value)| {
+                                json!({
+                                    "address": address,
+                                    "value": value,
+                                })
+                            })
+                        })
+                        .collect::<Result<Vec<_>, _>>();
```
ltseeley: This doesn't currently support paging through the state entries because that isn't supported by transact.

peterschwarz: Mainly, this is due to the fact that it is currently difficult to find a starting point for a leaf, but I have some ideas for that in transact.

arsulegai: By adding more meta information to the tree nodes?

splinterd/api/static/openapi.yml
```diff
@@ -599,6 +599,106 @@ paths:
...
+            example: 000000a87cb5eafdcca6a814e4add97c4b517d3c530c2f44b31d18e3b0c44298fc1c14
+      responses:
+        200:
+          description: The value at the requested address
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: integer
```
arsulegai: Is this `array` of `integer` a `type: string` of `format: byte`?

ltseeley: I don't think so; it's an actual JSON array of integers. I'm not sure if this is the correct way to represent that in the OpenAPI spec, though.

libsplinter/src/service/scabbard/rest_api.rs
peterschwarz: I don't think that's needed, but I think some simple calculations can be made based on an address to find where to start in the tree.  Most of the paging in sawtooth initially relied on a  position in the result set.  This was simple, and allowed users to create the page links without any knowledge of the underlying system.  In order to actually optimize these queries, the "next" item needs to really be specific to the store being iterated over.  A good example of this was the change in block iteration, where it uses a representation of block height for iteratation.  

For state, we could use the combination of `(state root, Option<Subtree>, address)` as the location in the iterator.  This can get us a starting point in the tree, and iterate over the leaves from there.

peterschwarz: In order to implement this, we need to implement some changes in [hyperledger/transact](https://github.com/hyperledger/transact) first.

PR #475, opened 2020-01-31
Add contract list/show subcommands to scabbard CLI 

ltseeley:


services/scabbard/src/cli/main.rs
```diff
@@ -748,6 +814,107 @@ fn run() -> Result<(), CliError> {
...
+                let service_id = matches
+                    .value_of("service-id")
+                    .ok_or_else(|| CliError::MissingArgument("service-id".into()))?;
+
+                let format = matches
+                    .value_of("format")
+                    .expect("default not set for --format");
+
+                let registries = client
+                    .get_state_with_prefix(circuit, service_id, Some("00ec01"))?
```
ltseeley: Ultimately, this namespace should be defined as a constant in the Sabre SDK and just pulled in by the CLI.

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -50,4 +53,92 @@ impl ScabbardClient {
             Ok(())
         }
     }
+
+    pub fn get_state_at_address(
+        &self,
+        circuit_id: &str,
+        service_id: &str,
```
vaporos: If we create a struct "ServiceId" here, it could include both circuit_id and service_id. That struct could enforce correctness (validate that they appear to be valid from a length and character set perspective), making it safer input into these functions.

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -50,4 +53,92 @@ impl ScabbardClient {
...
+                status, msg
+            )))
+        }
+    }
+
+    pub fn get_state_with_prefix(
+        &self,
+        circuit_id: &str,
+        service_id: &str,
+        prefix: Option<&str>,
```
vaporos: We should validate the prefix is an appropriate character set (hex) and set an upper bound on length (70) before constructing the URL. We could also verify here that it is an even number of characters.

libsplinter/src/service/scabbard/client/mod.rs
```diff
@@ -50,4 +53,92 @@ impl ScabbardClient {
...
+            Err(Error::new(&format!(
+                "failed to get state with prefix: {}: {}",
+                status, msg
+            )))
+        }
+    }
+}
+
+#[derive(Deserialize, Debug)]
+pub struct StateEntry {
```
vaporos: If we made the fields internal, we could enforce address to be well-formed.

libsplinter/src/service/scabbard/client/mod.rs
ltseeley: Add the `ServiceId` struct.

libsplinter/src/service/scabbard/client/mod.rs
ltseeley: Added validation for the prefix here (valid hex, <70 characters) and added validation for the get_state_by_address method's address as well (just valid hex).

libsplinter/src/service/scabbard/client/mod.rs
ltseeley: Made the fields provided, accessible by methods.

vaporos: To realize this, we would need a constructor which enforces address be well-formed.

ltseeley:
> This particular error should probably be logged as error, as it is very serious.
> 
> From a security perspective, I’m not sure that I would send this as a message to the client as it may reveal information about the implementation that could lead to exploits.

Updated

ltseeley:
> For any of these that are returning a 500 error, I would suggest using a canned message about “internal server error, contact your sys admin”

Updated

ltseeley:
> I noticed the directory layout for services/scabbard seems a bit wonky. It has an empty `lib.rs` file and a `main.rs` file in the cli directory. I think the empty lib.rs file should be removed, but I think that's out of scope for this PR.

This structure is intentional. `lib.rs` is a placeholder; eventually, we'll move the scabbard service itself there.

PR #476, opened 2020-02-03
Add rest-api feature to Cargo toml

RyanLassigBanks:

### DIFF
--- a/services/health/Cargo.toml
+++ b/services/health/Cargo.toml
@@ -26,7 +26,7 @@ description = """\
 
 [dependencies]
 log = "0.3.0"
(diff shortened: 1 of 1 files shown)


RyanLassigBanks:
> Could you add some context to this commit?

I feel the context is fairly obvious, but nevertheless updated

vaporos:
> > Could you add some context to this commit?
> 
> I feel the context is fairly obvious, but nevertheless updated

Not really obvious when looking through commit history 6 months from now. From that perspective, we would want to capture what we know now (that we are fixing a bug introduced in a commit which made the rest api off by default).

PR #478, opened 2020-02-03
Add command to set default value for service type and management type

eloaverona:
PR adds the commands bellow that will support creating circuits via cli
- `splinter circuit default`:
  - `set service-type <service-type>`: set a default service-type, the default value will be used during circuit creation if no service type is passed.
  - `set management-type <management-type>`: set a default management-type, the default value will be used during circuit creation if no management type is passed.
  - ` get service-type`: get the default value for service-type
  - `get management-type`: get the default value for management-type
  - `list`: list all defaults
  - `unset service-type` unset the default value for service-type
  - `unset management-type`: unset the default value for management-type

For more context on how these command can be used refer to [Shawn's comment](https://zenhub.cglcloud.com/app/workspaces/product-core-5d37298b0add7306b30efde2/issues/digitalsupplychain/product-core/1743#issuecomment-110397).

### DIFF
--- a/cli/src/action/circuit/defaults.rs
+++ b/cli/src/action/circuit/defaults.rs
@@ -1,0 +1,80 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::ArgMatches;
+
+use crate::error::CliError;
+use crate::store::default_value::{
+    DefaultStoreError, DefaultValue, DefaultValueStore, FileBackedDefaultStore,
+};
+
+const MANAGEMENT_TYPE_KEY: &str = "management_type";
+const SERVICE_TYPE_KEY: &str = "service_type";
+
+use super::Action;
+
+pub struct SetDefaultValueAction;
+
+impl Action for SetDefaultValueAction {
+    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+        let args = arg_matches.ok_or_else(|| CliError::RequiresArgs)?;
+
+        let name = match args.value_of("name") {
+            Some(key) => key,
+            None => return Err(CliError::ActionError("name is required".into())),
+        };
+
+        let key = get_key(name)?;
+
+        let value = match args.value_of("value") {
+            Some(value) => value,
+            None => return Err(CliError::ActionError("value is required".into())),
+        };
+
+        let store = get_default_value_store();
+
+        if !args.is_present("force") && store.get_default_value(key)?.is_some() {
+            return Err(CliError::ActionError(format!(
+                "Default value for {} is already in use",
+                key
+            )));
+        }
+
+        let default_value = DefaultValue::new(key, value);
+        store.set_default_value(&default_value)?;
+
+        Ok(())
+    }
+}
+
+fn get_key(name: &str) -> Result<&str, CliError> {
+    match name {
+        "service-type" => Ok(SERVICE_TYPE_KEY),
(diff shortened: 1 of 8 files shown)


cli/src/store/default_value/yaml_store.rs
```diff
@@ -0,0 +1,181 @@
...
+        if let Some(index) = existing_default_index {
+            defaults.remove(index);
+        }
+
+        defaults.push(SerdeDefaultValue::from(new_default_value));
+
+        self.save(&defaults)?;
+
+        Ok(())
+    }
```
agunde406: Can you add a blank line here

cli/src/main.rs
```diff
@@ -319,6 +319,30 @@ fn run() -> Result<(), CliError> {
...
+                        .about("Manage default values for circuit creation")
+                        .subcommand(
+                            SubCommand::with_name("set")
+                                .about("Set a default value")
+                                .setting(AppSettings::SubcommandRequiredElseHelp)
+                                .subcommand(
+                                    SubCommand::with_name("service-type")
+                                        .about("Set default value for service type")
+                                        .arg(
+                                            Arg::with_name("service_type")
```
agunde406: This should have a `.value_name("service-type")`

cli/src/main.rs
```diff
@@ -341,6 +341,23 @@ fn run() -> Result<(), CliError> {
                                                 .long("force")
                                                 .help("Overwrite default if it is already set"),
                                         ),
+                                )
+                                .subcommand(
+                                    SubCommand::with_name("management-type")
+                                        .about("Set default value for management type")
+                                        .arg(
+                                            Arg::with_name("management_type")
```
agunde406: Needs `.value_name("management-type")`

cli/src/action/circuit/defaults.rs
```diff
@@ -51,6 +51,16 @@ impl Action for SetServiceTypeDefaultAction {
     }
 }
 
+pub struct UnsetServiceTypeDefaultAction;
```
agunde406: Typo in commit message: commant -> command

cli/src/main.rs
```diff
@@ -373,6 +373,15 @@ fn run() -> Result<(), CliError> {
                                         .about("Unset default value for management type"),
                                 ),
                         )
+                        .subcommand(
+                            SubCommand::with_name("get")
```
agunde406: Any reason we are doing "get" here instead of "show"?

peterschwarz: It probably should be

cli/src/main.rs
```diff
@@ -319,6 +319,30 @@ fn run() -> Result<(), CliError> {
...
+                )
+                .subcommand(
+                    SubCommand::with_name("default")
+                        .about("Manage default values for circuit creation")
+                        .subcommand(
+                            SubCommand::with_name("set")
+                                .about("Set a default value")
+                                .setting(AppSettings::SubcommandRequiredElseHelp)
+                                .subcommand(
+                                    SubCommand::with_name("service-type")
```
peterschwarz: Is there a reason why this is a separate sub command vs `set <value key> <value>` where `<value key>` is in the set `{service-type, management-type}`.  That seems more extensible, where we can add more default keys, as we decide which can be configured with defaults.

cli/src/store/default_value/yaml_store.rs
eloaverona: Added

cli/src/main.rs
eloaverona: I changed it to show. Also added a commit to change the `node alias get` to `node alias show`, which was in the PR merged yesterday

cli/src/main.rs
eloaverona: Okay, I updated the PR to be as you suggested.

cli/src/action/circuit/defaults.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let name = match args.value_of("name") {
+            Some(key) => key,
+            None => return Err(CliError::ActionError("name is required".into())),
+        };
+
+        let key = match name {
+            "service-type" => SERVICE_TYPE_KEY,
+            "management-type" => MANAGEMENT_TYPE_KEY,
+            _ => return Err(CliError::ActionError("Invalid default name".into())),
```
peterschwarz: This should probably be `format!("{} is not a valid default name", name)`

peterschwarz: Likewise  in the other places used.

eloaverona: updated

PR #479, opened 2020-02-03
Add protocol guards to the Scabbard REST API endpoints

peterschwarz:
This change required a bit of gymnastics, with respect to the RequestGuard trait and providing what's required by the Resource builder.

### DIFF
--- a/libsplinter/src/protocol/mod.rs
+++ b/libsplinter/src/protocol/mod.rs
@@ -16,12 +16,12 @@
 
 pub const ADMIN_PROTOCOL_VERSION: u32 = 1;
 
-pub const ADMIN_APPLICATION_REGISTRATION_PROTOCOL_MIN: u32 = 1;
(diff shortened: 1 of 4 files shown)


libsplinter/src/protocol/mod.rs
```diff
@@ -25,3 +25,14 @@ pub(crate) const ADMIN_FETCH_PROPOSALS_PROTOCOL_MIN: u32 = 1;
 
 #[cfg(feature = "proposal-read")]
 pub(crate) const ADMIN_LIST_PROPOSALS_PROTOCOL_MIN: u32 = 1;
+
+pub const SUBSCRIBE_PROTOCOL_VERSION: u32 = 1;
```
ltseeley: Should this be `SCABBARD_PROTOCOL_VERSION`?

peterschwarz: Fixed

PR #480, opened 2020-02-03
Add scabbard to splinter-dev dockerfile

rbuysse:


peterschwarz:
Could you add some "why" to the commit message?   This is very different than what we do with the other dependencies.  It would be good to have some history as to why this particular copy is needed.

PR #481, opened 2020-02-03
Splinter State Interface

RyanLassigBanks:
This PR removes the use of lock wrapped splinter states (`Arc<RwLock<SplinterState>>`). The circuit and service directories are now wrapped in arc locks and the interface for splinter state has been updated so the internal locks are hidden.

### DIFF
--- a/libsplinter/src/circuit/mod.rs
+++ b/libsplinter/src/circuit/mod.rs
@@ -476,6 +476,48 @@ pub enum WriteError {
     GetStorageError(String),
 }
 
+#[derive(Debug)]
+pub struct SplinterStateError {
+    context: String,
+    source: Option<Box<dyn Error + Send + 'static>>,
+}
+
+impl SplinterStateError {
+    pub fn new(context: String) -> Self {
+        Self {
+            context,
+            source: None,
+        }
+    }
+
+    pub fn from_source<T: Error + Send + 'static>(context: String, source: T) -> Self {
+        Self {
+            context,
+            source: Some(Box::new(source)),
+        }
+    }
+
+    pub fn context(&self) -> String {
+        self.context.clone()
+    }
+}
+
+impl Error for SplinterStateError {}
+
+impl fmt::Display for SplinterStateError {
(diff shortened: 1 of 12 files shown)


libsplinter/src/circuit/mod.rs
```diff
@@ -380,94 +381,167 @@ impl<'r> IntoIterator for &'r Roster {
...
-        **storage.write() = self.circuit_directory.clone();
+        **storage.write() = self.circuit_directory()?;
         Ok(())
     }
 
     // ---------- methods to access service directory ----------
-    pub fn service_directory(&self) -> &HashMap<ServiceId, Service> {
-        &self.service_directory
+
+    pub fn service_directory(&self) -> Result<HashMap<ServiceId, Service>, SplinterStateError> {
```
peterschwarz: This function might not make sense any more, if the API is expected to be sync.  The use case for this would be looking up a service by id, so perhaps this should be changed to `get_service(&self, service_id: ServiceId)`  This depends on usage.  This get all version should be used infrequently, as it requires a pretty large clone.

libsplinter/src/circuit/mod.rs
```diff
@@ -380,94 +381,167 @@ impl<'r> IntoIterator for &'r Roster {
...
-        &self.circuit_directory.nodes()
+    pub fn nodes(&self) -> Result<BTreeMap<String, SplinterNode>, SplinterStateError> {
+        Ok(self.circuit_directory()?.nodes().clone())
     }
 
-    pub fn node(&self, node_id: &str) -> Option<&SplinterNode> {
-        self.circuit_directory.node(node_id)
+    pub fn node(&self, node_id: &str) -> Result<Option<SplinterNode>, SplinterStateError> {
+        Ok(self
+            .circuit_directory()?
```
peterschwarz: This results in a clone of the entire directory for every internal examination, which will be memory and time expensive as the number of circuits grow.  It should use the lock directly and return just a clone of the item being queried.  Same with the other usages

PR #483, opened 2020-02-03
Reorganize Canopy for export

dplumb94:
This PR reorganizes most of the Canopy code. The goal is to get Canopy to a state where it can be imported as a library to any React app to make it a Canopy application.

This PR implements a new component from CanopyJS, `CanopyProvider`, which provides the loading and config functionality to a canopy app. Most of the code for this comes directly from canopy/app, but it is reformatted to be a [React Context Provider](https://reactjs.org/docs/context.html).


canopy/canopyjs/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+  return userSaplingsResponse;
+};
+
+const fetchConfigSaplings = async saplingURL => {
+  const response = await get(`${saplingURL}/configSaplings`);
+  const configSaplingsResponse = response.json;
+  return configSaplingsResponse;
+};
+
+export function CanopyProvider({ saplingURL, splinterURL, children }) {
```
dnewh: Could this and its associated methods get moved into a canopy-provider.js file?

dplumb94: Is there a specific rationale for that? CanopyProvider and the useContext wrappers are the only things that canopyJS is going to export, so I didn't think there was much of a point to move them to a separate file.

canopy/canopyjs/README.md
```diff
@@ -0,0 +1,125 @@
...
+- Exposes shared configuration to saplings and Canopy application components
+
+## Configuration
+
+CanopyJS makes use of two endpoints, `splinterURL` and `saplingURL`.
+
+### splinterURL
+
+`splinterURL` is the URL where the Splinter daemon is running. This URL will be
+used by Canopy and saplings to interact with Splinter via the Splinter Daemon's
```
chenette: daemon's

canopy/canopyjs/src/index.js
dplumb94: After an in person discussion, we decided to resolve this in a future PR.

canopy/app/saplings/userSaplings
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "displayName": "Vanilla",
+    "namespace": "vanilla",
+    "runtimeFiles": [
+      "0.0.0.0:8675/vanilla.js"
+    ],
+    "styleFiles": [],
+    "workerFiles": [],
+    "icon": "https://via.placeholder.com/48"
```
jsmitchell: is it possible to have this stuff self-hosted out of a local directory?

canopy/canopyjs/src/loadSaplings.js
```diff
@@ -14,11 +14,10 @@
...
-import promiseLoader from 'promiseLoader';
-import styleLoader from 'styleLoader';
-import { getUserSaplings, getConfigSaplings } from './getSaplings.macro';
+import promiseLoader from './promiseLoader';
+import styleLoader from './styleLoader';
 
-async function loadCurrentSapling(userSaplingsResponse) {
+export async function mountCurrentSapling(userSaplingsResponse) {
   // Saplings will be guaranteed to to have a collision-free namespace that
   // coresponds with the place entry in the URL's path
```
jsmitchell: I know it's not in the diff, but s/coresponds/corresponds/

canopy/app/saplings/userSaplings
dplumb94: The icon specifically? In practice it would come from wherever the sapling is being hosted.

jsmitchell: Per in-person conversation this was converted from an existing json example. I'm ok deferring this for a future story which does a pass on self hosting as many assets as are reasonable (gameroom vue icons, this, etc.)

PR #485, opened 2020-02-04
Add protocol version guards to biome endpoints

peterschwarz:
Add protocol version guards to all the biome REST API resources.


eloaverona:
I didn't add the protocol version param to the openapi doc. Can you add those in this PR too?

peterschwarz:
> I didn't add the protocol version param to the openapi doc. Can you add those in this PR too?

Added

PR #487, opened 2020-02-05
Move config-* features to default

shannynalayna:
Rearranges the tests previously in `/config/mod` to the individual module the test concerns. Also gets rid of the `config-builder` feature. Finally, moves the `config-*` features to default.

This is in support of reworking splinterd's config modules.


splinterd/Cargo.toml
```diff
@@ -54,7 +54,6 @@ experimental = [
     "biome-credentials",
     "biome-key-management",
     "circuit-read",
-    "config-builder",
```
ltseeley: Can you add the reasoning for this change to the commit message?

PR #488, opened 2020-02-05
Update generate help to have more information

agunde406:
Example:
```
$ splinter cert generate -h
splinter-cert-generate 
Generates test certificates and keys for running splinterd with TLS (in insecure mode)

USAGE:
    splinter cert generate [FLAGS] [OPTIONS]

FLAGS:
        --force      Overwrites files if they exist. If this flag is not provided and the file exists, an error is
                     returned.
                                                 
    -h, --help       Prints help information
    -q, --quiet      Do not display output
        --skip       Checks if the files exists and generates the files that are missing. If this flag is not provided
                     and the file exists, an error is returned.
    -V, --version    Prints version information
    -v               Log verbosely

OPTIONS:
    -d, --cert-dir <cert_dir>          Path to the directory certificates are created in. Defaults to
                                       /etc/splinter/certs/. This location can also be changed with the
                                       SPLINTER_CERT_DIR environment variable. This directory must exist.
                                                                   
        --common-name <common_name>    String that specifies a common name for the generated certificate (defaults to
                                       localhost). Use this option if the splinterd URL uses a DNS address instead of a
                                       numerical IP address.

DETAILS: 

The files are generated in the location specified by --cert-dir, the SPLINTER_CERT_DIR environment variable, or in the
default location /etc/splinter/certs/. 

The following files are created: 
    - client.crt 
    - client.key 
    - server.crt 
    - server.key 
    - generated_ca.pem 
    - generated_ca.key


```

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -41,8 +41,6 @@ pub fn log_format(
 }
 
 fn run() -> Result<(), CliError> {
-    // ignore unused_mut while there are experimental features
-    #[allow(unused_mut)]
     let mut app = clap_app!(myapp =>
         (name: APP_NAME)
         (version: VERSION)
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -21,7 +21,7 @@ mod action;
 mod error;
 mod store;
 
-use clap::clap_app;
+use clap::{clap_app, AppSettings, Arg, SubCommand};
 use flexi_logger::{DeferredNow, LogSpecBuilder, Logger};
 use log::Record;
 
@@ -92,8 +92,6 @@ fn run() -> Result<(), CliError> {
 
     #[cfg(feature = "keygen")]
     {
-        use clap::{Arg, SubCommand};
-
         app = app.subcommand(
             SubCommand::with_name("keygen")
                 .about(
@@ -123,8 +121,6 @@ fn run() -> Result<(), CliError> {
 
     #[cfg(feature = "health")]
     {
-        use clap::{Arg, SubCommand};
-
         app = app.subcommand(
             SubCommand::with_name("health")
                 .about("Displays information about network health")
@@ -146,8 +142,6 @@ fn run() -> Result<(), CliError> {
 
     #[cfg(feature = "database")]
     {
-        use clap::{Arg, SubCommand};
-
         app = app.subcommand(
             SubCommand::with_name("database")
                 .about("Database commands")
@@ -166,8 +160,6 @@ fn run() -> Result<(), CliError> {
 
     #[cfg(feature = "circuit")]
     {
-        use clap::{AppSettings, Arg, SubCommand};
-
         let create_circuit = SubCommand::with_name("create")
             .about("Propose that a new circuit is created")
             .arg(
@@ -453,8 +445,6 @@ fn run() -> Result<(), CliError> {
 
     #[cfg(feature = "node-alias")]
     {
-        use clap::{AppSettings, Arg, SubCommand};
-
         app = app.subcommand(
             SubCommand::with_name("node")
                 .about("Provides node management functionality")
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -77,24 +75,69 @@ fn run() -> Result<(), CliError> {
...
+    );
+
+    app = app.subcommand(
+        SubCommand::with_name("cert")
+            .about("Generate certificates that can be used for development")
+            .setting(AppSettings::SubcommandRequiredElseHelp)
+            .subcommand(
+                SubCommand::with_name("generate")
+                    .long_about(
+"This command generates certificate and keys that can be used for running the Splinter daemon with
```
vaporos: The placement of this feel unnatural (in terms of how --help options normally present).

cli/src/main.rs
```diff
@@ -77,24 +75,69 @@ fn run() -> Result<(), CliError> {
...
+TLS and in insecure mode. The following files are created:
+    - client.crt
+    - client.key
+    - server.crt
+    - server.key
+    - generated_ca.pem
+    - generated_ca.key
+
+The files are generated in the location specified by --cert-dir, the SPLINTER_CERT_DIR
+environment variable, or in the default location /etc/splinter/certs/.
```
chenette: How about moving the list of files and default value to an `after_help` section?  Doing this and slightly expanding the `about` description would mean that this `long_help` would be necessary.

cli/src/main.rs
```diff
@@ -77,24 +75,69 @@ fn run() -> Result<(), CliError> {
...
-                  "Name of the directory in which to create the certificates")
-                (@arg force: --force  conflicts_with[skip] "Overwrite files if they exist")
-                (@arg skip: --skip conflicts_with[force] "Check if files exists, generate if missing")
-            )
-        )
+    );
+
+    app = app.subcommand(
+        SubCommand::with_name("cert")
+            .about("Generate certificates that can be used for development")
```
chenette: I think the extra info in the `long_about` is helpful and would be appropriate in this description (in a briefer form). For example:

> Generates test certificates and keys for running splinterd with TLS (in insecure mode)

cli/src/main.rs
```diff
@@ -77,24 +75,69 @@ fn run() -> Result<(), CliError> {
...
+environment variable, or in the default location /etc/splinter/certs/.
+                    ",
+                    )
+                    .arg(
+                        Arg::with_name("common_name")
+                            .long("common-name")
+                            .takes_value(true)
+                            .long_help(
+                                "The common name that should be used in the generated \
+                                certificate. This option takes a string, defaults to localhost.",
```
chenette: How about expanding with info on when you must use this option?  Perhaps something like this (though please check my assumptions and terminology):

> String that specifies a common name for the generated certificate. 
> Default: localhost. Use this option if the splinterd URL does *not* use 
> a numerical IP address.

chenette: Better version (because I finally found the term for the non-numerical whatsit):

> String that specifies a common name for the generated certificate
> (defaults to localhost). Use this option if the splinterd URL uses a DNS address 
> instead of a numerical IP address.

cli/src/main.rs
agunde406: Put this in the help for generate instead of at the top level of cert

cli/src/main.rs
```diff
@@ -77,24 +75,81 @@ fn run() -> Result<(), CliError> {
...
+    );
+
+    app = app.subcommand(
+        SubCommand::with_name("cert")
+            .about("Generate certificates that can be used for development")
+            .setting(AppSettings::SubcommandRequiredElseHelp)
+            .subcommand(
+                SubCommand::with_name("generate")
+                    .long_about(
+                        "Generates test certificates and keys for running splinterd with \
```
chenette: Is there a reason for the `about` and `long_about` content to be different?  I recommend consistency whenever possible. 

The `long_about` info seems more helpful than the short `about` blurb, because it clarifies that the command generates keys for splinterd and mentions TLS/insecure mode. 

Also, line 82 should start with "Generates".

chenette: I have been enlightened in an off-line chat, so I am retracting my comment.

PR #489, opened 2020-02-05
Final Config and ConfigBuilder

shannynalayna:
Adds the final Config and ConfigBuilder, as well as tests and supporting changes for these. Also incorporates the Config object into the main of splinterd. The previous config object being used here had optional values, whereas this Config object does not. 

Also adds a `ConfigSource` to support displaying the location a configuration variable was sourced.

### DIFF
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::config::PartialConfig;
+use std::path::Path;
+
+use crate::config::error::ConfigError;
+use crate::config::{Config, ConfigSource, PartialConfig};
 
 pub trait PartialConfigBuilder {
     /// Takes all values set in a config object to create a PartialConfig object.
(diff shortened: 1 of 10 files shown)


splinterd/src/config/default.rs
```diff
@@ -79,8 +79,8 @@ impl DefaultConfig {
             bind: Some(String::from("127.0.0.1:8080")),
             #[cfg(feature = "database")]
             database: Some(String::from("127.0.0.1:5432")),
-            registry_backend: None,
-            registry_file: None,
+            registry_backend: Some(String::from("FILE")),
+            registry_file: Some(String::from("/etc/splinter/")),
```
vaporos: Is this option expecting a filename or a directory?

splinterd/src/config/command_line.rs
```diff
@@ -107,9 +107,20 @@ impl PartialConfigBuilder for CommandLineConfig {
...
+
     use clap::ArgMatches;
 
-    /// Example configuration values.
+    use crate::config::{Config, ConfigBuilder};
+
+    use crate::config::{DefaultConfig, EnvVarConfig, TomlConfig};
+
+    /// Path to existing example config toml files from the top-level Splinterd directory.
+    static TEST_TOML: &str = "sample_configs/splinterd.toml.example";
```
vaporos: We should de-couple "sample" files from the tests, because updating a sample config shouldn't break the tests.

splinterd/src/config/builder.rs
```diff
@@ -12,14 +12,441 @@
...
+        }
+    }
+
+    /// Adds a `storage` value to the ConfigBuilder object.
+    ///
+    /// # Arguments
+    ///
+    /// * `storage` - The type of storage that should be used to store circuit state.
+    ///
+    pub fn with_storage(&mut self, source: ConfigSource, storage: String) {
```
vaporos: Can these with_* statements be private? Seems like we would only need with_partial_config() to be public. Seems like they could be inlined in the above code where they are used.

splinterd/src/config/builder.rs
```diff
@@ -12,14 +12,441 @@
...
+                "Source unset, unable to add to ConfigBuilder",
+            ))),
+        }?;
+        self.partial_configs.insert(source, partial);
+        Ok(())
+    }
+
+    /// Builds a Config object by incorporating the values from each PartialConfig object.
+    ///
+    pub fn build(mut self) -> Result<Config, ConfigError> {
```
vaporos: This should build the config based on the order the partialconfigs were added and should not explicitly consider the source itself.

splinterd/src/config/builder.rs
```diff
@@ -12,14 +12,441 @@
...
+    registry_file: Option<(String, ConfigSource)>,
+    heartbeat_interval: Option<(u64, ConfigSource)>,
+    admin_service_coordinator_timeout: Option<(Duration, ConfigSource)>,
+    state_dir: Option<(String, ConfigSource)>,
+}
+
+impl ConfigBuilder {
+    pub fn new() -> Self {
+        ConfigBuilder {
+            partial_configs: HashMap::new(),
```
vaporos: This should be an ordered list.

splinterd/src/config/builder.rs
```diff
@@ -12,14 +12,441 @@
...
+            storage: None,
```
vaporos: There should be no need for these fields, only the partial configs should be used when building the config result and could be localized to that method.

splinterd/src/config/partial.rs
```diff
@@ -14,11 +14,22 @@
 
 use std::time::Duration;
 
+/// ConfigSource displays the source of configuration values, used to identify which of the various
+/// config modules were used to create a particular PartialConfig object.
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ConfigSource {
+    TomlFile,
+    Default,
+    EnvironmentVar,
```
vaporos: Consider calling this 'Environment'

splinterd/src/config/partial.rs
```diff
@@ -14,11 +14,22 @@
 
 use std::time::Duration;
 
+/// ConfigSource displays the source of configuration values, used to identify which of the various
+/// config modules were used to create a particular PartialConfig object.
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ConfigSource {
+    TomlFile,
```
vaporos: Consider calling this 'Toml' . Consider adding the toml filename so that can be output in the debug output table.

splinterd/src/config/partial.rs
vaporos: Eventually we may want to make this source per-field instead of per-config so we can fit the environment variable into the debug output. (Not in this PR though.)

splinterd/src/main.rs
```diff
@@ -506,117 +393,45 @@ fn get_transport(
...
-                if let Ok(s) = env::var(CERT_DIR_ENV) {
-                    s
-                } else {
-                    DEFAULT_CERT_DIR.to_string()
-                }
-            };
+            let client_cert = config.client_cert();
+            if !Path::new(&client_cert).is_file() {
+                return Err(GetTransportError::CertError(
+                    "Must provide a valid client certificate.".into(),
```
vaporos: Remove the period from the errors.

splinterd/src/config/mod.rs
```diff
@@ -36,3 +38,215 @@ pub use crate::config::toml::TomlConfig;
...
+    fn admin_service_coordinator_timeout_source(&self) -> ConfigSource {
+        self.admin_service_coordinator_timeout.1.clone()
+    }
+
+    fn state_dir_source(&self) -> ConfigSource {
+        self.state_dir.1.clone()
+    }
+
+    /// Displays the configuration value along with where the value was sourced from.
+    pub fn display(&self) {
```
vaporos: Rename this to log_as_debug() or similar.

splinterd/src/config/mod.rs
```diff
@@ -36,3 +38,215 @@ pub use crate::config::toml::TomlConfig;
...
+        debug!("Configuration values: storage: {} (source: {:?}) \
```
vaporos: Break this up to one debug statement per value with the shorter prefix "Config: ". Use the same format option for each line so if we want to change it (for example, try and line up the fields), we can do that easily.

splinterd/src/config/default.rs
shannynalayna: It's expecting a filename, but it's passed in as an option to create the splinter daemon, so I figured this would ensure the default would pass along a None option to that if the registry_file isn't a file.

splinterd/src/config/builder.rs
shannynalayna: Are we not concerned about the source in the sense that config values from the command line should have precedence over the default values? This takes care of that concern, regardless of when the PartialConfigs are added.

splinterd/src/config/builder.rs
shannynalayna: This is supporting the ConfigBuilder checking the source, to give the correct PartialConfig object values precedence over others. I assumed the ConfigBuilder should take care of that precedence concern

splinterd/src/config/default.rs
vaporos: This should include the filename in that case.

splinterd/src/config/builder.rs
vaporos: We are, but the code using the builder should determine the order. Adding another partial config shouldn't involve changing this implementation.

splinterd/src/config/builder.rs
vaporos: The value and source should be found by scanning the list until a config with a value is found.

splinterd/src/config/mod.rs

splinterd/src/config/mod.rs

splinterd/src/main.rs
shannynalayna: Addressed.

splinterd/src/config/partial.rs
shannynalayna: Addressed.

splinterd/src/config/partial.rs
shannynalayna: Agreed, the ConfigBuilder's build method could be simplified with having the source per-field.

splinterd/src/config/builder.rs
shannynalayna: Addressed.

splinterd/src/config/builder.rs
shannynalayna: Addressed, these methods were removed.

splinterd/src/config/command_line.rs
shannynalayna: I added a commit to de-couple the sample files.

splinterd/src/config/partial.rs
```diff
@@ -14,11 +14,22 @@
 
 use std::time::Duration;
 
+/// ConfigSource displays the source of configuration values, used to identify which of the various
+/// config modules were used to create a particular PartialConfig object.
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum ConfigSource {
+    Toml { file: String },
+    Default,
```
ltseeley: Is `Some(ConfigSource::Default)` different from `None`? Maybe we could scrap the `Option` if `PartialConfig` is constructed with a `new(source: ConfigSource) -> Self` method rather than `default()`, which would simplify some things.

splinterd/src/config/default.rs
```diff
@@ -79,8 +79,8 @@ impl DefaultConfig {
             bind: Some(String::from("127.0.0.1:8080")),
             #[cfg(feature = "database")]
             database: Some(String::from("127.0.0.1:5432")),
-            registry_backend: None,
-            registry_file: None,
+            registry_backend: Some(String::from("FILE")),
+            registry_file: Some(String::from("/etc/splinter/registry_file.yaml")),
```
vaporos: Based on a conversation w/rbb, this should be /etc/splinter/nodes.yaml.

splinterd/src/config/mod.rs
```diff
@@ -36,3 +38,285 @@ pub use crate::config::toml::TomlConfig;
...
+    database: (String, ConfigSource),
+    registry_backend: (String, ConfigSource),
+    registry_file: (String, ConfigSource),
+    heartbeat_interval: (u64, ConfigSource),
+    admin_service_coordinator_timeout: (Duration, ConfigSource),
+    state_dir: (String, ConfigSource),
+}
+
+impl Config {
+    pub fn storage(&self) -> String {
```
peterschwarz: Can't these return references?

splinterd/src/config/mod.rs
```diff
@@ -36,3 +38,285 @@ pub use crate::config::toml::TomlConfig;
...
+    fn admin_service_coordinator_timeout_source(&self) -> ConfigSource {
+        self.admin_service_coordinator_timeout.1.clone()
+    }
+
+    fn state_dir_source(&self) -> ConfigSource {
+        self.state_dir.1.clone()
+    }
+
+    /// Displays the configuration value along with where the value was sourced from.
+    pub fn log_as_debug(&self) {
```
ltseeley: Is there a reason this was chosen over implementing the `Debug` trait and just calling `debug!("{}", config}`? Seems more idiomatic and flexible.

splinterd/src/main.rs
```diff
@@ -721,6 +722,9 @@ impl fmt::Display for UserError {
             UserError::TransportError(err) => write!(f, "unable to get transport: {}", err),
             UserError::MissingArgument(msg) => write!(f, "missing required argument: {}", msg),
             UserError::InvalidArgument(msg) => write!(f, "required argument is invalid: {}", msg),
+            UserError::ConfigError(msg) => {
+                write!(f, "Error happened building config object: {}", msg)
```
ltseeley: "error occurred" sounds more natural than "error happened", IMO. Also, this should probably start with a lowercase letter to be consistent with the other errors.

splinterd/src/config/partial.rs
shannynalayna: Yes, the Default is specifically from the DefaultConfig module.

splinterd/src/config/mod.rs
shannynalayna: I believe that having the separate debug statements will allow for better formatting than just calling debug on the entire object, which is ultimately the intention.

splinterd/src/config/partial.rs
shannynalayna: I addressed this by changing the PartialConfig source field to `ConfigSource` from `Option<ConfigSource>`

splinterd/src/config/mod.rs
vaporos: I feel like debug!("{}", config) would be better reserved for a single-line piece of information about the config rather than this very long dump of all the configuration.

vaporos: But further - we are looking for more than one line in the debug output. So what is here now LGTM.

shannynalayna:
Ok, so I've made a few changes in the last few pushes: 

1. Moved the function which built the file paths to the `builder` module. Therefore, the file paths will not be built until they are put into the final `Config` object. Previously, this function lived in the `config/default` module. This should fix the issue brought up by @agunde406 where the cert directory passed in as a command line arg wasn't being appended to the other files. 

2. The `log_as_debug` was broken into multiple functions to fix the issue of "generated" not being used as the source, as well as the specifics of which paths should be printed based on the conditionals of the `generate-certs` and `insecure` flag. Along with the main `log_as_debug` method, there are two additional methods, `log_transport_as_debug` and `log_generated_transport_as_debug` which will hopefully more accurately display the correct configuration values. 

3. The `PathBuf` objects passed into both transport log methods have been passed through the `fs::canonicalize` function, so should be showing up as absolute paths.

splinterd/src/main.rs
```diff
@@ -513,164 +399,82 @@ fn get_transport(
...
-            let log_value = format!(
-                "transport_type: tls, ca_certs: {:?}, client_cert: {:?}, \
-                 client_key: {:?}, server_cert: {:?}, server_key: {:?}",
-                ca_file_log,
-                fs::canonicalize(client_cert.clone())?,
-                fs::canonicalize(client_key_file.clone())?,
-                fs::canonicalize(server_cert.clone())?,
-                fs::canonicalize(server_key_file.clone())?,
+            config.log_transport_as_debug(
+                PathBuf::from(config.cert_dir()),
```
agunde406: When this is logged, it should also be an absolute path

shannynalayna: Addressed.

vaporos:
Disagree with the changes which complicate what is logged - the goal within the scope of this PR is to log the raw configuration values with their source, to help with debugging. The requirements to print full paths should also be considered, but in a separate PR and separate from the config logging done here. What we are trying to achieve here is enough of the "raw" input to know what the user has specified.

splinterd/src/main.rs
```diff
@@ -502,164 +395,70 @@ fn get_transport(
...
-                if let Ok(s) = env::var(CERT_DIR_ENV) {
-                    s
-                } else {
-                    DEFAULT_CERT_DIR.to_string()
-                }
-            };
+            let client_cert = config.client_cert();
+            if !Path::new(&client_cert).is_file() {
+                return Err(GetTransportError::CertError(
+                    "Must provide a valid client certificate".into(),
```
agunde406: These errors messages should also log the file that is being tested.

splinterd/src/config/builder.rs
```diff
@@ -12,19 +12,229 @@
...
+use crate::config::error::ConfigError;
+use crate::config::{Config, PartialConfig};
 
 pub trait PartialConfigBuilder {
     /// Takes all values set in a config object to create a PartialConfig object.
     ///
     fn build(self) -> PartialConfig;
 }
 
+fn get_file_path(cert_dir: &str, file: &str) -> String {
```
agunde406: At least before, if ca_certs, client/server_cert, client/server_keys they took a full path, and would not be prepended with the cert_dir. 

For example if:
`--server-key ../certs/server.key` was passed that location would be used. 

Now it searches this location `/etc/splinter/certs/../certs/server.key` instead.

shannynalayna: Addressed this by checking the source for the certificate file path in the main `build` function. Now, if the certificate file comes from a config (besides default), the value is passed directly like that. However, if the certificate file is from the default config, the cert_dir will be appended to it.

splinterd/src/main.rs
shannynalayna: Addressed.

PR #495, opened 2020-02-06
Validate requester public key

peterschwarz:
Validate the requester's public key length when acting on a circuit proposal or vote.

### DIFF
--- a/libsplinter/src/admin/service/shared.rs
+++ b/libsplinter/src/admin/service/shared.rs
@@ -978,6 +978,8 @@ impl AdminServiceShared {
             ));
         }
 
+        self.validate_key(signer_public_key)?;
+
         let key_info = self
             .key_registry
             .get_key(signer_public_key)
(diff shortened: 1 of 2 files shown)


libsplinter/src/admin/service/shared.rs
```diff
@@ -1837,7 +1854,42 @@ mod tests {
...
+            Box::new(AllowAllKeyPermissionManager),
+            "memory",
+        )
+        .unwrap();
+        let circuit = setup_test_circuit();
+
+        let pub_key = (0u8..50).collect::<Vec<_>>();
+        // too short
+        if let Ok(_) = admin_shared.validate_create_circuit(&circuit, &pub_key[0..10], "node_a") {
+            panic!("Should have been invalid due to signer not being registered to a node");
```
agunde406: Can you update the panic messages to say something about the key lengths

peterschwarz: Updated

PR #496, opened 2020-02-06
Add batch submission function

dnewh:
This adds a batch submission function to saplingjs. This function takes
a url to the endpoint and the batch bytes as a payload. It also adds a
http request convenience function for making other requests.

### DIFF
--- a/canopy/saplingjs/src/index.ts
+++ b/canopy/saplingjs/src/index.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+export { submitBatchList } from './submitter';
+
 interface User {
   userId: string;
   displayName?: string;
--- a/canopy/saplingjs/src/submitter.ts
+++ b/canopy/saplingjs/src/submitter.ts
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2018-2020 Cargill Incorporated
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+interface BatchStatus {
+  statusType: string;
+  message: BatchMessage[];
+}
+
+interface BatchMessage {
+  transactionId: string;
+  errorMessage: string;
+  errorData: number[];
+}
+
+interface BatchInfo {
+  id: string;
+  status: BatchStatus;
+}
+
+const HTTPMethods = ['GET', 'POST', 'PUT', 'PATCH', 'DELETE'];
+
+/**
+ * Wrapper function to set up XHR.
+ * @param {string}      method    HTTP method for the request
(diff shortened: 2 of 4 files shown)


canopy/saplingjs/src/submitter.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export async function http(
+  method: string,
+  url: string,
+  data: Uint8Array,
+  headerFn: (request: XMLHttpRequest) => void
+): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const request = new XMLHttpRequest();
+    request.open(method, url);
```
dplumb94: check if this is a valid method or restrict with a string literal

canopy/saplingjs/src/submitter.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    };
+    request.onerror = (): void => {
+      reject(
+        Error(
+          'The server has encountered an error. Please contact the administrator.'
+        )
+      );
+    };
+    request.send(data);
```
dplumb94: What happens if the method doesn't take a payload? For instance if you are using this for a GET

canopy/saplingjs/src/index.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+export { http, submitBatch } from './submitter';
```
dplumb94: I don't think we should expose `http`.  This should just be a helper function for other things in this library.

canopy/saplingjs/src/submitter.ts
```diff
@@ -0,0 +1,83 @@
...
+  errorMessage: string;
+  errorData: number[];
+}
+
+interface BatchInfo {
+  id: string;
+  status: BatchStatus;
+}
+
+export async function http(
```
peterschwarz: Could you write some documentation for these functions?

canopy/saplingjs/src/submitter.ts
dnewh: For `GET` and `HEAD` requests, the body/data argument is just ignored: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send

canopy/saplingjs/src/submitter.ts
dnewh: Added check. Now rejects with error message if not valid.

canopy/saplingjs/src/submitter.ts
dnewh: added docstrings

canopy/saplingjs/src/index.ts
dnewh: no longer exported

canopy/saplingjs/src/submitter.ts
dplumb94: But it is required in the function signature

dnewh: added none option

canopy/saplingjs/src/submitter.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * Submits a single batch of transactions
+ * @param {string}      url   The endpoint to submit the batch to
+ * @param {Uint8Array}  batch The serialized batch of transactions
+ */
+export async function submitBatch(
+  url: string,
+  batch: Uint8Array
```
peterschwarz: Late to the realization that this needs to be a batch list.  That doesn't change the type but the name and documentation.

Or, the content needs to be added to a BatchList protobuf instance before it is submitted

dnewh: name/docs updated

PR #497, opened 2020-02-06
Include node_id in circuit vote error message

eloaverona:


libsplinter/src/admin/service/shared.rs
```diff
@@ -1187,8 +1187,10 @@ impl AdminServiceShared {
 
         if signer_node != node_id {
             return Err(AdminSharedError::ValidationFailed(format!(
-                "Payload requester node id does not match the node the key is registered to: {}",
-                to_hex(circuit_proposal.get_requester())
+                "Payload requester node id does not match the node the key is registered to: {}. \
```
peterschwarz: This might read better now as "Payload requester {pubkeyhex} is not registered to {node-id}"

desktophero:
LGTM ... thanks!

PR #500, opened 2020-02-07
Improve shutdown of event reactor

peterschwarz:
This PR removes the delay on shutdown caused by the `event::Reactor` using the combination of a running flag and reacting to that flag based on the, as well as signalling and joining shutdown in the same function (delaying other shutdown activities).
 

- replaces the use of the atomic boolean as a  stop/close/disconnect signal with a sender that takes a union type, WebSocketClientCmd.  This union type is either a Frame or the stop - signal.
 This change makes the shutdown essentially immediate, instead of waiting  for a message from the server (most-likely a ping) before disconnecting the websocket
- splits `event::Reactor` shutdown function into two parts: a shutdown signaler and a wait-for-shutdown function.  This allows the signaling to be done in a non-blocking model and the wait-for-shutdown to be done as part of a join process.

To test:

Build and up gameroom:

```
$ docker-compose -f examples/gameroom/docker-compose.yaml up --build
```

Once it has started and the admin websockets have been created, in another terminal:

```
$ docker-compose -f examples/gameroom/docker-compose.yaml stop gameroomd-acme
```
This should exit `0`, instead of `137`.

### DIFF
--- a/examples/gameroom/daemon/src/main.rs
+++ b/examples/gameroom/daemon/src/main.rs
@@ -122,8 +122,14 @@ fn run() -> Result<(), GameroomDaemonError> {
         public_key.as_hex(),
     )?;
 
+    let reactor_shutdown_signaler = reactor.shutdown_signaler();
+
     ctrlc::set_handler(move || {
         info!("Received Shutdown");
+
+        if let Err(err) = reactor_shutdown_signaler.signal_shutdown() {
+            error!("Unable to cleanly shutdown event reactor: {}", err);
+        }
 
         if let Err(err) = rest_api_shutdown_handle.shutdown() {
             error!("Unable to cleanly shutdown REST API server: {}", err);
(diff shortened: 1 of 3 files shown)


libsplinter/src/events/reactor.rs
```diff
@@ -135,15 +136,29 @@ impl Reactor {
         }
     }
 
+    /// Return a ReactorShutdownSignaler, for use to send messages to
```
agunde406: This comment is confusing. Can it be reworded?

peterschwarz: Updated

examples/gameroom/daemon/src/main.rs
```diff
@@ -122,9 +122,15 @@ fn run() -> Result<(), GameroomDaemonError> {
         public_key.as_hex(),
     )?;
 
+    let reactor_shutdown_signaler = reactor.shutdown_signaler();
+
     ctrlc::set_handler(move || {
         info!("Received Shutdown");
 
+        if let Err(err) = reactor_shutdown_signaler.signal_shutdown() {
+            error!("Unable to cleanly shutdown Event Reactor: {}", err);
```
jsmitchell: Be consistent with capitalization of Event Reactor (reactor.rs logs it as "event reactor")

peterschwarz: Updated.

PR #505, opened 2020-02-10
Scabbard batch status wait fix

ltseeley:
Update the scabbard state to allow specifying a wait time for batches to
complete when getting batch statuses. An iterator is now returned by
this method that, if a timeout is specified, will only provide completed
batch infos. This change removes the need for the REST API to actively
poll for completed batches.

### DIFF
--- a/libsplinter/src/service/scabbard/rest_api.rs
+++ b/libsplinter/src/service/scabbard/rest_api.rs
@@ -12,11 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp::min;
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::thread::sleep;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use transact::protocol::batch::BatchPair;
 use transact::protos::FromBytes;
@@ -30,11 +28,10 @@ use crate::rest_api::{
 use crate::service::rest_api::ServiceEndpoint;
 
 use super::error::StateSubscriberError;
-use super::state::{BatchInfo, BatchStatus, StateChangeEvent, StateSubscriber};
+use super::state::{StateChangeEvent, StateSubscriber};
 use super::{Scabbard, SERVICE_TYPE};
 
 const DEFAULT_BATCH_STATUS_WAIT_SECS: u64 = 300;
-const BATCH_STATUS_RETRY_INTERVAL_MILLIS: u64 = 1000;
 
 struct WsStateSubscriber {
     sender: EventSender<StateChangeEvent>,
@@ -247,7 +244,7 @@ pub fn make_get_batch_status_endpoint() -> ServiceEndpoint {
                 };
 
             let ids = if let Some(ids) = query.get("ids") {
-                ids.split(',').map(String::from).collect::<Vec<_>>()
+                ids.split(',').map(String::from).collect()
             } else {
                 return Box::new(
                     HttpResponse::BadRequest()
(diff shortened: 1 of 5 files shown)


libsplinter/src/service/scabbard/rest_api.rs
```diff
@@ -272,27 +269,23 @@ pub fn make_get_batch_status_endpoint() -> ServiceEndpoint {
...
+            match scabbard.get_batch_info(ids, wait) {
+                Ok(iter) => match iter.collect::<Result<Vec<_>, _>>() {
+                    Ok(batch_infos) => Box::new(HttpResponse::Ok().json(batch_infos).into_future()),
+                    Err(err) => {
+                        error!(
+                            "Failed to get completed batch statuses before timeout: {}",
+                            err
+                        );
+                        Box::new(
                             HttpResponse::InternalServerError()
```
ltseeley: I don't think this is an InternalServerError, but I couldn't find an error code that seemed to fit this situation.

libsplinter/src/service/scabbard/state.rs
```diff
@@ -609,22 +609,44 @@ impl BatchHistory {
 
     fn update_batch_status(&mut self, signature: &str, status: BatchStatus) {
         match self.history.get_mut(signature) {
-            Some(ref mut batch) if batch.status == BatchStatus::Pending => {
-                batch.set_status(status);
+            Some(info) => {
+                if info.status == BatchStatus::Pending {
+                    info.set_status(status);
+                } else {
```
peterschwarz: This could still have been done as an additional match arm

libsplinter/src/service/scabbard/state.rs
```diff
@@ -609,22 +609,44 @@ impl BatchHistory {
...
-        let info = if let Some(info) = self.history.get_mut(signature) {
-            info
-        } else {
-            return;
-        };
-
-        if let BatchStatus::Valid(t) = info.status.clone() {
-            info.set_status(BatchStatus::Committed(t));
+        match self.history.get_mut(signature) {
+            Some(info) => match info.status.clone() {
```
peterschwarz: This clone could be done as late as possible - i.e. when you're going to return set the status

libsplinter/src/service/scabbard/rest_api.rs
peterschwarz: It could be `408: Request Timeout`

libsplinter/src/service/scabbard/state.rs
```diff
@@ -219,7 +225,7 @@ impl ScabbardState {
 
         // Get the results and shutdown the scheduler
         let batch_result = result_rx
-            .recv_timeout(std::time::Duration::from_secs(EXECUTION_TIMEOUT))
+            .recv_timeout(Duration::from_secs(EXECUTION_TIMEOUT))
```
peterschwarz: recv_timeout is panic-prone

libsplinter/src/service/scabbard/state.rs
ltseeley: That actually produces a warning:

warning: cannot borrow `*info` as mutable because it is also borrowed as immutable
   --> libsplinter/src/service/scabbard/state.rs:634:21
    |
632 |             Some(info) => match &info.status {
    |                                 ------------ immutable borrow occurs here
633 |                 BatchStatus::Valid(txns) => {
634 |                     info.set_status(BatchStatus::Committed(txns.clone()));
    |                     ^^^^ mutable borrow occurs here        ---- immutable borrow later used here
    |
    = note: `#[warn(mutable_borrow_reservation_conflict)]` on by default
    = warning: this borrowing pattern was not meant to be accepted, and may become a hard error in the future
    = note: for more information, see issue #59159 <https://github.com/rust-lang/rust/issues/59159>

libsplinter/src/service/scabbard/rest_api.rs
ltseeley: Updated to 408

libsplinter/src/service/scabbard/state.rs
ltseeley: Updated

libsplinter/src/service/scabbard/state.rs
ltseeley: This isn't changed in this PR, just added the import for `std::time::Duration`

peterschwarz: That may be true, but you use it in your iterator later.

ltseeley: Created a story to address this issue for splinter as a whole

libsplinter/src/service/scabbard/rest_api.rs
```diff
@@ -272,27 +269,23 @@ pub fn make_get_batch_status_endpoint() -> ServiceEndpoint {
...
-                                    "message": "An internal error occurred"
-                                }))
-                                .into_future(),
-                        );
-                    }
+            match scabbard.get_batch_info(ids, wait) {
+                Ok(iter) => match iter.collect::<Result<Vec<_>, _>>() {
+                    Ok(batch_infos) => Box::new(HttpResponse::Ok().json(batch_infos).into_future()),
+                    Err(err) => Box::new(
+                        HttpResponse::RequestTimeout()
```
dplumb94: openapi spec should be updated with this new error response

ltseeley: Added to the OpenAPI spec

PR #512, opened 2020-02-12
Clean up loading spinner implementation

dplumb94:
- Implements a new component, `Loading`, which renders a spinner and a message supplied by a prop. This standardizes the approach to loading indicators throughout gameroom.
- Updates the vuex page loading store to store a message.
- Trigger a loading indicator when pages are being lazy loaded or data has to be fetched before the page can fully load

This eliminates the following issues:
- Previously, a spinner that was positioned at the bottom of the page was bumping the margins of the website which caused scrollbars to flash on and off on some browsers. This has been replaced by the `Loading` component which renders in the middle of the screen.
- Previously, data would cascade on to the page, which could cause unintended empty spaces where data should be. This was especially evident on slower connections. Now the loading indicator will display until data is ready to be displayed.

To test:
1. start up Gameroom in docker as normal.
2. create a gameroom and several games.
3. Navigate around to different pages to ensure that there is a smooth experience. You will notice that navigating to different gamerooms will trigger the loader for a brief period while the games for that gameroom are fetched.

If you are on chrome (there is probably a way to do this on firefox too, but I am not sure):
1. Open devtools and select the device toolbar (phone/tablet icon on the top left).
2. Expand the page to a size supported by gameroom.
3. Go to the 'Online' dropdown and select mid-tier mobile or low-tier mobile. This simulates a slower network.
4. Navigate around and ensure that things are loading properly.

### DIFF
--- a/examples/gameroom/gameroom-app/src/views/GameroomDetail.vue
+++ b/examples/gameroom/gameroom-app/src/views/GameroomDetail.vue
@@ -42,7 +42,7 @@ limitations under the License.
       <h2 class="gameroom-name">{{ gameroom.alias }}</h2>
       <span> {{ gemeroomMembers }} </span>
     </div>
-        <div class="data-container">
+        <div v-if="gameroom.status === 'Active'" class="data-container">
           <div class="header">
           <div class="tab-buttons">
           <button class="tab-button"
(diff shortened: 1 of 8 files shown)


examples/gameroom/gameroom-app/src/views/GameroomDetail.vue
```diff
@@ -95,65 +95,57 @@ limitations under the License.
...
+          next({ name: 'not-found' });
+        }
       }
 
-      async listGames(circuitID: string) {
-        this.loadingGames = true;
-        await this.$store.dispatch('games/listGames', circuitID).then(() => {
-          this.loadingGames = false;
-        });
+      async beforeRouteUpdate(to: any, from: any , next: any) {
```
dnewh: Could a type be created for a route to avoid using the `any` type here?

examples/gameroom/gameroom-app/src/views/GameroomDetail.vue
```diff
@@ -95,65 +95,57 @@ limitations under the License.
...
-        store.commit('pageLoading/setPageLoading', true);
-        store.dispatch('selectedGameroom/updateSelectedGameroom', to.params.id).catch((e) => {
-            setPageLoading(false);
-            next({ name: 'not-found' });
-        }).then(() => {
-            setPageLoading(false);
-            this.listGames(to.params.id);
-            next();
-        });
+      async beforeRouteEnter(to: any, from: any , next: any) {
```
dnewh: Could a type be created for a route to avoid using the any type here?

dplumb94: Probably should be, but that would be a separate improvement since the only change to this line is making the function async

dnewh: After talking in person, decided to do this in a later change

PR #515, opened 2020-02-14
Implement a variety of fixes to circuit-read 

agunde406:
Fixes these issues raised when reviewing the code: 

 REST API Error responses to the user should be standardized. The Json response returned should use similar format for all of them, some are a string, some are a struct {"message": "error message"}, should match documentation (recommend use the response models in the rest_api module)
 -Used ErrorResponse from rest_api mod.

In circuit_read.rs, ListCircuitsResponse, CircuitResponse should not be public
  -Made them pub(crate)

The actix_web and futures libraries are pulled in from the crate’s re-export, which may not exist in the future; these should probably just be pulled in from their respective crates directly.
  - This was already done with the actix feature

 In fetch_circuit, a request without a circuit ID will return the same error as a request with a circuit ID that does not exist; this should return a different (ultimately, an invalid request).
  - This is not valid because /circuit/ would return not found because it is not a valid route
 
In fetch_circuit, the fetched circuit is cloned (state.circuit(&circuit_id).cloned()), then each of the fields is cloned (e.g. auth: circuit.auth().clone(),). Since these are all just references, these clones should be unnecessary. May be able to remove some clones from query_list_circuits as well.
   - This was no longer an issue after the addition of CircuitStore

When returning InternalServerError, the error should always be logged and/or included in the response. 
    - Error logs added to missing locations
 
In list_circuits, the returned link will have an empty query string if no filters are provided in the request; if no filters are provided, the returned link should have no query string (? should be moved to push_str).
   - Fixed

query_list_circuits has some duplication that can probably be eliminated by reorganizing it.
   - Reorganized to reduce duplication

Also fixed a feature issue where Errors were not available if only using circuit-read and removed some warnings from tests.

Also adds tests for offset and limit.

### DIFF
--- a/libsplinter/src/admin/rest_api/error.rs
+++ b/libsplinter/src/admin/rest_api/error.rs
@@ -14,14 +14,17 @@
 
 use std::error::Error;
 
+#[cfg(feature = "circuit-read")]
 use crate::circuit::store;
 
+#[cfg(feature = "proposal-read")]
 #[derive(Debug)]
 pub enum ProposalRouteError {
     NotFound(String),
     InternalError(String),
 }
 
+#[cfg(feature = "proposal-read")]
 impl Error for ProposalRouteError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match self {
@@ -31,6 +34,7 @@ impl Error for ProposalRouteError {
     }
 }
 
+#[cfg(feature = "proposal-read")]
 impl std::fmt::Display for ProposalRouteError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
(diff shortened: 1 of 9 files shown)


libsplinter/src/admin/rest_api/error.rs
```diff
@@ -18,7 +18,9 @@ use crate::circuit::store;
 
 #[derive(Debug)]
 pub enum ProposalRouteError {
+    #[allow(dead_code)]
```
vaporos: Can these be protected by the conditionals that require them, instead of marking them dead_code? The problem with dead_code is that it gives no clues as to when we should remove these allow statements.

agunde406: updated

PR #520, opened 2020-02-17
Remove Saplingjs and Canopyjs from Splinter

dplumb94:
WIP pending PRs getting merged that allow CanopyJS and SaplingJS to be installed via git dependencies.


canopy/app/Dockerfile
```diff
@@ -14,14 +14,13 @@
...
 
-COPY app/package*.json /project/app
+COPY package*.json ./
 
-COPY canopyjs /project/canopyjs
+RUN apk add git
 
-RUN cd ../canopyjs && yarn install && yarn build && rm -rf node_modules/ && cd ../app
+# Gives npm permission to run the prepare script in splinter-canopyjs as root
+RUN npm config set unsafe-perm true && npm install
```
peterschwarz: is there a way to isolate this just to splinter-canopyjs?

dplumb94: I'm not aware of a way to do it. The only reason it is run as root is because scripts in Docker containers are running as root anyway, but that action is blocked by npm when running as root. If running as a non-root user, the prepare script is run. This is also only an issue with git/local dependencies, because when publishing to the npm registry the prepare script is run on publish.

PR #522, opened 2020-02-18
Use Sabre txn builder pattern and .scar loading with Transact

ltseeley:
Updates the Transact, Sawtooth SDK, and Sabre SDK dependencies to the latest versions for all crates that use them. This allows Sabre batches to be built and submitted with the new builder patterns, and replaces scabbard's own .scar file loading functionality with Transact's (which is more robust).

~~This PR will not build until Transact has been updated to v0.2.2, since it requires the changes from https://github.com/hyperledger/transact/pull/47~~

~~This PR requires fixes in the Sabre SDK to work properly: https://github.com/hyperledger/sawtooth-sabre/pull/102/~~


examples/gameroom/daemon/src/authorization_handler/error.rs
```diff
@@ -152,17 +154,24 @@ macro_rules! impl_from_sabre_errors {
...
+            impl From<$x> for AppAuthHandlerError {
+                fn from(e: $x) -> Self {
+                    AppAuthHandlerError::TransactError(e.to_string())
+                }
+            }
+        )*
+    };
+}
+
+impl_from_transact_errors!(BatchBuildError, TransactionBuildError, ProtoConversionError);
```
peterschwarz: This macro doesn’t seem all that useful for one error

ltseeley: Do you want it changed?

libsplinter/src/service/scabbard/state.rs
```diff
@@ -342,42 +342,64 @@ impl ScabbardState {
...
+                state_changes.iter().cloned().map(|change| match change {
                     transact::protocol::receipt::StateChange::Set { key, value } => {
                         TransactStateChange::Set { key, value }
                     }
                     transact::protocol::receipt::StateChange::Delete { key } => {
                         TransactStateChange::Delete { key }
                     }
                 })
+            }
+            TransactionResult::Invalid { .. } => panic!("already checked for invalid results"),
```
vaporos: The above could be reworked to avoid this panic (to avoid this match completely).

examples/gameroom/daemon/src/authorization_handler/error.rs
vaporos: That would be my preference -- the threshold for using any macros should be pretty high.

ltseeley: Removed the macro and implemented the `From`s for each.

libsplinter/src/service/scabbard/state.rs
ltseeley: Moved this conversion into Transact

PR #526, opened 2020-02-20
Documentation updates for circuit-read

ltseeley:
Add basic mod and public API documentation, update OpenAPI spec for circuit-read endpoints.

### DIFF
--- a/splinterd/api/static/openapi.yml
+++ b/splinterd/api/static/openapi.yml
@@ -198,9 +198,15 @@ paths:
 
   /admin/circuits:
     get:
+      summary: Fetches a list of circuits that the node belongs to
+      description: |
+        This endpoint can be used to view all or some of the circuits that the
+        node is a member of. If a node ID is provided via the "filter" query
+        parameter, only circuits that have the given node ID as a member will be
+        returned; if no filter is provided, all of the node's circuits will be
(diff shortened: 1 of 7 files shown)


splinterd/api/static/openapi.yml
```diff
@@ -198,9 +198,16 @@ paths:
...
+        This endpoint can be used to view all or some of the circuits that the
+        node is a member of. If a node ID is provided via the "filter" query
+        parameter, only circuits that have the given node ID as a member will be
+        returned; if no filter is provided, all of the node's circuits will be
+        returned.
       tags:
-          - Circuits
-      description: Experimental - List of circuits in Splinter state
+        - admin
+        - circuit
```
ltseeley: I'm not sure if these are the correct tags (or if they should be capitalized?)

agunde406: Looking at the other ones they should be capital (Key Registry, Node Registry, Biome) Also have both --circuit and --admin cause the routes to show up twice the generate docs so I think we should only have one. Currently we have the admin service stuff separated out between the Admin Server (websockets and submit), Circuits, Proposals (also includes Node Registry and Key registry now) So i think it makes sense to just leave this as Circuits

PR #528, opened 2020-02-21
Initial Gameroom biome

RyanLassigBanks:
This PR migrates users and authentication in gameroom to biome

* Login and register endpoints now forward login info to splinterd biome endpoints for key management and login
* Removes `gameroom_users` tables and helper methods
* Updates gameroom docker compose to user biome enabled splinterd containers


examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,153 @@ pub struct AuthResponseData {
...
+            let body = login_response.body().await?;
+            let body_value: serde_json::Value = serde_json::from_slice(&body)?;
+            let message = match body_value.get("message") {
+                Some(value) => value.as_str().unwrap_or("Request was malformed."),
+                None => "Request malformed.",
+            };
+            return Ok(HttpResponse::Unauthorized().json(ErrorResponse::unauthorized(&message)));
+        }
+        _ => {
+            debug!(
```
eloaverona: I think this should be logged as an error, same for all other internal errors.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,153 @@ pub struct AuthResponseData {
...
+                "Internal Server Error. Splinterd responded with error {}",
+                login_response.status(),
+            );
+            return Ok(HttpResponse::InternalServerError().json(ErrorResponse::internal_error()));
+        }
+    };
+
+    info!("{}", key.encrypted_private_key);
```
eloaverona: Is this log here intentional? If yes, add a message for more context.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -66,63 +185,111 @@ pub struct UserCreate {
...
+            debug!(
+                "Internal Server Error. Failed to create key {}",
+                create_key_response.status(),
+            );
+            return Ok(HttpResponse::InternalServerError().json(ErrorResponse::internal_error()));
         }
     }
-    Err(RestApiResponseError::Unauthorized)
+
+    info!("enc private key: {}", new_user.encrypted_private_key);
```
eloaverona: Is this log here intentional? If yes, change the message for more context.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -66,63 +185,111 @@ pub struct UserCreate {
...
+            debug!(
+                "Internal Server Error. Splinterd responded with error {}",
+                login_response.status(),
+            );
+            return Ok(HttpResponse::InternalServerError().json(ErrorResponse::internal_error()));
+        }
+    };
+
+    // Create Key
+
```
agunde406: Remove this blank line

examples/gameroom/tests/cypress/scripts/clear_db.sh
```diff
@@ -12,6 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PGPASSWORD=gameroom_test psql -h db-cypress-test -U gameroom_test -c \
-    'TRUNCATE gameroom, gameroom_proposal, proposal_vote_record,
-      gameroom_member, gameroom_service, gameroom_notification, xo_games';
+PGPASSWORD=admin psql -h splinterd-db-node -U admin -d splinter -c \
+    'TRUNCATE keys, notification_properties, notifications, splinter_user,
```
agunde406: Do the gameroom tables not need to be cleared anymore?

RyanLassigBanks: I don't think they do. All the cypress tests are around login and register, which are now biome operations, so they no longer modify the gameroom database.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,151 @@ pub struct AuthResponseData {
...
-                .json(ErrorResponse::unauthorized("Invalid email or password"))),
-            error::BlockingError::Canceled => {
-                debug!("Internal Server Error: {}", err);
-                Ok(HttpResponse::InternalServerError().json(ErrorResponse::internal_error()))
+    // forward login to splinterd
+    let mut login_response = client
+        .post(format!("{}/biome/login", splinterd_url.get_ref()))
+        .header(
+            "SplinterProtocolVersion",
+            protocol::ADMIN_PROTOCOL_VERSION.to_string(),
```
ltseeley: This should be `BIOME_PROTOCOL_VERSION` (also elsewhere in this file).

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,151 @@ pub struct AuthResponseData {
...
+    let token: Token = match login_response.status() {
+        StatusCode::OK => login_response.json().await?,
+        StatusCode::BAD_REQUEST => {
+            let body = login_response.body().await?;
+            let body_value: serde_json::Value = serde_json::from_slice(&body)?;
+            let message = match body_value.get("message") {
+                Some(value) => value.as_str().unwrap_or("Request was malformed."),
+                None => "Request malformed.",
+            };
+            return Ok(HttpResponse::BadRequest().json(ErrorResponse::bad_request(&message)));
```
ltseeley: Should this really be returned to the calling code as a BadRequest? This function is making the request internally, so if it's bad it's not necessarily the user's fault; 500 might be more appropriate. Same elsewhere in this file

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,151 @@ pub struct AuthResponseData {
...
+        }
+        StatusCode::UNAUTHORIZED => {
+            let body = login_response.body().await?;
+            let body_value: serde_json::Value = serde_json::from_slice(&body)?;
+            let message = match body_value.get("message") {
+                Some(value) => value.as_str().unwrap_or("Request was malformed."),
```
ltseeley: Shouldn't this be "Response was malformed"?

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
```diff
@@ -32,28 +26,151 @@ pub struct AuthResponseData {
...
+
+    let key = match key_response.status() {
+        StatusCode::OK => {
+            let data = key_response.json::<Keys>().await?.data;
+            if let Some(key) = data.get(0) {
+                key.clone()
+            } else {
+                debug!("User does not have any keys");
+                return Ok(
+                    HttpResponse::InternalServerError().json(ErrorResponse::internal_error())
```
ltseeley: This doesn't seem like an internal server error. Maybe the response should allow for the keys to be None?

examples/gameroom/database/tables/gameroom_tables.sql
```diff
@@ -12,14 +12,6 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
--- Create tables
```
ltseeley: Maybe we should add a migration for this change?

examples/gameroom/docker-compose.yaml
```diff
@@ -253,12 +260,46 @@ services:
...
+              --database postgres://admin:admin@splinterd-db-bubba:5432/splinter \
               --service-endpoint 0.0.0.0:8043 \
               --network-endpoint 0.0.0.0:8044 \
               --bind 0.0.0.0:8085 \
-              --insecure
+              --insecure \
+              --enable-biome
         "
+
+    splinterd-db-budda:
```
ltseeley: s/budda/bubba/

examples/gameroom/docker-compose.yaml
```diff
@@ -253,12 +260,46 @@ services:
...
+      container_name: splinterd-db-acme
+      restart: always
+      expose:
+        - 5432
+      ports:
+        - "5435:5432"
+      environment:
+        POSTGRES_USER: admin
+        POSTGRES_PASSWORD: admin
+        POSTGRES_DB: splinter
```
ltseeley: Can you move this up by the acme services for consistency?

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
RyanLassigBanks: I think this type of thing should still be a 400. Gameroom is taking a request from the UI converting it to a request from Biome and sending it to splinterD. Biome would only return a 400 if the user entered invalid info or if gameroom sent the wrong payload. Since the later case can is weeded out with unit tests, I think it makes more sense to surface the former to the UI as a 400 instead of a 500.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
RyanLassigBanks: Yeah this wrong, but it's trying to convey an unauthorized error to the UI, so it should be something like "unauthorized user"

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
RyanLassigBanks: When a user registers via gameroom, part of that process is creating a key for that user via biome key api. So if on login no key is found for that user it would be an error.

examples/gameroom/database/tables/gameroom_tables.sql
RyanLassigBanks: That's a good question. I don't think it's a trivial migration since we'd also need to supply a way of porting existing users from the old database to biome, in addition to providing the migration files in order to do an actual migration. Just adding the migration files seems weird since you'd just be deleting a preexisting users, and adding a script also seems weird since we'd be migrating users for an app that plays tic-tac-toe.

examples/gameroom/daemon/src/rest_api/routes/authenticate.rs
RyanLassigBanks: updated

examples/gameroom/docker-compose.yaml
RyanLassigBanks: updated

examples/gameroom/docker-compose.yaml
RyanLassigBanks: updated

PR #530, opened 2020-02-24
Circuit Read Iterator

peterschwarz:
Convert the vector to an iterator, which provides more flexibility for circuit store back-ends.

### DIFF
--- a/libsplinter/src/circuit/store.rs
+++ b/libsplinter/src/circuit/store.rs
@@ -16,6 +16,25 @@ use std::collections::BTreeMap;
 
 use crate::circuit::Circuit;
 
+/// A filter that matches on aspects of a circuit definition.
+///
+/// Each variant applies to a different field on the circuit defition.
+#[derive(Debug)]
+pub enum CircuitFilter {
+    /// Matches any circuits that have the given node as a member.
+    WithMember(String),
+}
+
+impl CircuitFilter {
+    /// Returns true if the given circuit matches the filter criteria, false otherwise.
+    pub fn matches(&self, circuit: &Circuit) -> bool {
+        match self {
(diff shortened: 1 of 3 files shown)


libsplinter/src/circuit/store.rs
```diff
@@ -36,11 +36,46 @@ impl CircuitFilter {
 }
 
 pub trait CircuitStore: Send + Sync + Clone {
-    fn circuits(&self) -> Result<BTreeMap<String, Circuit>, CircuitStoreError>;
+    /// Return an iterator over the circuits in this store.
+    ///
+    /// A circuit filter may optionally provided, to reduces the results.
+    fn circuits(&self, filter: Option<CircuitFilter>) -> Result<Circuits, CircuitStoreError>;
```
vaporos: circuit/circuits Circuit/Circuits are very closely named, maybe list_circuits()->CircuitList?

vaporos: Should we just return Iterator<Item = Circuit> here so we avoid exposing the implementation completely?

peterschwarz: Unfortunately, no.  I tried to go down that route with event iteration, and the compiler did not like it due to the async layer at the REST API level. We may be able to change that when we more cleanly separate the rest API business logic layer from the underlying http server library. 

It's not a list, though, so I don't think CircuitList isn't the right name.  This is also a pretty common pattern in the standard library:  See `String::chars() -> Chars` as a good example. I could probably be convinced to call it `CircuitIter`, though.

vaporos: CircuitIter seems fine

peterschwarz: Will add an new PR for this.

PR #531, opened 2020-02-24
Fix-up config logging

shannynalayna:
Cleans up the conditionals in the config logging to give a completely raw output for the config values and adds logging of the fully qualified file path whenever one is used / opened throught splinterd.

### DIFF
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -14,6 +14,7 @@
 
 use std::error::Error;
 use std::fmt;
+use std::fs;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread;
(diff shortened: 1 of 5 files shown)


splinterd/src/daemon.rs
```diff
@@ -920,14 +921,20 @@ fn create_node_registry(
...
-                RestApiServerError::StartUpError(format!(
-                    "Failed to initialize YamlNodeRegistry: {}",
-                    err
-                ))
-            })?,
-        )),
+        RegistryConfig::File { registry_file } => {
+            debug!(
+                "Creating node registry with registry file: {:?}",
+                fs::canonicalize(&registry_file)?
```
agunde406: Do these paths get check if they exist before they are logged? If not we should not use ? and instead catch the error and add which path failed in the error message

shannynalayna: Yes, it is checked here  https://github.com/Cargill/splinter/blob/master/splinterd/src/main.rs#L282
and set to None if the file doesn't exist before it gets to this point.

shannynalayna: Same with all of the other config files, all are checked with `is_file` before being logged.

PR #532, opened 2020-02-25
Update scabbard endpoint description

shannynalayna:
Updates the endpoint description for the scabbard batch statuses
endpoint in the openapi spec.

### DIFF
--- a/splinterd/api/static/openapi.yml
+++ b/splinterd/api/static/openapi.yml
@@ -581,8 +581,8 @@ paths:
                 $ref: '#/components/schemas/Error'
 
   /scabbard/{circuit}/{service_id}/batch_statuses:
-    post:
-      description: Send a list of Sabre batches to the specified Scabbard service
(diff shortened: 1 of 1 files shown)


splinterd/api/static/openapi.yml
```diff
@@ -581,8 +581,8 @@ paths:
                 $ref: '#/components/schemas/Error'
 
   /scabbard/{circuit}/{service_id}/batch_statuses:
-    post:
-      description: Send a list of Sabre batches to the specified Scabbard service
+    get:
+      description: Fetches the committed statuses for a set of Sabre batches submitted
```
ltseeley: I would remove "committed" from this description, because it seems to imply that it only gets status when they are committed. Also, why "submitted" at the end of the description? Same for response description below.

shannynalayna: Addressed.

PR #538, opened 2020-02-28
Refactor biome key management database modules to store pattern

shannynalayna:
Refactors the biome::key_management::database::postgres and biome::key_management::store::postgres modules to fit under a biome::key_management::store module. This also genericizes the database helpers to allow for various connection types to be used.


libsplinter/src/biome/key_management/store/diesel/postgres/mod.rs
```diff
@@ -8,22 +8,42 @@
...
 use crate::database::ConnectionPool;
 
+/// Run database migrations to create tables defined in the key management module
+///
+/// # Arguments
+///
+/// * `conn` - Connection to database
+///
+// Allow dead code if biome-key-management feature is not enabled
+#[allow(dead_code)]
```
peterschwarz: Shouldn't the use of this be guarded as well, so that this is never dead code?

libsplinter/src/biome/key_management/store/diesel/operations/insert_key.rs
```diff
@@ -0,0 +1,74 @@
+// Copyright 2018-2020 Cargill Incorporated
```
agunde406: typo in commit message diesel::operatoins::keys -> diesel::operations::keys

libsplinter/src/biome/key_management/store/diesel/postgres/mod.rs
shannynalayna: I addressed this by changing the `diesel/postgres` dependency in the `database` feature to `postgres` (which has `diesel/postgres` as a dependency) so that this module could be behind the `postgres` feature and be compiled in with the `biome-key-management` feature (which has `database` as a feature dependency). Before, this wasn't happening hence the allow dead code

libsplinter/src/biome/key_management/store/diesel/operations/insert_key.rs
shannynalayna: Addressed

PR #539, opened 2020-02-28
Implement reconnection retry for ConnectionManger 

agunde406:
When a connection is discovered to be disconnected, the connection
manager will try to reconnect to the connection. If the
reconnection is not successful, the connection manager will wait
(with an exponential back off) until reconnection is successful.
The max retry frequency is configurable and defaults to 300 seconds.

This commit also simplifies ConnectionManagerNotification to
only include Connected and Disconnected. These messages will
be sent to subscribers when connection and reconnection is
successful. Disconnected is sent if a heartbeat fails and
reconnection is started.

This PR also makes the heartbeat interval configurable.

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -17,11 +17,13 @@ mod messages;
 mod pacemaker;
 
 use std;
+use std::cmp::min;
 use std::collections::HashMap;
 #[cfg(feature = "connection-manager-notification-iter-try-next")]
 use std::sync::mpsc::TryRecvError;
 use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
 use std::thread;
+use std::time::Instant;
 
 pub use error::ConnectionManagerError;
 pub use messages::{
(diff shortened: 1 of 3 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -747,4 +840,90 @@ pub mod tests {
...
+            .next()
+            .expect("Cannot get message from subscriber");
+        assert!(
+            reconnecting_notification
+                == ConnectionManagerNotification::Disconnected {
+                    endpoint: endpoint.clone(),
+                }
+        );
+
+        // receive sucessful reconnect attempt
```
vaporos: s/sucessful/successful/

agunde406: Fixed

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -36,6 +38,8 @@ use crate::transport::Transport;
 
 const DEFAULT_HEARTBEAT_INTERVAL: u64 = 10;
 const CHANNEL_CAPACITY: usize = 15;
+const INITIAL_RETRY_FREQUENCY: u64 = 10;
```
RyanLassigBanks: 10 seconds seems kinda  high for the initial reconnect. You could probably start lower since it grows exponentially (say 1 to 2 seconds).

agunde406: The reconnection attempts are driven by heartbeats, so it does not make sense to have a lower initial retry frequency then the default heartbeat interval, which is also set to 10. Also the first reconnection attempt will happen immediately when the disconnection is discovered.

PR #540, opened 2020-02-28
Update actix, tokio, futures, and actix dependecies in libsplinter

RyanLassigBanks:
This PR updates the tokio, futures, and actix libraries used by libsplinter. It also removes hyper, tokio, and futures from the gameroom daemon and replaces them with reqwest.


RyanLassigBanks:
Closing due to the presence of a bug with extracting the payload from an actix payload.

PR #543, opened 2020-03-02
Add new rules to circuit template API

eloaverona:
### Overview
Add rules for creating services and set application metadata to circuit template API:
- The rule `create-services`, will create a service for each node in the circuit. 
  - The services type will be set to the value of `service-type`. 
  - The service ids are generated starting from the value provided in`first-service`. The following id's will be the following base62 string.  
  - The arguments in the service definition are set via the `service-args` field. If `peer-services` is one of the arguments and the value is set to `$(cs:ALL_OTHER_SERVICES)`, the `peer-services` argument will be set to the id of all other services created by this rule. 

- The rule `set-metadata` sets the application metadata field in the circuit proposal. The `encoding` field determines how the metadata should be serialized, before it is set in the circuit proposal. Currently only json is supported.

This PR also updates the API so that users can load a template into a CircuitCreateTemplate object and work with before turning it into builders to create a circuit.
  
### Example template: 
```yaml
version: v1
args:
    - name: $(a:ADMIN_KEYS)
      required: false
      default: $(a:SIGNER_PUB_KEY)
    - name: $(a::NODES)
      required: true
    - name: $(a:SIGNER_PUB_KEY)
      required: false
    - name: $(a:GAMEROOM_NAME)
      required: true
rules:
    set-management-type:
        management-type: "gameroom"
    create-services:
        service-type: 'scabbard'
        service-args:
        - key: 'admin_keys'
          value: [$(a:ADMIN_KEYS)]
        - key: 'peer_services'
          value: '$(r:ALL_OTHER_SERVICES)'
        first-service: 'a000'
    set-metadata:
        encoding: json
        metadata:
            - key: "scabbard_admin_keys"
              value: ["$(a:ADMIN_KEYS)"]
            - key: "alias"
              value: "$(a:GAMEROOM_NAME)"

```

### To run tests:

```
cd libsplinter
cargo test template --features circuit-template
```

### DIFF
--- a/libsplinter/src/circuit/template/mod.rs
+++ b/libsplinter/src/circuit/template/mod.rs
@@ -16,6 +16,8 @@ mod error;
 mod rules;
 mod yaml_parser;
 
+use std::convert::TryFrom;
+
 pub use error::CircuitTemplateError;
 
 use rules::{RuleArgument, Rules};
@@ -33,7 +35,7 @@ impl CircuitCreateTemplate {
     pub fn from_yaml_file(path: &str) -> Result<Self, CircuitTemplateError> {
         let circuit_template = CircuitTemplate::load_from_file(path)?;
         match circuit_template {
-            CircuitTemplate::V1(template) => Ok(Self::from(template)),
+            CircuitTemplate::V1(template) => Ok(Self::try_from(template)?),
         }
     }
 
@@ -47,21 +49,21 @@ impl CircuitCreateTemplate {
 
         Ok(builders)
     }
-
 }
 
-impl From<v1::CircuitCreateTemplate> for CircuitCreateTemplate {
-    fn from(create_circuit_template: v1::CircuitCreateTemplate) -> Self {
-        CircuitCreateTemplate {
+impl TryFrom<v1::CircuitCreateTemplate> for CircuitCreateTemplate {
+    type Error = CircuitTemplateError;
+    fn try_from(create_circuit_template: v1::CircuitCreateTemplate) -> Result<Self, Self::Error> {
+        Ok(CircuitCreateTemplate {
             version: create_circuit_template.version().to_string(),
             arguments: create_circuit_template
                 .args()
                 .to_owned()
                 .into_iter()
-                .map(RuleArgument::from)
-                .collect(),
+                .map(RuleArgument::try_from)
+                .collect::<Result<_, CircuitTemplateError>>()?,
             rules: Rules::from(create_circuit_template.rules().clone()),
-        }
+        })
     }
 }
 
@@ -99,7 +101,7 @@ mod test {
 
     const EXAMPLE_TEMPLATE_YAML: &[u8] = br##"version: v1
 args:
-    - name: admin-keys
+    - name: $(ADMIN_KEYS)
       required: true
       default: $(a:SIGNER_PUB_KEY)
 rules:
--- a/libsplinter/src/circuit/template/mod.rs
+++ b/libsplinter/src/circuit/template/mod.rs
@@ -49,6 +49,34 @@ impl CircuitCreateTemplate {
 
         Ok(builders)
     }
+
+    pub fn set_argument_value(
+        &mut self,
+        key: &str,
+        value: &str,
+    ) -> Result<(), CircuitTemplateError> {
+        let name = key.to_lowercase();
+        let (index, mut arg) = self
+            .arguments
+            .iter()
+            .enumerate()
+            .find_map(|(index, arg)| {
+                if arg.name() == name {
+                    Some((index, arg.clone()))
+                } else {
+                    None
+                }
+            })
+            .ok_or_else(|| {
+                CircuitTemplateError::new(&format!(
+                    "Argument {} is not defined in the template",
+                    key
+                ))
+            })?;
+        arg.set_user_value(value);
+        self.arguments[index] = arg;
+        Ok(())
+    }
 }
 
 impl TryFrom<v1::CircuitCreateTemplate> for CircuitCreateTemplate {
--- a/libsplinter/src/circuit/template/mod.rs
+++ b/libsplinter/src/circuit/template/mod.rs
@@ -141,16 +141,37 @@ mod test {
 
     const EXAMPLE_TEMPLATE_YAML: &[u8] = br##"version: v1
 args:
-    - name: $(ADMIN_KEYS)
-      required: true
+    - name: $(a:ADMIN_KEYS)
+      required: false
       default: $(a:SIGNER_PUB_KEY)
+    - name: $(a:NODES)
+      required: true
+    - name: $(a:SIGNER_PUB_KEY)
+      required: false
+    - name: $(a:GAMEROOM_NAME)
+      required: true
 rules:
     set-management-type:
-        management-type: "gameroom" "##;
+        management-type: "gameroom"
+    create-services:
+        service-type: 'scabbard'
+        service-args:
+        - key: 'admin-keys'
+          value: [$(a:ADMIN_KEYS)]
+        - key: 'peer-services'
+          value: '$(r:ALL_OTHER_SERVICES)'
+        first-service: 'a000'
+    set-metadata:
+        encoding: json
+        metadata:
+            - key: "scabbard_admin_keys"
+              value: ["$(a:ADMIN_KEYS)"]
+            - key: "alias"
+              value: "$(a:GAMEROOM_NAME)" "##;
 
     /*
      * Verifies that Builders can be parsed from template v1 and correctly
-     * applies the set-management-type rule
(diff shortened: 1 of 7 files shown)


libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +128,182 @@ impl From<v1::ServiceArgument> for ServiceArgument {
...
+                    format!("{}{}", alphanumeric[0..1].to_string(), next_id_sub_string);
+                next_id.replace_range(char_index.., &new_sub_string);
+            }
+        }
+    }
+
+    Err(CircuitTemplateError::new(
+        "Exceed number of services that can be built",
+    ))
+}
```
ltseeley: This could probably be simplified by just incrementing the chars (https://users.rust-lang.org/t/char-how-to-get-next-char-in-rust/17720), checking if the resulting char is alphanumeric (https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric), and overflowing to the next char if necessary.

libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +128,182 @@ impl From<v1::ServiceArgument> for ServiceArgument {
...
+) -> Result<Vec<SplinterServiceBuilder>, CircuitTemplateError> {
+    let peers = service_builders.iter().try_fold::<_, _, Result<Vec<String>, CircuitTemplateError>>(Vec::new(), |mut acc, builder| {
+        let service_id = builder.service_id()
+            .ok_or_else(|| {
+                error!("The service_id must be set before the service argument PEER_SERVICES can be set");
+                CircuitTemplateError::new("Failed to parse template due to an internal error")
+            })?;
+        acc.push(service_id);
+        Ok(acc)
+    })?;
```
ltseeley: Instead of using `.try_fold(...)?` here, you can use `.map(...).collect::<Result<Vec<String>, _>>()?` to simplify things a bit.

libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +128,182 @@ impl From<v1::ServiceArgument> for ServiceArgument {
...
+    ))
+}
+
+fn all_services(
+    service_builders: Vec<SplinterServiceBuilder>,
+) -> Result<Vec<SplinterServiceBuilder>, CircuitTemplateError> {
+    let peers = service_builders.iter().try_fold::<_, _, Result<Vec<String>, CircuitTemplateError>>(Vec::new(), |mut acc, builder| {
+        let service_id = builder.service_id()
+            .ok_or_else(|| {
+                error!("The service_id must be set before the service argument PEER_SERVICES can be set");
```
ltseeley: Is the log here necessary? Might be better to include this message in the error below and let the caller handle it.

libsplinter/src/circuit/template/rules/set_metadata.rs
```diff
@@ -0,0 +1,59 @@
...
+        SetMetadata {
+            metadata: Metadata::from(set_metadata.metadata().clone()),
+        }
+    }
+}
+
+pub(super) enum Metadata {
+    #[allow(dead_code)]
+    // This enum is never directly build only converted from the Yaml version,
+    // #[allow(dead_code)] avoids lint warnings about dead code.
```
ltseeley: If you switch the `From<v1::Metadata>` implementation below to create the variant using `Metadata::Json` instead of `Self::Json`, this lint warning should go away (weird, I know!).

libsplinter/src/circuit/template/mod.rs
```diff
@@ -16,39 +16,61 @@ mod error;
 mod rules;
 mod yaml_parser;
 
-use std::collections::HashMap;
-
 pub use error::CircuitTemplateError;
 
-use rules::CircuitCreateTemplate;
-use yaml_parser::CircuitTemplate;
+use rules::{RuleArgument, Rules};
```
ltseeley: I'm not sure I understand the commit description (I think there may be a typo); could you reword it? This is the "Move CircuitCreateTemplate to module root file" commit

libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +132,187 @@ impl From<v1::ServiceArgument> for ServiceArgument {
...
     }
 }
+
+fn add1_char(c: char) -> Result<String, CircuitTemplateError> {
+    let char_value = c as u32;
+    if char_value >= 'z' as u32 {
+        return Ok('0'.to_string());
+    }
+    let next_char = std::char::from_u32(char_value + 1)
+        .ok_or_else(|| CircuitTemplateError::new("Failed to generate service id"))?;
```
ltseeley: This error message should probably be more specific; if it occurs, it should be clear where in the code it came from. I think this error is impossible to reach, but just in case.

libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +132,187 @@ impl From<v1::ServiceArgument> for ServiceArgument {
         }
     }
 }
+
+fn add1_char(c: char) -> Result<String, CircuitTemplateError> {
```
ltseeley: Can probably just return a `char` instead of `String`

libsplinter/src/circuit/template/rules/create_services.rs
```diff
@@ -50,3 +132,187 @@ impl From<v1::ServiceArgument> for ServiceArgument {
...
+}
+
+fn get_next_service_id(current_id: &str) -> Result<String, CircuitTemplateError> {
+    generate_id(current_id, current_id.len() - 1)
+}
+
+fn generate_id(current_id: &str, index: usize) -> Result<String, CircuitTemplateError> {
+    let mut next_id = current_id.to_string();
+    let character = char::from_str(&next_id[index..=index])
+        .map_err(|_| CircuitTemplateError::new("Failed to generate service id"))?;
```
ltseeley: This error should be more specific

libsplinter/src/circuit/template/rules/set_metadata.rs
```diff
@@ -12,13 +12,71 @@
...
+                                            } else {
+                                                value.to_string()
+                                            };
+                                            acc.push(value);
+                                            Ok(acc)
+                                        },
+                                    )?;
+                                    acc.push_str(&format!(
+                                        "\"{}\":{:?},",
+                                        metadata.key, processed_values
```
ltseeley: Should probably use the `join` method here again instead of using the debug formatting.

libsplinter/src/circuit/template/rules/set_metadata.rs
```diff
@@ -12,13 +12,71 @@
...
+        &self,
+        builder: CreateCircuitBuilder,
+        template_arguments: &[RuleArgument],
+    ) -> Result<CreateCircuitBuilder, CircuitTemplateError> {
+        match &self.metadata {
+            Metadata::Json { metadata } => {
+                let mut metadata = metadata
+                    .iter()
+                    .try_fold::<_, _, Result<_, CircuitTemplateError>>(
+                        "{[".to_string(),
```
ltseeley: Would probably be easier to do this by just constructing a `Vec` in the `try_fold`, `collecting`ing, and then `join`ing (plus it would get rid of any trailing commas).

libsplinter/src/circuit/template/rules/create_services.rs
ltseeley: Looks good!

libsplinter/src/circuit/template/rules/create_services.rs
ltseeley: Seems fine

libsplinter/src/circuit/template/mod.rs
eloaverona: Addressed

libsplinter/src/circuit/template/rules/create_services.rs
eloaverona: Addressed

libsplinter/src/circuit/template/rules/create_services.rs
eloaverona: Addressed

libsplinter/src/circuit/template/rules/create_services.rs
eloaverona: Addressed

libsplinter/src/circuit/template/rules/set_metadata.rs
eloaverona: Addressed

libsplinter/src/circuit/template/rules/set_metadata.rs
eloaverona: Addressed

PR #545, opened 2020-03-02
Validate circuit vote before attempting consensus

peterschwarz:
Like circuit proposals, validate the contents of a circuit vote before it can be submitted to the consensus system.  This allows the client to detect rejected  votes at submission time, instead of at consensus time, long after the HTTP request has completed.

Additionally

- fix a typo
- ignore IO errors when scanning the path for pandoc

To test:

1. Build gameroom with standard flags
2. Create Alice, with her key in the registry as normal
3. Propose a gameroom
4. Create Charlie, with a new key
5. Accept the invite
6. An error message should be displayed immediately

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -31,11 +31,25 @@ fn main() -> Result<(), BuildError> {
         .map_err(|_| BuildError("Unable to read PATH environment variable".into()))?;
     let mut pandoc_exist = false;
     for path in paths.split(':') {
-        for entry in fs::read_dir(path)
-            .map_err(|err| BuildError(format!("Cannot check if pandoc is in PATH: {}", err)))?
-        {
-            let entry = entry
-                .map_err(|err| BuildError(format!("Cannot check if file is pandoc: {}", err)))?;
+        let entries = match fs::read_dir(path) {
+            Ok(entries) => entries,
+            Err(err) => {
+                // skip a directory in the path that cannot be read.
+                println!("Unable to read path entry {}: {}", path, err);
+                continue;
+            }
+        };
+
+        for entry in entries {
+            let entry = match entry {
+                Ok(entry) => entry,
+                Err(err) => {
+                    // skip an entry in the path that cannot be read.
+                    println!("Unable to read entry in {}: {}", path, err);
+                    continue;
+                }
+            };
+
             let path = entry.path();
             if path.ends_with("pandoc") {
                 pandoc_exist = true;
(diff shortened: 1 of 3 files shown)


cli/build.rs
```diff
@@ -31,11 +31,25 @@ fn main() -> Result<(), BuildError> {
...
+                continue;
+            }
+        };
+
+        for entry in entries {
+            let entry = match entry {
+                Ok(entry) => entry,
+                Err(err) => {
+                    // skip an entry in the path that cannot be read.
+                    println!("Unable read entry in {}: {}", path, err);
```
jsmitchell: s/Unable read/Unable to read/

libsplinter/src/admin/service/shared.rs
```diff
@@ -664,6 +664,38 @@ impl AdminServiceShared {
...
+                let proposal_vote = payload.get_circuit_proposal_vote();
+
+                // validate vote proposal
+                // check that the circuit proposal exists
+                let circuit_proposal = self
+                    .get_proposal(proposal_vote.get_circuit_id())
+                    .map_err(|err| {
+                        ServiceError::UnableToHandleMessage(Box::new(
+                            AdminSharedError::ValidationFailed(format!(
+                                "error occured when trying to get proposal {}",
```
jsmitchell: s/occured/occurred/

peterschwarz: Damn, I fixed that typo in one place, but reproduced it else where...

Fixed

cli/build.rs
peterschwarz: Fixed

libsplinter/src/admin/service/shared.rs
jsmitchell: s/else where/elsewhere/

libsplinter/src/admin/service/shared.rs
```diff
@@ -664,6 +664,38 @@ impl AdminServiceShared {
                 self.propose_circuit(payload, "local".to_string())
             }
             CircuitManagementPayload_Action::CIRCUIT_PROPOSAL_VOTE => {
+                let proposal_vote = payload.get_circuit_proposal_vote();
```
agunde406: Can you add a log statement somewhere so we can see why the vote failed, I don't see any indication in the logs (maybe we should add one for circuit proposal too?)

peterschwarz: Added as a new commit (at the submit handler layer).

ltseeley:
Tested locally

PR #546, opened 2020-03-02
Fix user DELETE route to match HTTP 1.1 spec

shannynalayna:
Fixes the `/biome/users/{id}` DELETE route to not take in
a request body, per HTTP 1.1 specificatiions. Instead, this
route is validated using JWT tokens.

Also adds passing the BiomeRestConfig and SecretManager to the
make_user_routes function to allow for the JWT token validation.

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/user.rs
+++ b/libsplinter/src/biome/rest_api/actix/user.rs
@@ -22,9 +22,15 @@ use crate::rest_api::{into_bytes, ErrorResponse, Method, ProtocolVersionRangeGua
 use crate::biome::credentials::store::{
     diesel::SplinterCredentialsStore, CredentialsStore, CredentialsStoreError,
 };
-use crate::biome::user::store::{diesel::SplinterUserStore, SplinterUser, UserStore};
+use crate::biome::rest_api::BiomeRestConfig;
+use crate::biome::secrets::SecretManager;
+use crate::biome::user::store::{
+    diesel::SplinterUserStore, SplinterUser, UserStore, UserStoreError,
+};
 
+use super::super::resources::authorize::AuthorizationResult;
 use super::super::resources::credentials::UsernamePassword;
+use super::authorize::authorize_user;
 
 /// Defines a REST endpoint to list users from the db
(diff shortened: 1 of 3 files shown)


libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -277,69 +277,35 @@ fn add_delete_user_method(
...
-        let username_password = match serde_json::from_slice::<UsernamePassword>(&bytes) {
-            Ok(val) => val,
+    match authorize_user(&request, &user_id, &secret_manager, &rest_config) {
+        AuthorizationResult::Authorized => match user_store.remove_user(&user_id) {
+            Ok(()) => Box::new(
+                HttpResponse::Ok()
+                    .json(json!({ "message": "User deleted sucessfully" }))
+                    .into_future(),
+            ),
             Err(err) => {
```
eloaverona: You should probably check what error is returned. Currently if the user does not exists in the database (for example it has already been deleted) the REST API is returning a 500 error. It should probably return a 404.

libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -277,69 +277,35 @@ fn add_delete_user_method(
...
-                                .json(ErrorResponse::internal_error())
-                                .into_future()
-                        }
-                    }
-                } else {
-                    HttpResponse::BadRequest()
-                        .json(ErrorResponse::bad_request("Invalid password"))
-                        .into_future()
-                }
+                debug!("Failed to delete user in database {}", err);
```
eloaverona: Log InternalServerErrors as an error not a debug

libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -277,69 +277,35 @@ fn add_delete_user_method(
...
+        },
+        AuthorizationResult::Unauthorized(msg) => {
+            return Box::new(
+                HttpResponse::Unauthorized()
+                    .json(ErrorResponse::unauthorized(&msg))
+                    .into_future(),
+            )
+        }
+        AuthorizationResult::Failed => {
+            return Box::new(
```
eloaverona: Add an error log for this error

PR #547, opened 2020-03-03
Add more doc comments to biome module

shannynalayna:
Adds more Rust doc comments to the biome module for existing public APIs.


libsplinter/src/biome/credentials/store/mod.rs
```diff
@@ -60,7 +60,7 @@ pub struct UserCredentialsBuilder {
 }
 
 impl UserCredentialsBuilder {
-    /// Sets the user_id for the credentials belong to
+    /// Sets the user_id for user the credentials belong to
```
ltseeley: Probably meant "...for the user the credentials..."

libsplinter/src/biome/rest_api/actix/login.rs
```diff
@@ -28,6 +28,11 @@ use crate::biome::sessions::{AccessTokenIssuer, ClaimsBuilder, TokenIssuer};
 use super::super::resources::credentials::UsernamePassword;
 
 /// Defines a REST endpoint for login
+/// The payload should be in the JSON format:
```
ltseeley: If you want this to display as a new paragraph in the rust docs, you'll want a blank line between these lines.

PR #548, opened 2020-03-03
Proposal read doc updates

ltseeley:


libsplinter/src/admin/rest_api/actix/proposals_read.rs
```diff
@@ -11,6 +11,8 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 
+//! This module provides the `GET /admin/proposals` endpoint for listing circuit proposals.
```
chenette: Per best practices and consistency, each module description should start with a verb (Provides), not "This module". 

For an example:
https://github.com/Cargill/splinter/blob/master/libsplinter/src/rest_api/cors.rs 
https://docs.rs/splinter/0.3.12/splinter/rest_api/cors/index.html

ltseeley: Updated

PR #552, opened 2020-03-04
Use RestApiBuilder in circuit-read tests

ltseeley:
Updates the tests for the `/admin/circuits` endpoints to not use the
actix-specific testing code. This change will allow us to carry the unit
tests over if we decide to re-implement the REST API without actix.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -66,6 +66,7 @@ uuid = { version = "0.7", features = ["v4"]}
 zmq = { version = "0.9", optional = true }
 
 [dev-dependencies]
+reqwest = { version = "0.10", features = ["blocking", "json"] }
 serial_test = "0.3"
 tempdir = "0.3"
 
--- a/libsplinter/src/admin/rest_api/actix/circuits.rs
+++ b/libsplinter/src/admin/rest_api/actix/circuits.rs
@@ -164,36 +164,30 @@ fn query_list_circuits<T: CircuitStore + 'static>(
 mod tests {
     use super::*;
 
-    use crate::actix_web::{
-        http::{header, StatusCode},
-        test, web, App,
-    };
+    use reqwest::{blocking::Client, StatusCode, Url};
+
     use crate::circuit::{
         directory::CircuitDirectory, AuthorizationType, Circuit, DurabilityType, PersistenceType,
         Roster, RouteType, ServiceDefinition, SplinterState,
     };
-    use crate::rest_api::paging::Paging;
+    use crate::rest_api::{paging::Paging, RestApiBuilder, RestApiShutdownHandle};
     use crate::storage::get_storage;
 
     #[test]
     /// Tests a GET /admin/circuits request with no filters returns the expected circuits.
     fn test_list_circuits_ok() {
-        let splinter_state = filled_splinter_state();
+        let (_shutdown_handle, _join_handle, bind_url) =
+            run_rest_api_on_open_port(vec![make_list_circuits_resource(filled_splinter_state())]);
 
-        let mut app = test::init_service(
-            App::new().data(splinter_state.clone()).service(
-                web::resource("/admin/circuits")
-                    .route(web::get().to_async(list_circuits::<SplinterState>)),
-            ),
-        );
-
-        let req = test::TestRequest::get().uri("/admin/circuits").to_request();
-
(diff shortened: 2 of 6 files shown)


libsplinter/src/admin/rest_api/actix/circuits.rs
```diff
@@ -164,36 +164,39 @@ fn query_list_circuits<T: CircuitStore + 'static>(
...
+            .expect("Failed to build REST API")
+            .run()
+            .expect("Failed to run REST API");
 
-        let resp = test::call_service(&mut app, req);
+        let url = Url::parse(&format!("http://{}/admin/circuits", bind_url))
+            .expect("Failed to parse URL");
+        let req = Client::new()
+            .get(url)
+            .header("SplinterProtocolVersion", protocol::ADMIN_PROTOCOL_VERSION);
```
ltseeley: Interested to hear peoples' thoughts on if we should also add tests for other protocol versions; may not be necessary right now since everything is 1, but when these versions get updated tests should probably be added for old versions.

vaporos: Agree, but is there anything to do now?

libsplinter/src/admin/rest_api/actix/circuits.rs
```diff
@@ -203,83 +206,95 @@ mod tests {
...
     fn test_list_circuit_with_filters_ok() {
-        let splinter_state = filled_splinter_state();
-
-        let mut app = test::init_service(
-            App::new().data(splinter_state.clone()).service(
-                web::resource("/admin/circuits")
-                    .route(web::get().to_async(list_circuits::<SplinterState>)),
-            ),
-        );
+        let bind_url = "127.0.0.1:8111";
```
vaporos: I wonder if we should bind to a random high port within a range.

libsplinter/src/admin/rest_api/actix/circuits.rs
ltseeley: Don't think so; we just have to be sure (as reviewers) going forward that we make sure these tests get added with new versions.

libsplinter/src/admin/rest_api/actix/circuits.rs
ltseeley: As in randomly generated ports? Seems like it could cause ephemeral failures (probably very infrequent, but still).

vaporos: Yeah, we would need some way to recover (then I suppose it doesn't need to be random).

ltseeley: Updated to programmatically find an open port in a range

libsplinter/src/admin/rest_api/actix/circuits.rs
```diff
@@ -396,4 +375,26 @@ mod tests {
...
+                    .with_bind(&bind_url)
+                    .add_resources(resources.clone())
+                    .build()
+                    .expect("Failed to build REST API")
+                    .run();
+                match result {
+                    Ok((shutdown_handle, join_handle)) => {
+                        Some((shutdown_handle, join_handle, bind_url))
+                    }
+                    Err(_) => None,
```
vaporos: This should match the error which occurs when binding itself fails (as opposed to anything that can happen), so we stop the test with the error if its something unrelated to something already using the port.

libsplinter/Cargo.toml
```diff
@@ -111,9 +112,9 @@ biome-credentials = ["biome", "biome-user", "database", "bcrypt"]
 biome-key-management = ["biome", "database"]
 biome-notifications = ["biome", "database"]
 biome-user = ["biome", "database"]
-circuit-read = []
+circuit-read = ["rest-api", "rest-api-actix"]
```
vaporos: rest-api-actix should not be automatically pulled in; instead, both circuit-read and rest-api-actix should be required to turn on these implementations.

ltseeley: Fixed

libsplinter/src/admin/rest_api/actix/circuits.rs
ltseeley: Added a commit to differentiate a bind error in the REST API's `run` method and used that in the tests.

PR #553, opened 2020-03-04
Remove return to fix clippy warning

shannynalayna:
Removes `return` from a return statement in the build file
as this was causing a clippy warning as it is unnecessary.


peterschwarz:
> The environment variable is `SPLINTER_FORCE_PANDOC`, but the variable used within this file is just `FORCE_PANDOC`, so the error message does match the environment variable that it checks.

Ahh, it was confusing due to the const - that const seems unnecessary, but not in the scope of this PR.

PR #555, opened 2020-03-05
Update dockerfiles based on splinter-dev to include a tag

rbuysse:
This will allow us to version updates to the base image so significant changes
don't silently break building splinter components.


desktophero:
albeit late :shipit:

PR #556, opened 2020-03-05
Add response body to biome key management routes

shannynalayna:
Adds response bodies to the `/biome/users/{user_id}/keys` POST route and the  `/biome/register` POST route.

### DIFF
--- a/libsplinter/src/biome/key_management/mod.rs
+++ b/libsplinter/src/biome/key_management/mod.rs
@@ -19,12 +19,11 @@ pub mod store;
 use store::diesel::models::KeyModel;
 
 // Represents a public and private key pair
-#[derive(Serialize)]
 pub struct Key {
-    public_key: String,
-    encrypted_private_key: String,
-    user_id: String,
-    display_name: String,
+    pub public_key: String,
+    pub encrypted_private_key: String,
+    pub user_id: String,
+    pub display_name: String,
 }
 
 impl Key {
@@ -64,6 +63,7 @@ impl From<KeyModel> for Key {
         }
     }
 }
+
 impl Into<KeyModel> for Key {
     fn into(self) -> KeyModel {
         KeyModel {
--- a/libsplinter/src/biome/key_management/mod.rs
+++ b/libsplinter/src/biome/key_management/mod.rs
@@ -19,6 +19,7 @@ pub mod store;
 use store::diesel::models::KeyModel;
 
 // Represents a public and private key pair
+#[derive(Clone)]
 pub struct Key {
     pub public_key: String,
     pub encrypted_private_key: String,
--- a/libsplinter/src/biome/rest_api/actix/mod.rs
+++ b/libsplinter/src/biome/rest_api/actix/mod.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(feature = "biome-key-management")]
+#[cfg(any(feature = "biome-key-management", feature = "biome-credentials"))]
 pub(crate) mod authorize;
 #[cfg(feature = "biome-key-management")]
 pub(super) mod key_management;
--- a/libsplinter/src/biome/rest_api/resources/credentials.rs
+++ b/libsplinter/src/biome/rest_api/resources/credentials.rs
@@ -19,3 +19,9 @@ pub(crate) struct UsernamePassword {
     pub username: String,
     pub hashed_password: String,
 }
+
+#[derive(Serialize)]
+pub(crate) struct NewUser<'a> {
+    pub user_id: &'a str,
+    pub username: &'a str,
+}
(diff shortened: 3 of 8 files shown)


libsplinter/src/biome/rest_api/resources/credentials.rs
```diff
@@ -14,7 +14,7 @@
 
 //! Defines credentials used to register and authenticate users.
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Serialize)]
```
peterschwarz: This shouldn't actually be sent back as is - we should minimize the number of times the hashed password should be sent over the wire.

libsplinter/src/biome/key_management/mod.rs
```diff
@@ -21,7 +21,7 @@ pub(in crate::biome) mod store;
 use database::postgres::models::KeyModel;
 
 // Represents a public and private key pair
-#[derive(Serialize)]
+#[derive(Serialize, Clone)]
```
peterschwarz: This is the internal (native, in other parlance) object for Key.  The key-management rest api should really use a separate serialized view, so we can have a clean separation between what is and isn't exposed over the wire.

libsplinter/src/biome/rest_api/resources/credentials.rs
shannynalayna: Addressed by creating a `NewUser` object which is now returned by the /register endpoint

libsplinter/src/biome/key_management/mod.rs
shannynalayna: Also created an object, `ResponseKey` for this object. And converted another endpoint which had been returning this internal Key object to the new ResponseKey

libsplinter/src/biome/rest_api/actix/mod.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(feature = "biome-key-management")]
+#[cfg(any(feature = "biome-key-management", feature = "biome-credentials"))]
```
peterschwarz: Should this be `all`?  Based on your commit message it sounds like both are required, not either.

libsplinter/src/biome/rest_api/resources/credentials.rs
```diff
@@ -19,3 +19,9 @@ pub(crate) struct UsernamePassword {
     pub username: String,
     pub hashed_password: String,
 }
+
+#[derive(Deserialize, Serialize)]
+pub(crate) struct NewUser {
```
peterschwarz: Given how it's used - you can probably do this as
```rust
struct NewUser<'a> {
    pub user_id: &'a str,
    pub username: &'a str,
}
```
Plus, it doesn't need to be deserializable , since it is only for output.

libsplinter/src/biome/key_management/mod.rs
```diff
@@ -21,10 +21,10 @@ use store::diesel::models::KeyModel;
 // Represents a public and private key pair
 #[derive(Serialize)]
```
peterschwarz: Does this struct still need to be `Serialize`?

libsplinter/src/biome/rest_api/resources/key_management.rs
```diff
@@ -26,3 +28,22 @@ pub(crate) struct UpdatedKey {
     pub public_key: String,
     pub new_display_name: String,
 }
+
+#[derive(Serialize, Clone)]
+pub(crate) struct ResponseKey {
```
peterschwarz: The same pattern could apply here:
```rust
struct ResponseKey<'a>{
    public_key: &'a str,
    ...
}
```

Given the `From` implementation, you probably don't need the fields to be `pub`, either.

libsplinter/src/biome/rest_api/actix/mod.rs
shannynalayna: No, it's either. The authorize module is being used separately by both the key management and credentials routes.

libsplinter/src/biome/rest_api/resources/key_management.rs
shannynalayna: I don't think it's possible to implement this one using the same pattern, as the `key` object used to fill in the values for the ResponseKey is being consumed before the ResponseKey is returned. The fields probably don't have to be pub, though.

libsplinter/src/biome/key_management/mod.rs
shannynalayna: No, it does not. Addressed.

libsplinter/src/biome/rest_api/resources/credentials.rs
shannynalayna: Addressed.

PR #558, opened 2020-03-08
Socket transport versioning

peterschwarz:
A versioned frame is used for reading and writing data over socket connections (or any binary Read/Write stream, for that matter).

The version for a connection is established via small negotiation exchange, where the (min, max) pair of version are sent by the outbound (local) stream, and the compatible version is sent by the
inbound (remote) stream.

Frame versions have an associated `FrameHeader`.  This header specifies things about its version, the length of the frame as well as other details about the frame.  The V1 header specifies length and a checksum of the header, to ensure that the data received is the correct length,
and the length value is not a random value from anywhere along the stream.

Frames themselves are containers for the data.  These are either owned (for received data), or borrowed (for sent data). Only the sent data needs to know the outbound version.

This PR deprecates the `TlsConnection` constructor, as the version negotiation should be performed based on the direction of the connection. This is managed by the transport.

**Note: This is a breaking change**

It is not possible to accurately detect that a connection has not negotiated the versions correctly, as older versions will not send the the initial handshake messages.


libsplinter/src/transport/frame.rs
```diff
@@ -11,3 +11,651 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use std::thread;
+
+    #[test]
+    fn read_version_1_zero_length() {
```
agunde406: Can you add descriptions to each test?

peterschwarz: Done. And added one missing test.

PR #559, opened 2020-03-09
Add /biome/users/{user_id}/keys/{pub_key} GET and DELETE routes

shannynalayna:
Adds implementations for the `fetch_key` and `remove_key` methods to the KeyStore as well as the corresponding endpoints. This adds a new endpoint `/biome/users/{user_id}/keys/{pub_key}` which has a GET and a DELETE endpoint.

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/key_management.rs
+++ b/libsplinter/src/biome/rest_api/actix/key_management.rs
@@ -32,7 +32,7 @@ use super::super::resources::authorize::AuthorizationResult;
 use super::super::resources::key_management::{NewKey, ResponseKey, UpdatedKey};
 use super::authorize::authorize_user;
 
-/// Defines the `/biome/users/{user_id}/keys` REST resource for managing keys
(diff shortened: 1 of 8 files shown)


libsplinter/src/biome/rest_api/actix/key_management.rs
```diff
@@ -278,3 +278,192 @@ fn handle_patch(
         }))
     })
 }
+
+/// Defines a REST endpoint for managing keys including fetching and deleting a user's key
+pub fn make_key_management_route_with_public_key(
+    rest_config: Arc<BiomeRestConfig>,
```
vaporos: Can we get rid of these Arc?

shannynalayna: I added a story to review the use of Arc within Biome: https://zenhub.cglcloud.com/app/workspaces/product-core-5d37298b0add7306b30efde2/issues/digitalsupplychain/product-core/2056, as I don't think this is necessary within the scope of this story/pr

PR #563, opened 2020-03-10
Update admin service endpoint documentation

ltseeley:
Add endpoint documentation to the `RestResourceProvider` implementations
of `AdminService` and `CircuitResourceProvider`. This documentation
mentions that one of the REST API backend features must be enabled for
these endpoints to be available, even though `rest-api-actix` is the
only backend currently.

Even though some of the endpoints are guarded by the `circuit-read` and
`proposal-read` features, these are not called out in the documentation
because these features will likely be removed when they are stabilized.

### DIFF
--- a/libsplinter/src/admin/rest_api/mod.rs
+++ b/libsplinter/src/admin/rest_api/mod.rs
@@ -37,6 +37,18 @@ use self::actix::submit::make_submit_route;
 #[cfg(feature = "rest-api-actix")]
 use self::actix::ws_register_type::make_application_handler_registration_route;
 
+/// The admin service provides the following endpoints as REST API resources:
+///
+/// * `GET /ws/admin/register/{type}` - Register as an application authorization handler for the
+///   given circuit management type
+/// * `POST /admin/submit` - Submit a circuit management payload
+/// * `GET /admin/proposals` - List circuit proposals in Splinter's state
+/// * `GET /admin/proposals/{circuit_id}` - Fetch a specific circuit proposal in Splinter's state
+///   by circuit ID
+///
+/// These endpoints are only available if the following REST API backend feature is enabled:
+///
+/// * `rest-api-actix`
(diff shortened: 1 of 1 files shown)


libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -63,6 +76,11 @@ impl RestResourceProvider for AdminService {
 /// * `GET /admin/circuits` - List circuits in Splinter's state
 /// * `GET /admin/circuits/{circuit_id}` - Fetch a specific circuit in Splinter's state by circuit
 ///   ID
+///
+/// These endpoints are only available if one of the following REST API backend features is
+/// enabled:
+///
+/// * `rest-api-actix`
```
chenette: Are more features expected for this list?  If not, I'd pull the bullet item into the sentence; for example:  ".... only available if the REST API back-end feature `rest-api-actix` is enabled."  Otherwise, I'd change "one of the following ... features" to "the following ... feature" (so that the text matches the number of items in the list). 

Ditto for the similar doc comments on lines 102-105.

ltseeley: Yes, more features are expected for this list.

chenette: In that case, I recommend correcting the text to match the number of items currently on the list: "the following REST API back-end feature is enabled"). 

Later, when more bullets are added, the text can change to "any of the following...."

ltseeley: Updated

eloaverona: Per @chenette comment, should this say `REST API backend feature` instead of `REST API backend features`? Same for the comment bellow.

ltseeley: Oof. Fixed.

PR #565, opened 2020-03-11
Move biome migrations

RyanLassigBanks:
This PR merges the diesel migrations scripts, schema.rs files, and models.rs files into a single module within biome called database.


libsplinter/src/biome/credentials/store/diesel/mod.rs
```diff
@@ -14,8 +14,6 @@
 
 pub(in crate::biome) mod models;
 mod operations;
-#[cfg(feature = "postgres")]
```
ltseeley: Typos in commit msg: s/into in/into/ and s/hourse/house/

RyanLassigBanks: Added "Part of centralizing migrations" to all commit messages

libsplinter/src/biome/mod.rs
```diff
@@ -32,6 +32,8 @@ pub mod notifications;
 
 #[cfg(feature = "rest-api")]
 pub mod rest_api;
+
+pub mod database;
```
peterschwarz: This should probably be guarded by `#[cfg(feature="postgres")]` (and an use of `any` once we add more db targets).

It also should be above rest_api for alphabetical sorting

cli/Cargo.toml
```diff
@@ -55,10 +55,6 @@ experimental = [
     "circuit",
     "health",
     "database",
-    "database-migrate-biome-credentials",
```
peterschwarz: There should be a single `"database-migrate-biome"` feature, as biome is now the component boundary.  When splinter admin state is in a database, it will have an equivalent `"database-migrate-admin"` which can be enabled separately from the biome migrations.

libsplinter/src/biome/mod.rs
RyanLassigBanks: All the components within database are already feature guarded. Currently database contains `models`, `schemas`, and `run_postgres_migrations`. The individual structs in `models` and `schemas` are feature guarded, and `run_postgres_migrations` is guarded under the postgres feature. Because of this I don't think database itself needs a guard or if it does have a guard it shouldn't be postgres since it contains components used by other biome modules.

peterschwarz: Is database the right name for this module then?

cli/Cargo.toml
RyanLassigBanks: Added database-migrate-biome feature to cli

libsplinter/src/biome/mod.rs
RyanLassigBanks: I think so. It contains the schemas, models, and migrations for biome databases, so it seems appropriate to me, but I'm open to suggestions.

peterschwarz: It should probably match the other biome features:

`biome::storage`
`biome::user::storage`
etc. 

This might make sense as diesel is a cross cutting concern, and should be enabled with the postgres flag.

PR #566, opened 2020-03-11
Enforce new circuit/service ID formats in circuit CLI

ltseeley:
This PR updates the circuit CLI to submit circuit proposals with the new circuit and service ID formats.

The new circuit ID format is a string composed of two, 5 character base62 strings joined by a dash. Example: aB3dE-7gH0j

The new service ID format is a 4 character base62 string.

To test, attempt to create a circuit using the circuit CLI with service IDs that are invalid (this should fail with a helpful error message). Then create a circuit proposal with valid service IDs and check that the circuit ID generated for the new circuit is in the correct format.


libsplinter/src/admin/service/messages/builders.rs
```diff
@@ -129,28 +129,28 @@ impl CreateCircuitBuilder {
     }
 
     pub fn build(self) -> Result<CreateCircuit, BuilderError> {
-        let circuit_id = self.circuit_id.ok_or_else(|| {
```
eloaverona: I think there's a typo in the commit message. "When a service ID is not given to the `CreateCircuitBuilder`" should be "When a circuit ID is not given to the `CreateCircuitBuilder`"

libsplinter/src/base62.rs
```diff
@@ -11,3 +11,77 @@
...
 // limitations under the License.
+
+#[cfg(feature = "circuit-template")]
+use std::str::FromStr as _;
+
+#[cfg(feature = "circuit-template")]
+/// Get the next base62 string of the same length; if the given string is the maximum value (all
+/// characters are `z`), return `None`.
+pub fn next_base62_string(string: &str) -> Option<String> {
+    if string.is_empty() {
```
eloaverona: I think next_base62_string should check that the input string only contains base62 characters.

libsplinter/src/admin/service/messages/builders.rs
ltseeley: Fixed

libsplinter/src/base62.rs
ltseeley: Updated to check if string is base62 and return an error if it's not.

ltseeley:
> I wonder if the CLI code that sets up the services builders should also check that the service-id is valid before attempting to build, as part of validating the user input early. It might be overkill.

Doesn't seem totally necessary to me; to the user, it makes no difference since they'll get a message either way telling them the value was incorrect.

libsplinter/Cargo.toml
```diff
@@ -48,7 +48,7 @@ mio-extras = "2"
 openssl = "0.10"
 percent-encoding = { version = "2.0", optional = true }
 protobuf = "2"
-rand = { version = "0.7", optional = true }
+rand = "0.7"
```
peterschwarz: This should probably still be optional and required by the "circuit-template" feature.

ltseeley: The `generate_random_base62_string` function that uses `rand` isn't used by the `circuit-template` feature; it's used by the circuit and service builders in the admin service, which aren't behind any feature flags.

PR #569, opened 2020-03-12
ConnnectionManager improvements 

agunde406:
This PR contains bug fixes and improvements to the connection manager that are required for the PeerManager. This includes adding a Heartbeat notifications and removing reference counting. Reference counting will be the job of a higher level component such as the PeerManager.

This is required for #568


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 mod error;
-mod notification;
+pub mod notification;
```
peterschwarz: This is unnecessary as they are referenced via `pub use` on line 27.

PR #572, opened 2020-03-16
Enforce circuit/service id formats

ltseeley:
Updates the admin service's circuit validation to check that the circuit
and service IDs of proposed circuits are valid.

Also updates gameroom to use valid circuit and service IDs.

To test, run gameroom with `docker-compose -f examples/gameroom/docker-compose.yaml up --build` and verify that you can create a gameroom successfully.


examples/gameroom/daemon/src/rest_api/routes/gameroom.rs
```diff
@@ -22,7 +22,7 @@ use gameroom_database::{
 use openssl::hash::{hash, MessageDigest};
 use protobuf::Message;
 use splinter::admin::messages::{
-    CreateCircuit, CreateCircuitBuilder, SplinterNode, SplinterService,
+    CreateCircuit, CreateCircuitBuilder, SplinterNode, SplinterServiceBuilder,
```
agunde406: Typo in commit messages grXX -> gmXX

examples/gameroom/daemon/src/authorization_handler/mod.rs
```diff
@@ -1323,7 +1326,7 @@ mod test {
     fn get_new_gameroom_service(circuit_id: &str, timestamp: SystemTime) -> NewGameroomService {
         NewGameroomService {
             circuit_id: circuit_id.to_string(),
-            service_id: "scabbard_123".to_string(),
+            service_id: "gr00".to_string(),
```
agunde406: should these be gm based on the previous commit?

examples/gameroom/daemon/src/rest_api/routes/gameroom.rs
ltseeley: Should be `gr` (for gameroom). Fixed the code.

examples/gameroom/daemon/src/authorization_handler/mod.rs
ltseeley: Fixed previous commit (should be `gr`)

PR #578, opened 2020-03-17
Add Refresh token endpoint and update biome login endpoint

RyanLassigBanks:
In this PR

1. Add new endpoint for requesting new authentication tokens `POST /biome/refresh`

The purpose of this endpoint is to offer a way to get new auth tokens after the a client's auth token has expired.

2. Adds refresh token database tables
3. Adds RefreshTokenStore trait
4. Adds SplinterRefrshTokenStore object
5. Updates `POST /biome/login` to return a refresh token
6. Removes `user_id` from key_management path
7. Refactors `authorize_users` to accept a validation object and not a user id

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -90,6 +90,7 @@ experimental = [
     "biome-credentials",
     "biome-key-management",
     "biome-notifications",
+    "biome-refresh-tokens",
     "biome-rest-api",
     "biome-user",
     "circuit-read",
@@ -115,6 +116,7 @@ biome = ["database"]
 biome-credentials = ["biome", "biome-user", "database", "bcrypt"]
 biome-key-management = ["biome", "database"]
 biome-notifications = ["biome", "database"]
+biome-refresh-tokens = []
 biome-rest-api = ["biome"]
 biome-user = ["biome", "database"]
 circuit-read = []
--- a/libsplinter/src/biome/refresh_tokens/store/diesel/models.rs
+++ b/libsplinter/src/biome/refresh_tokens/store/diesel/models.rs
@@ -1,0 +1,31 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::schema::refresh_tokens;
+
+#[derive(Queryable, Identifiable, PartialEq, Debug)]
+#[table_name = "refresh_tokens"]
+#[primary_key(id)]
+pub struct RefreshToken {
+    pub id: i64,
+    pub user_id: String,
+    pub token: String,
+}
+
+#[derive(AsChangeset, Insertable, PartialEq, Debug)]
+#[table_name = "refresh_tokens"]
+pub struct NewRefreshToken<'a> {
+    pub user_id: &'a str,
+    pub token: &'a str,
+}
--- a/libsplinter/src/biome/rest_api/actix/key_management.rs
+++ b/libsplinter/src/biome/rest_api/actix/key_management.rs
@@ -25,10 +25,10 @@ use crate::biome::rest_api::resources::key_management::{NewKey, ResponseKey, Upd
 use crate::biome::rest_api::BiomeRestConfig;
 use crate::futures::{Future, IntoFuture};
 use crate::protocol;
-use crate::rest_api::secrets::SecretManager;
 use crate::rest_api::{
     into_bytes, ErrorResponse, HandlerFunction, Method, ProtocolVersionRangeGuard, Resource,
 };
+use crate::rest_api::{secrets::SecretManager, sessions::default_validation};
 
 /// Defines a REST endpoint for managing keys including inserting, listing and updating keys
 pub fn make_key_management_route(
@@ -71,20 +71,10 @@ fn handle_post(
 ) -> HandlerFunction {
     Box::new(move |request, payload| {
         let key_store = key_store.clone();
-        let user_id = match request.match_info().get("user_id") {
-            Some(id) => id.to_owned(),
-            None => {
-                error!("User ID is not in path request");
-                return Box::new(
-                    HttpResponse::InternalServerError()
-                        .json(ErrorResponse::internal_error())
-                        .into_future(),
-                );
-            }
-        };
+        let validation = default_validation(&rest_config.issuer());
 
-        match authorize_user(&request, &user_id, &secret_manager, &rest_config) {
-            AuthorizationResult::Authorized => (),
+        let user_id = match authorize_user(&request, &secret_manager, &validation) {
+            AuthorizationResult::Authorized(claims) => claims.user_id(),
             AuthorizationResult::Unauthorized(msg) => {
                 return Box::new(
                     HttpResponse::Unauthorized()
@@ -99,7 +89,7 @@ fn handle_post(
                         .into_future(),
                 );
             }
-        }
+        };
 
         Box::new(into_bytes(payload).and_then(move |bytes| {
             let new_key = match serde_json::from_slice::<NewKey>(&bytes) {
@@ -153,20 +143,10 @@ fn handle_get(
 ) -> HandlerFunction {
     Box::new(move |request, _| {
         let key_store = key_store.clone();
-        let user_id = match request.match_info().get("user_id") {
-            Some(id) => id.to_owned(),
-            None => {
-                error!("User ID is not in path request");
-                return Box::new(
-                    HttpResponse::InternalServerError()
-                        .json(ErrorResponse::internal_error())
-                        .into_future(),
-                );
-            }
-        };
+        let validation = default_validation(&rest_config.issuer());
 
-        match authorize_user(&request, &user_id, &secret_manager, &rest_config) {
-            AuthorizationResult::Authorized => (),
+        let user_id = match authorize_user(&request, &secret_manager, &validation) {
+            AuthorizationResult::Authorized(claims) => claims.user_id(),
             AuthorizationResult::Unauthorized(msg) => {
                 return Box::new(
                     HttpResponse::Unauthorized()
@@ -181,7 +161,7 @@ fn handle_get(
                         .into_future(),
                 );
             }
-        }
+        };
 
         match key_store.list_keys(Some(&user_id)) {
             Ok(keys) => Box::new(
@@ -209,20 +189,10 @@ fn handle_patch(
 ) -> HandlerFunction {
     Box::new(move |request, payload| {
         let key_store = key_store.clone();
-        let user_id = match request.match_info().get("user_id") {
-            Some(id) => id.to_owned(),
-            None => {
-                error!("User ID is not in path request");
-                return Box::new(
-                    HttpResponse::InternalServerError()
-                        .json(ErrorResponse::internal_error())
-                        .into_future(),
-                );
-            }
-        };
+        let validation = default_validation(&rest_config.issuer());
 
-        match authorize_user(&request, &user_id, &secret_manager, &rest_config) {
-            AuthorizationResult::Authorized => (),
+        let user_id = match authorize_user(&request, &secret_manager, &validation) {
+            AuthorizationResult::Authorized(claims) => claims.user_id(),
             AuthorizationResult::Unauthorized(msg) => {
                 return Box::new(
                     HttpResponse::Unauthorized()
@@ -237,7 +207,7 @@ fn handle_patch(
                         .into_future(),
                 );
             }
-        }
+        };
 
         Box::new(into_bytes(payload).and_then(move |bytes| {
             let updated_key = match serde_json::from_slice::<UpdatedKey>(&bytes) {
@@ -310,19 +280,7 @@ fn handle_fetch(
 ) -> HandlerFunction {
     Box::new(move |request, _| {
         let key_store = key_store.clone();
-        let user_id = match request.match_info().get("user_id") {
-            Some(id) => id.to_owned(),
-            None => {
-                error!("User ID is not in path request");
-                return Box::new(
-                    HttpResponse::BadRequest()
-                        .json(ErrorResponse::bad_request(
-                            &"Failed to process request: no user id".to_string(),
-                        ))
-                        .into_future(),
-                );
-            }
-        };
+        let validation = default_validation(&rest_config.issuer());
 
         let public_key = match request.match_info().get("public_key") {
             Some(id) => id.to_owned(),
@@ -338,8 +296,8 @@ fn handle_fetch(
             }
         };
 
-        match authorize_user(&request, &user_id, &secret_manager, &rest_config) {
-            AuthorizationResult::Authorized => (),
+        let user_id = match authorize_user(&request, &secret_manager, &validation) {
+            AuthorizationResult::Authorized(claims) => claims.user_id(),
             AuthorizationResult::Unauthorized(msg) => {
                 return Box::new(
                     HttpResponse::Unauthorized()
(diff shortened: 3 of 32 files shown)


libsplinter/src/biome/rest_api/config.rs
```diff
@@ -29,7 +29,7 @@ pub struct BiomeRestConfig {
     /// Duration of JWT tokens issued by this service
     access_token_duration: Duration,
     #[cfg(feature = "biome-credentials")]
-    /// Cost for encripting users password
+    /// Cost for encrypting user's password
```
peterschwarz: Typo in commit message (also the summary is very long).

libsplinter/src/biome/rest_api/config.rs
```diff
@@ -28,6 +29,8 @@ pub struct BiomeRestConfig {
     issuer: String,
     /// Duration of JWT tokens issued by this service
     access_token_duration: Duration,
+
+    refresh_token_duration: Duration,
```
peterschwarz: For consistency with the other fields, please add a comment about this field.

libsplinter/src/biome/refresh_tokens/store/diesel/postgres.rs
```diff
@@ -0,0 +1,131 @@
...
+
+pub struct SplinterRefreshTokenStore {
+    connection: ConnectionPool,
+}
+
+impl RefreshTokenStore for SplinterRefreshTokenStore {
+    fn add_token(&self, user_id: &str, token: &str) -> Result<(), RefreshTokenError> {
+        let conn = self.connection.get()?;
+
+        splinter_user::table
```
peterschwarz: This should follow the operations pattern, like other biome stores.

splinterd/api/static/openapi.yml
```diff
@@ -897,6 +906,67 @@ paths:
...
+            application/json:
+                schema:
+                  $ref: '#/components/schemas/ErrorBiome'
+        401:
+          description: Access token is invalid
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorBiome'
+        403:
```
peterschwarz: Do more endpoints now return a 403 on invalid refresh tokens?

libsplinter/src/biome/rest_api/config.rs
RyanLassigBanks: The "typo" in the commit message was the typo being fixed in the commit. I've updated the message to make that clearer

splinterd/api/static/openapi.yml
RyanLassigBanks: The refresh token is only passed to the `/tokens` endpoint, so currently that's the only endpoint in biome that returns a 403.

libsplinter/src/biome/rest_api/config.rs
RyanLassigBanks: updated

libsplinter/src/biome/refresh_tokens/store/diesel/postgres.rs
RyanLassigBanks: I also have serious reservations about implementing a trait for a struct in a different file or module than the one that it's defined in. I think that should be avoided whenever possible, and we should follow this layout whenever possible.

```
struct Foo {}
impl Foo {}
impl Bar for Foo {}
impl Bizz for Foo {}
```

peterschwarz: The operations pattern allows us to target as broad a slice of diesel-supported databases as possible at the operation level.  For example, we can query both PostgreSQL and sqlite using a single implementation, but inserts may require different implementations, depending on the table structure or insert statements. 

For example, sqlite doesn't support default column values and PostgreSQL does.  In the `UserStore`, this will require two implementations of the `UserStoreAddUserOperation` for each type of connection. 

As for the second question, I'm not sure I understand what's raising the objection.  Can you elaborate with a more concrete example?

RyanLassigBanks: I issue is with how the operations object is defined [here](https://github.com/Cargill/splinter/blob/master/libsplinter/src/biome/credentials/store/diesel/operations/mod.rs), but its trait implementations are spread across the operations module. For examle,

https://github.com/Cargill/splinter/blob/03e9bfaaba4a4070ad5060628901566ac53b57f3/libsplinter/src/biome/credentials/store/diesel/operations/fetch_credential_by_username.rs#L28

Basically, I think those implementations should be in on module along side the operations definition.

vaporos: The operations pattern is necessary to deal with how Diesel works -- it's complex, but essentially boils down to being able to pick/choose which implementations of operations work for each backend database. For example, we may have 4 of 5 operations work for both, and 1 of 5 different between postgresql and sqlite.

vaporos: re:Organization of operations - follow the currently pattern, it has been carefully considered. If we change it at some point, we will do so across all of libsplinter at once. That discussion can happen outside this PR, and needs to occur after the sqlite implementation has been finished.

libsplinter/src/rest_api/sessions/mod.rs
```diff
@@ -78,7 +78,7 @@ where
     extra_validation(claims)
 }
 
-fn default_validation(issuer: &str) -> Validation {
+pub(crate) fn default_validation(issuer: &str) -> Validation {
```
vaporos: Should this be restricted to biome?

libsplinter/src/biome/rest_api/actix/key_management.rs
```diff
@@ -38,7 +38,7 @@ pub fn make_key_management_route(
     key_store: Arc<dyn KeyStore<Key>>,
     secret_manager: Arc<dyn SecretManager>,
 ) -> Resource {
-    Resource::build("/biome/users/{user_id}/keys")
+    Resource::build("/biome/users/keys")
```
vaporos: This change should really be in a separate PR, because it is going to require additional work to look at how we are doing routes to see if this is the only one we are going to change.

libsplinter/src/biome/datastore/diesel/models.rs
```diff
@@ -91,3 +91,21 @@ impl From<SplinterUser> for UserModel {
         UserModel { id: user.id() }
     }
 }
+
```
shannynalayna: Small typo in commit message: s/ grated / created ? Also, need to add a period at the end of the message.

libsplinter/src/biome/rest_api/config.rs
```diff
@@ -52,6 +59,7 @@ impl BiomeRestConfig {
 pub struct BiomeRestConfigBuilder {
     issuer: Option<String>,
     access_token_duration: Option<Duration>,
+    refresh_token_duration: Option<Duration>,
```
shannynalayna: Should this have a feature-guard, same with other instances of the `refresh_token_duration` within this file ?

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -125,8 +127,14 @@ impl RestResourceProvider for BiomeRestResourceManager {
                 ));
                 resources.push(make_login_route(
                     credentials_store.clone(),
+                    self.refresh_token_store.clone(),
+                    self.rest_config.clone(),
+                    Arc::new(AccessTokenIssuer::new(
+                        self.token_secret_manager.clone(),
+                        self.refresh_token_secret_manager.clone(),
```
shannynalayna: Also need to add a feature guard for this, which includes the biome-refresh-tokens

libsplinter/Cargo.toml
```diff
@@ -90,6 +90,7 @@ experimental = [
     "biome-credentials",
     "biome-key-management",
     "biome-notifications",
+    "biome-refresh-tokens",
```
shannynalayna: Should this be separated into a separate commit, since this feature guard appears to have been used before this commit?

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -115,6 +114,7 @@ impl RestResourceProvider for BiomeRestResourceManager {
 
         #[cfg(all(
             feature = "biome-credentials",
+            feature = "biome-refresh-tokens",
```
shannynalayna: Are the rest of the routes here guarded by the biome-refresh-tokens feature? It doesn't seem like they are since this includes the credentials list route and the users routes.

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -70,10 +70,9 @@ use crate::rest_api::secrets::{AutoSecretManager, SecretManager};
 pub use config::{BiomeRestConfig, BiomeRestConfigBuilder};
 pub use error::BiomeRestResourceManagerBuilderError;
 
-#[cfg(all(feature = "biome-credentials", feature = "rest-api-actix"))]
```
shannynalayna: Shouldn't be deleted, just add the biome-refresh-tokens guard.

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -78,6 +80,7 @@ pub use error::BiomeRestResourceManagerBuilderError;
 ))]
 use self::actix::{
     login::make_login_route,
+    register::make_register_route,
```
shannynalayna: It doesn't look like this should be added in this commit, but one of the previous commit. Also, it looks like the `make_register_route` is the only one that should be behind the `biome-refresh-tokens` feature. It's also added twice within this file.

libsplinter/src/biome/rest_api/actix/login.rs
```diff
@@ -15,6 +15,7 @@
 use std::sync::Arc;
```
shannynalayna: S / a auth / an auth in commit message

libsplinter/src/biome/rest_api/actix/login.rs
```diff
@@ -36,6 +37,7 @@ use super::super::resources::credentials::UsernamePassword;
 ///   }
 pub fn make_login_route(
     credentials_store: Arc<SplinterCredentialsStore>,
+    refresh_token_store: Arc<dyn RefreshTokenStore>,
```
shannynalayna: Either these lines with the refresh token store, or the login file itself, needs the 'biome-refresh-tokens` feature guard.

libsplinter/src/biome/rest_api/actix/mod.rs
```diff
@@ -23,5 +23,11 @@ pub(super) mod key_management;
 pub(super) mod login;
```
shannynalayna: Commit message: s/ a auth / an auth

libsplinter/src/biome/datastore/diesel/models.rs
RyanLassigBanks: updated

libsplinter/src/biome/rest_api/config.rs
RyanLassigBanks: Guarded by feature

libsplinter/Cargo.toml
RyanLassigBanks: Moved commit earlier in history, so this is introduced earlier

libsplinter/src/biome/rest_api/actix/login.rs
RyanLassigBanks: Fixed typo

libsplinter/src/biome/rest_api/actix/mod.rs
RyanLassigBanks: fixed type

libsplinter/src/biome/rest_api/mod.rs
RyanLassigBanks: I don't think adding the feature guard here is necessary since, the entire match statement is feature guarded

libsplinter/src/biome/rest_api/mod.rs
RyanLassigBanks: Re arranged commits so this gets added when the refresh token store gets added to the login route

libsplinter/src/biome/rest_api/mod.rs
RyanLassigBanks: rearranged commits so this makes sense

libsplinter/src/biome/rest_api/mod.rs
RyanLassigBanks: Fixed commits

libsplinter/src/biome/rest_api/actix/login.rs
RyanLassigBanks: Guarded where it's used it Biome Rest Resource provider

libsplinter/src/rest_api/sessions/mod.rs
RyanLassigBanks: We would have to move them into the biome crate, and I think the sessions stuff should stay together in the rest-api module.

libsplinter/src/biome/rest_api/actix/key_management.rs
RyanLassigBanks: Are you referring to the updates to route or the updates to `authorize_user` as a whole. Those changed were need to write the refresh token endpoint.

libsplinter/src/biome/rest_api/mod.rs
shannynalayna: Ok, didn't realize the refresh token secret manager isn't necessarily behind the refresh-tokens feature.

libsplinter/src/biome/rest_api/actix/token.rs
```diff
@@ -0,0 +1,179 @@
...
+
+                HttpResponse::Ok()
+                    .json(json!({ "token": token }))
+                    .into_future()
+            }))
+        })
+}
+
+#[derive(Deserialize)]
+pub struct RefreshToken {
```
shannynalayna: This should probably get moved into the resources module, under `resources/token` to follow the pattern that's currently used

libsplinter/src/biome/rest_api/actix/mod.rs
```diff
@@ -27,6 +27,12 @@ pub(super) mod key_management;
...
 #[cfg(feature = "biome-credentials")]
 pub(super) mod register;
+#[cfg(all(
```
vaporos: It should be fairly rare that things are guarded by multiple features; it implies that some of the items are cross-cutting concerns that can't be stated as dependencies. Here, I don't believe that is the case. This should probably only be guarded by biome-refresh-token with biome-refresh-token dependent on what it needs.

splinterd/Cargo.toml
```diff
@@ -66,9 +66,10 @@ experimental = [
     "service-arg-validation"
 ]
 
-biome = ["splinter/biome", "database"]
-biome-credentials = ["splinter/biome-credentials", "json-web-tokens", "biome"]
-biome-key-management = ["splinter/biome-key-management", "json-web-tokens", "biome"]
+biome = ["splinter/biome",  "database"]
```
vaporos: There is an additional space added here, looks like a typo.

splinterd/Cargo.toml
```diff
@@ -66,9 +66,10 @@ experimental = [
...
+biome-credentials = ["splinter/biome-credentials", "biome-refresh-tokens", "json-web-tokens", "biome"]
```
vaporos: It seems like biome-key-management and biome-credentials shouldn't depend on the biome-refresh-tokens feature, since there are no splinterd code changes here and the dependencies should be handled inside libsplinter itself.

vaporos: Maybe same with json-web-tokens?

libsplinter/src/biome/rest_api/actix/key_management.rs
vaporos: There are two things that could have been separate from this PR: a) the path change itself, which is a breaking change and completely unrelated to this PR; b) the fix which ignores {user_id} from the URL in favor of using the trusted user id.

I'm fine with it remaining in this PR at this point I guess.

libsplinter/src/biome/refresh_tokens/store/diesel/models.rs
```diff
@@ -0,0 +1,33 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::schema::refresh_tokens;
+
+#[cfg(feature = "biome-refresh-tokens")]
```
vaporos: This and the one below are unnecessary as it is already checked withing the biome:: module.

RyanLassigBanks: Updated

PR #579, opened 2020-03-17
Add /biome/verify route

shannynalayna:
Adds a /biome/verify REST endpoint to verify a user's password.
Also documents this endpoint in the openapi spec.

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/verify.rs
+++ b/libsplinter/src/biome/rest_api/actix/verify.rs
@@ -1,0 +1,126 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


shannynalayna:
@RyanLassigBanks My understanding is that it's necessary in the case of having some client resource that is password protected. This verify route is specific to just verifying the password, as the login route deals out a token for that user.

libsplinter/src/biome/rest_api/actix/verify.rs
```diff
@@ -0,0 +1,126 @@
...
+                    AuthorizationResult::Failed => {
+                        error!("Failed to authorize user");
+                            HttpResponse::InternalServerError()
+                                .json(ErrorResponse::internal_error())
+                                .into_future()
+                    }
+                }
+            }))
+        })
+}
```
ltseeley: Please add some tests if possible; see the the circuit/proposal read tests for examples of how to test endpoints: https://github.com/Cargill/splinter/blob/master/libsplinter/src/admin/rest_api/actix/circuits.rs

shannynalayna:
Closing this story based on the discussion that this functionality is redundant and not necessary for what is currently implemented.

libsplinter/src/biome/rest_api/actix/verify.rs
shannynalayna: As it stands, there is a lot of background work that needs to go into setting up the unit tests. It involved editing the existing *-stores and how they're passed around to make the biome resources. There is a story to add in these tests: https://zenhub.cglcloud.com/app/workspaces/product-core-5d37298b0add7306b30efde2/issues/digitalsupplychain/product-core/2019

PR #581, opened 2020-03-18
Biome user unit tests

RyanLassigBanks:
Unit tests for the biome users module


RyanLassigBanks:
After a discussion with @vaporos we concluded these tests don't offer enough value closing

PR #582, opened 2020-03-18
Add inbound connections to ConnectionManager

peterschwarz:
This PR adds the ability for the `ConnectionManager` to receive inbound connections and notify listeners when a connection is added.  

The internals of `ConnectionManager` now differentiate between outbound and inbound connections, where outbound connections are eligible for re-connection and heartbeats.  Inbound connections are not eligible - this is left to the remote end to manage these activities.

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -183,6 +183,14 @@ pub struct Connector {
 }
 
 impl Connector {
+    /// Request a connection to the given endpoint.
+    ///
+    /// This operation is idempotent: if a connection to that endpoint already exists, a new
+    /// connection is not created.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if the connection cannot be created
     pub fn request_connection(
         &self,
         endpoint: &str,
@@ -208,15 +216,15 @@ impl Connector {
         })?
     }
 
-    // Removes a connection
-    //
-    // # Returns
-    //
-    // The endpoint, if the connection exists; None, otherwise.
-    //
-    // # Errors
-    //
-    // Returns a ConnectionManagerError if the query cannot be performed.
+    /// Removes a connection
+    ///
+    ///  # Returns
+    ///
+    ///  The endpoint, if the connection exists; None, otherwise.
+    ///
+    ///  # Errors
+    ///
+    ///  Returns a ConnectionManagerError if the query cannot be performed.
     pub fn remove_connection(
         &self,
         endpoint: &str,
@@ -240,6 +248,16 @@ impl Connector {
         })?
     }
 
+    /// Subscribe to notfications for connection events
+    ///
+    /// # Returns
+    ///
+    /// Notifications are received via an iterator.  The iterator will block when using standard
+    /// `next`, but also provides a `try_next`
+    ///
+    /// # Errors
+    ///
+    /// Return a ConnectionManagerError if the notification iterator cannot be created.
     pub fn subscribe(&self) -> Result<NotificationIter, ConnectionManagerError> {
         let (send, recv) = channel();
         match self.sender.send(CmMessage::Subscribe(send)) {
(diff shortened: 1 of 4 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -55,6 +55,10 @@ enum CmRequest {
     ListConnections {
         sender: Sender<Result<Vec<String>, ConnectionManagerError>>,
     },
+    AddInboundConnection {
```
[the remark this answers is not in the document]
peterschwarz: I've kept the name, since it is still requesting a connection via an endpoint string (request makes sense as it could fail in ways that add_inbound connection cannot).  I have renamed the message variant for to `RequestOutboundConnection`

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -406,25 +435,38 @@ where
             ));
         };
 
+        if !meta.is_outbound() {
```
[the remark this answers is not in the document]
peterschwarz: For the usage, and the order it was added, outbound was the more interesting piece of information.  I.e. this next block of code only applies to outbound connections - if we end up adding other metadata states, we don't need to change this code.

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -711,6 +711,9 @@ fn create_heartbeat() -> Result<Vec<u8>, ConnectionManagerError> {
 #[cfg(test)]
 pub mod tests {
     use super::*;
+
```
agunde406: Typo in commit message "can be remove by" -> "can be removed by"

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -308,8 +308,35 @@ impl ShutdownHandle {
 }
 
 #[derive(Clone, Debug)]
-struct ConnectionMetadata {
```
agunde406: The commit message should be updated to not include mention of heartbeats anymore since, both inbound and outbound connections will receive heartbeats

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -604,6 +668,23 @@ fn send_heartbeats<T: MatrixLifeCycle, U: MatrixSender>(
...
+            // We leave it up to the remote to peform heartbeats
+            ConnectionMetadata::Inbound { id, .. } => {
+                info!("Sending heartbeat to {}", endpoint);
+                if let Err(err) = state.matrix_sender().send(id, heartbeat_message.clone()) {
+                    error!(
+                        "failed to send heartbeat: {:?} attempting reconnection",
+                        err
+                    );
+
+                    notify_subscribers(
```
agunde406: So is the idea here that the PeerManager will handle inbound peers differently when it comes to trying different endpoints? The PeerManager currently checks the number of re-connection attempts, and after a specified # of failures tries different endpoints. But these notifications are not sent for inbound connections. 

Also right now this will send a disconnected notification every time a heartbeat fails, (every 10 seconds by default).

peterschwarz: I've updated the inbound metadata to include a disconnected state - this allows for:
- a heartbeat to fail intermittently
- the layer is notified when the connection is initially detected as disconnected - it's still up to the listening layer to decide what to do at that point.

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: Fixed

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: Fixed (the typo was in the test comment too)

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -604,6 +668,23 @@ fn send_heartbeats<T: MatrixLifeCycle, U: MatrixSender>(
                     }
                 }
             }
+            // We leave it up to the remote to peform heartbeats
```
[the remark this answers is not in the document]
peterschwarz: Removed

PR #583, opened 2020-03-19
Mutable user store and Arc Removal

RyanLassigBanks:
PR that makes `add_user`, `remove_user`, and `update_user` take a mutable reference to self and remove `Arc` wrappings of `SplinterUserStore`

### DIFF
--- a/libsplinter/src/biome/user/store/diesel/mod.rs
+++ b/libsplinter/src/biome/user/store/diesel/mod.rs
@@ -42,15 +42,15 @@ impl SplinterUserStore {
 }
 
 impl UserStore<SplinterUser> for SplinterUserStore {
-    fn add_user(&self, user: SplinterUser) -> Result<(), UserStoreError> {
+    fn add_user(&mut self, user: SplinterUser) -> Result<(), UserStoreError> {
         UserStoreOperations::new(&*self.connection_pool.get()?).add_user(user.into())
     }
 
-    fn update_user(&self, updated_user: SplinterUser) -> Result<(), UserStoreError> {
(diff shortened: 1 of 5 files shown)


libsplinter/src/biome/user/store/diesel/mod.rs
```diff
@@ -42,15 +42,15 @@ impl SplinterUserStore {
 }
 
 impl UserStore<SplinterUser> for SplinterUserStore {
-    fn add_user(&self, user: SplinterUser) -> Result<(), UserStoreError> {
+    fn add_user(&mut self, user: SplinterUser) -> Result<(), UserStoreError> {
```
peterschwarz: I'm not sure that the commit message makes sense in the context of this PR (it makes more sense in the context of the original PR).

libsplinter/src/biome/user/store/mod.rs
```diff
@@ -56,7 +56,7 @@ impl From<UserModel> for SplinterUser {
 
 /// Defines methods for CRUD operations and fetching and listing users
 /// without defining a storage strategy
-pub trait UserStore<T> {
+pub trait UserStore<T>: Send + Sync + Clone {
```
peterschwarz: Using Clone this way, we can't create `Box<dyn UserStore>` (the compiler will complain).  We probably don't need this constraint, yet until we start using `UserStore` as a trait.  `NodeRegistry` is a good example of how we have to handle it.

RyanLassigBanks: Added CloneBoxUserStore trait

PR #584, opened 2020-03-19
Separate circuit/proposal REST API responses from internal structs

ltseeley:
Updates the serializable structs used by the `circuit-read` and `proposal-read` REST API endpoints to be distinct from internal splinter structs, such that they are defined and controlled by the REST API layer itself. These structs are built from the internal ones using `From`/`Into` implementations.

Also removes fields from these REST API responses that are not currently used by splinter (such as `auth` and `persistence`).


splinterd/api/static/openapi.yml
```diff
@@ -1505,53 +1505,39 @@ components:
                       type: array
                       items:
                         type: integer
+
     Circuit:
       type: object
       properties:
-        circuit_id:
```
ltseeley: I'll be putting up a PR against the splinter docs repo as well to reflect these changes

libsplinter/src/admin/rest_api/resources/circuits.rs
```diff
@@ -12,23 +12,62 @@
...
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-pub(crate) struct ListCircuitsResponse {
-    pub data: Vec<CircuitResponse>,
-    pub paging: Paging,
+pub(crate) struct ServiceResponse {
+    pub service_id: String,
+    pub service_type: String,
+    pub allowed_nodes: Vec<String>,
+    pub arguments: BTreeMap<String, String>,
```
ltseeley: Curious what peoples' thoughts are about whether the `arguments` field should be included in this response.

agunde406: I personally think it should be included.

libsplinter/src/admin/rest_api/resources/circuits.rs
```diff
@@ -12,23 +12,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::circuit::{AuthorizationType, DurabilityType, PersistenceType, Roster, RouteType};
+use std::collections::BTreeMap;
+
+use crate::circuit::{Circuit, Roster, ServiceDefinition};
 use crate::rest_api::paging::Paging;
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
peterschwarz: This should be Serialize-only.  The response structs should be updated accordingly.

libsplinter/src/admin/rest_api/resources/circuits.rs
```diff
@@ -12,23 +12,62 @@
...
+pub(crate) struct ListCircuitsResponse {
```
peterschwarz: Likewise, it doesn't need to own its data

libsplinter/src/admin/rest_api/resources/circuits.rs
ltseeley: Removed the `Deserialize` derivation for all responses

libsplinter/src/admin/rest_api/resources/circuits.rs
ltseeley: Updated to borrow when possible

PR #586, opened 2020-03-19
Add a BiHashMap to Mesh for a unique id to mesh id

agunde406:
This change enables components using mesh to not have
to know about the internal mesh id used in polling the
connections. This also allows reconnection with out having
to replace the id returned from mesh.

This commit updates mesh and all places mesh is used. It
also removes the use of rwlock unwrap macros and instead
returns a PoisonedLock error instead.

This will simplify the data future inteconnects will need
to keep track of.

### DIFF
--- a/examples/private_counter/service/src/main.rs
+++ b/examples/private_counter/service/src/main.rs
@@ -360,6 +360,10 @@ fn run_service_loop(
                 error!("Network has disconnected");
                 break;
             }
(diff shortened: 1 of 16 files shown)


peterschwarz:
There's a lot of calls to `to_string` added, but I think that is a consequence of our BiHapMap implementation.  We probably can clean that up at some point in the future, and remove a large number of allocations.

PR #589, opened 2020-03-20
Allow filtering circuit proposals by member

ltseeley:
Adds the `member` filter as a query parameter to the `GET /admin/proposals` endpoint and as an argument to the `splinter circuit proposals` CLI command. Also renames to the existing `filter` option to `management_type` to differentiate.

### DIFF
--- a/cli/src/action/circuit/api.rs
+++ b/cli/src/action/circuit/api.rs
@@ -155,10 +155,19 @@ impl<'a> SplinterRestClient<'a> {
     pub fn list_proposals(
         &self,
         management_type_filter: Option<&str>,
+        member_filter: Option<&str>,
     ) -> Result<ProposalListSlice, CliError> {
-        let mut request = format!("{}/admin/proposals", self.url);
+        let mut filters = vec![];
         if let Some(management_type) = management_type_filter {
-            request = format!("{}?management_type={}", &request, &management_type);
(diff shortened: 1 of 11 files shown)


libsplinter/src/admin/rest_api/actix/proposals.rs
```diff
@@ -85,17 +85,26 @@ fn list_proposals<PS: ProposalStore + 'static>(
         None => DEFAULT_LIMIT,
     };
 
-    let mut link = req.uri().path().to_string();
```
agunde406: Please update commit message to say `management_type` instead of `mgmt_type`

cli/src/action/circuit/api.rs
```diff
@@ -152,10 +152,13 @@ impl<'a> SplinterRestClient<'a> {
             })
     }
 
-    pub fn list_proposals(&self, filter: Option<&str>) -> Result<ProposalListSlice, CliError> {
+    pub fn list_proposals(
```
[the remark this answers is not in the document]
ltseeley: Fixed

libsplinter/src/admin/rest_api/actix/proposals.rs
ltseeley: Fixed

PR #592, opened 2020-03-20
Rename datastore to migrations in biome

RyanLassigBanks:


libsplinter/src/biome/migrations/mod.rs
```diff
@@ -12,6 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Provides sql migration scripts and methods for executing
```
ltseeley: This commit says "for biome", but it should say "for biome migrations"

RyanLassigBanks: Updated commit message

cli/src/action/database.rs
```diff
@@ -18,7 +18,7 @@ use super::Action;
 use crate::error::CliError;
 use diesel::{connection::Connection as _, pg::PgConnection};
 #[cfg(feature = "database-migrate-biome")]
-use splinter::biome::datastore::run_postgres_migrations;
```
agunde406: Typo in commit message "databasestore -> datastore"

RyanLassigBanks: Updated

libsplinter/src/biome/migrations/mod.rs
```diff
@@ -12,6 +12,21 @@
...
+//! migrations.
+//!
+//! ```ignore
+//!
+//! use biome::migrations::run_postgres_migratoins;
```
agunde406: typo "run_postgres_migratoins;" -> "run_postgres_migrations;"

PR #594, opened 2020-03-20
Make user module private in biome

RyanLassigBanks:
Make crate::biome::user non public


libsplinter/src/biome/user/store/mod.rs
```diff
@@ -33,6 +33,7 @@ impl SplinterUser {
     ///
     /// * `user_id`: unique identifier for the user being created
     ///
+    #[allow(dead_code)]
```
peterschwarz: This implies that something is not correctly guarded by a feature.

RyanLassigBanks: Changed to rest-api guard

PR #598, opened 2020-03-23
Update PUT /biome/users/{user_id} route

shannynalayna:
Adds functionality to the existing PUT /biome/users/{user_id} endpoint to update the keys object, as well as the user's password. Also changes the formats to the /biome/users/{user_id} routes so that the closure is handled in the handler function, rather than the function which makes the user routes.

### DIFF
--- a/libsplinter/src/biome/credentials/store/diesel/mod.rs
+++ b/libsplinter/src/biome/credentials/store/diesel/mod.rs
@@ -14,7 +14,7 @@
 
 pub(in crate::biome) mod models;
 mod operations;
-mod schema;
+pub(in crate::biome) mod schema;
 
 use super::{CredentialsStore, CredentialsStoreError, UserCredentials, UsernameId};
 use crate::database::ConnectionPool;
--- a/libsplinter/src/biome/rest_api/actix/user.rs
+++ b/libsplinter/src/biome/rest_api/actix/user.rs
@@ -14,7 +14,6 @@
 
 use std::sync::Arc;
 
-use super::authorize::authorize_user;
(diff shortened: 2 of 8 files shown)


libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -112,21 +126,27 @@ fn add_fetch_user_method(credentials_store: Arc<SplinterCredentialsStore>) -> Ha
...
 /// Defines a REST endpoint to edit a user's credentials in the database
 ///
 /// The payload should be in the JSON format:
 ///   {
 ///       "username": <existing username of the user>
 ///       "hashed_password": <hash of the user's existing password>
-///       "new_password": OPTIONAL <hash of the user's updated password>
+///       "new_password": <hash of the user's updated password>
+///       "display_name": <display name for key>
+///       "public_key": <public key of the user>
```
dplumb94: Users can be associated with multiple keys, so this needs to be a list of updated key pairs for all keys associated with the user.

shannynalayna: Addressed to take in a list of new key information, as well as return the successfully added keys.

libsplinter/src/biome/credentials/store/diesel/mod.rs
```diff
@@ -14,7 +14,7 @@
 
 pub(in crate::biome) mod models;
```
RyanLassigBanks, line 15: Could you add something to the commit message explaining that this so it can be used `update_keys_and_password`, or squash it into that commit?

PR #600, opened 2020-03-23
 Allow specifying URL for CLI with env variable

ltseeley:
Splinter CLI commands that take the URL of a splinter daemon as an
argument now check the `SPLINTERD_URL` environment variable if the
`-U`/`--url` argument is not specified.

### DIFF
--- a/cli/src/action/mod.rs
+++ b/cli/src/action/mod.rs
@@ -36,6 +36,8 @@ use super::error::CliError;
 
 #[cfg(any(feature = "health", feature = "circuit"))]
 const DEFAULT_ENDPOINT: &str = "http://127.0.0.1:8085";
+#[cfg(any(feature = "health", feature = "circuit"))]
+const SPLINTER_REST_API_URL_ENV: &str = "SPLINTER_REST_API_URL";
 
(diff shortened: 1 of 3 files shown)


cli/src/action/mod.rs
```diff
@@ -36,6 +36,8 @@ use super::error::CliError;
 
 #[cfg(any(feature = "health", feature = "circuit"))]
 const DEFAULT_ENDPOINT: &str = "http://127.0.0.1:8085";
+#[cfg(any(feature = "health", feature = "circuit"))]
+const ENDPOINT_ENV: &str = "SPLINTERD_URL";
```
vaporos: Let's start this with SPLIINTER_ as a prefix (for all splinter env variables), so something like SPLINTER_DAEMON_URL,  SPLINTER_URL, etc.

vaporos: other options: SPLINTER_REST_API, SPLINTER_API_URL, SPLINTER_REST_URL

ltseeley: Updated to SPLINTER_REST_API

PR #601, opened 2020-03-24
Generalize Connection Manager subscriptions

peterschwarz:
Generalize the `ConnectionManager` subscriptions to use a callback model. This allows for a reduction in threads required in the peer manager (and other users), with respect to notifications.  Previously, a thread was needed to read the notifications from the `NotificationIter`, only to take that notification and pass it on to another sender to the `PeerManager`'s background thread. With this change, the notification is immediately placed on the sender to the background thread.

This generalization restructures the subscribers into a `SubscriberMap`, where each subscriber has an `SubscriberId`.  This id can be used to unsubscribe from notifications if desired.

The original `subscribe` method has been renamed to `subscription_iter`.


peterschwarz:
Closing do to the desire to not allow any long running callbacks enter into the ConnectionManager's background thread.

PR #603, opened 2020-03-25
Add node-file arg to splinter circuit create cmd

ltseeley:
Add the `--node-file` argument to the `splinter circuit create` command.
This argument, which can be used instead of or in conjunction with the
`--node` argument, allows the user to load a list of nodes from a .yaml
file either on the local file system or from a remote server with HTTP.
This argument supports the existing node file types used in splinter,
namely those used by the node registry and the `splinter node alias`
commands.

To test:
1. Add `- ./node_registry:/node_registry` to the `generate-key-registry` container's volumes in `examples/gameroom/docker-compose.yaml`.
2. Run gameroom with `CARGO_ARGS="-- --features=experimental" docker-compose -f examples/gameroom/docker-compose.yaml up --build`
3. Start a session in the `generate-key-registry` container with `docker-compose -f examples/gameroom/docker-compose.yaml run generate-key-registry bash`
4. Create a circuit using a node registry file:
```
splinter circuit create \
  --key /key_registry/alice.priv \
  --url http://splinterd-node-acme:8085  \
  --node-file /node_registry/nodes.yaml \
  --service sc00::acme-node-000 \
  --service sc01::bubba-node-000 \
  --service-type *::scabbard \
  --management custom \
  --service-arg *::admin_keys=$(cat /key_registry/alice.pub) \
  --service-peer-group sc00,sc01
```
5. Create a circuit using a node alias file:
```
splinter node alias add acme acme-node-000 tls://splinterd-node-acme:8044 && \
splinter node alias add bubba bubba-node-000 tls://splinterd-node-bubba:8044 && \
splinter circuit create \
  --key /key_registry/alice.priv \
  --url http://splinterd-node-acme:8085  \
  --node-file ~/.splinter/node_alias.yaml \
  --service sc00::acme-node-000 \
  --service sc01::bubba-node-000 \
  --service-type *::scabbard \
  --management custom \
  --service-arg *::admin_keys=$(cat /key_registry/alice.pub) \
  --service-peer-group sc00,sc01
```
6. Create a circuit using a remote .yaml file:
```
apt-get install -y -q ca-certificates && \
splinter circuit create \
  --key /key_registry/alice.priv \
  --url http://splinterd-node-acme:8085  \
  --node-file http://raw.githubusercontent.com/Cargill/splinter/master/examples/gameroom/node_registry/nodes.yaml \
  --service sc00::acme-node-000 \
  --service sc01::bubba-node-000 \
  --service-type *::scabbard \
  --management custom \
  --service-arg *::admin_keys=$(cat /key_registry/alice.pub) \
  --service-peer-group sc00,sc01
```

### DIFF
--- a/cli/src/action/circuit/mod.rs
+++ b/cli/src/action/circuit/mod.rs
@@ -25,6 +25,7 @@ use std::fs::File;
 use std::io::Read;
 
 use clap::ArgMatches;
+use serde::Deserialize;
 use splinter::admin::messages::{CreateCircuit, SplinterService};
 
 use crate::error::CliError;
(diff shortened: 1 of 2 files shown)


cli/src/action/circuit/mod.rs
```diff
@@ -204,6 +214,49 @@ impl Action for CircuitCreateAction {
...
+struct Node {
+    #[serde(alias = "node_id")]
+    identity: String,
+    endpoint: String,
+}
+
+fn load_nodes_from_file(node_file: &str) -> Result<Vec<Node>, CliError> {
+    if node_file.starts_with("http://") || node_file.starts_with("https://") {
+        load_nodes_from_remote(node_file)
+    } else {
```
vaporos: Because we support http://, users may also expect that they can use file://, so it would be good to support that here (just stripping it off if it is given).

cli/src/main.rs
```diff
@@ -237,15 +237,30 @@ fn run() -> Result<(), CliError> {
                     .takes_value(true)
                     .help("Path to private key file"),
             )
+            .arg(
+                Arg::with_name("node_file")
+                    .long("node-file")
+                    .takes_value(true)
+                    .required_unless("node")
+                    .long_help(
+                        "File system path or URL to a .yaml file of node definitions. \
```
vaporos: This description is too long. Limit it to the first sentence. Remove the period at the end (these should all be the same). The rest could go in a details section below (there is an example in the generate cert command I believe).

Change .yaml to YAML or something like "Filesystem path or HTTP URL to nodes file", because the format needs more explanation than just "YAML file" and we can't do that here. (Probably a man page thing.)

cli/src/main.rs
```diff
@@ -237,15 +237,30 @@ fn run() -> Result<(), CliError> {
...
+                         Local files can be loaded from any valid path. \
+                         URLs for remote files must be prefixed with 'http://' or 'https://'. \
+                         A valid .yaml file will be a list nodes, where each node has an \
```
vaporos: If we do keep this description, in a details section, change it to be more example-based.

PR #606, opened 2020-03-26
Add a method to PeerManagerConnector to get connection ids

agunde406:
Adds a method to PeerManagerConnector to get a BiHashMap for
peer ids to connection ids. This will be used by PeerInterconnect
to get the peers and connection id that can be used for sending
messages.

To support this method, the PeerManager will handle a new message
PeerManagerRequest::ConnectionIds and the PeerMap has a new method
to return the BiHashMap.

This returns a BiHashMap because the PeerInterconnect will need to
go convert the ids to each other, where the direction depends on
if the message is incoming or outgoing.

### DIFF
--- a/libsplinter/src/network/peer_manager/peer_map.rs
+++ b/libsplinter/src/network/peer_manager/peer_map.rs
@@ -14,6 +14,8 @@
 
 use std::collections::HashMap;
 
+use crate::collections::BiHashMap;
+
 use super::error::PeerUpdateError;
 
 #[derive(Clone, PartialEq, Debug)]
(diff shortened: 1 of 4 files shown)


libsplinter/src/network/peer_manager/peer_map.rs
```diff
@@ -213,6 +227,41 @@ pub mod tests {
         assert_eq!(peers, vec!["new_peer".to_string(), "next_peer".to_string()]);
     }
 
+    // Test that connection_ids() returnscorrectly
```
ltseeley: Typo "returnscorrectly"

agunde406: fixed

PR #607, opened 2020-03-26
 Display new circuit for splinter circuit create

ltseeley:
Updates the `splinter circuit create` command to display the newly
proposed circuit after it is submitted. This allows the user to
immediately see the circuit that was generated without having to run
`splinter circuit list` or `splinter circuit show`.

Also reorganizes the code slightly and removes the extra newline from
the dry-run output.

### DIFF
--- a/cli/src/action/circuit/mod.rs
+++ b/cli/src/action/circuit/mod.rs
@@ -357,7 +357,7 @@ fn parse_service_type_argument(service_type: &str) -> Result<(String, String), C
 }
 
 /// Reads a private key from the given file name.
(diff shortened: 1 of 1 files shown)


cli/src/action/circuit/mod.rs
```diff
@@ -25,13 +25,15 @@ use std::fs::File;
 use std::io::Read;
 
 use clap::ArgMatches;
+use splinter::admin::messages::{CreateCircuit, SplinterService};
```
agunde406: The commit message should say without having to run`splinter circuit proposals` not list

ltseeley: Fixed

PR #609, opened 2020-03-26
Rename `circuit create` to `circuit propose`

ltseeley:
Renames the `circuit create` command to `circuit propose` to make the
command more clear about what it does; it doesn't create a circuit, it
proposes a circuit that needs to be voted on before being created.


vaporos:
Before merging this, please create an associated PR against the other repos which will be impacted (but don't merge them until after this one), including a PR to add information to the appropriate upgrading file in the splinter-docs repo.

PR #611, opened 2020-03-26
Use From conversion for transport IoError errors

vaporos:
The From<io::Error> impl for SendError and RecvError translates
io::ErrorKind::UnexpectedEof to Send/RecvError::Disconnected and
io::ErrorKine::WouldBlock to Send/RecvError::WouldBlock.

However, socket::tcp and socket::tls were not using from but instead
directly creating the errors via Send/RecvError::IoError(err).
Potentially, there are some bugs created because Mesh uses
Send/RecvError::Disconnected/WouldBlock in its logic.

Universally changed all transport code involving IoError to use ::from.
Although other errors don't currenlty differentiate (and thus it isn't
a functional changer for them), it seems best to do it the same way
throughout.

### DIFF
--- a/libsplinter/src/transport/socket/tcp.rs
+++ b/libsplinter/src/transport/socket/tcp.rs
@@ -55,7 +55,7 @@ impl Transport for TcpTransport {
                 FrameError::UnsupportedVersion => ConnectError::ProtocolError(
                     "Unable to connect; remote version is not with in range".into(),
                 ),
-                FrameError::IoError(err) => ConnectError::IoError(err),
+                FrameError::IoError(err) => ConnectError::from(err),
                 e => ConnectError::ProtocolError(format!("Unexpected protocol error: {}", e)),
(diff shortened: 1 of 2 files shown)


libsplinter/src/transport/socket/tcp.rs
```diff
@@ -55,7 +55,7 @@ impl Transport for TcpTransport {
                 FrameError::UnsupportedVersion => ConnectError::ProtocolError(
                     "Unable to connect; remote version is not with in range".into(),
                 ),
-                FrameError::IoError(err) => ConnectError::IoError(err),
+                FrameError::IoError(err) => ConnectError::from(err),
```
ltseeley: Typo in commit message: s/ErrorKine/ErrorKind/

vaporos: fixed

PR #613, opened 2020-03-27
Transport error module

vaporos:
Move errors in transport:: to transport::errors and implement Error and Display for all errors.

### DIFF
--- a/libsplinter/src/transport/error.rs
+++ b/libsplinter/src/transport/error.rs
@@ -1,0 +1,119 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error::Error;
+use std::io;
+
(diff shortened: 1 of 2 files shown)


libsplinter/src/transport/error.rs
```diff
@@ -21,6 +21,24 @@ pub enum AcceptError {
     ProtocolError(String),
 }
 
+impl Error for AcceptError {
```
agunde406: Should "ghe above " -> Say "the above"?

agunde406: In the commit message

ltseeley: Also s/is common/is a common/

vaporos: fixed

libsplinter/src/transport/mod.rs
```diff
@@ -22,10 +23,12 @@ pub mod tls;
 #[cfg(feature = "zmq-transport")]
 pub mod zmq;
 
-use mio::Evented;
+pub use error::{
+    AcceptError, ConnectError, DisconnectError, ListenError, PollError, RecvError, SendError,
+    StatusError,
+};
```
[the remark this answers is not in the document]
vaporos: fixed

libsplinter/src/transport/mod.rs
```diff
@@ -24,13 +25,14 @@ pub mod zmq;
...
-use std::io;
+pub use error::{
+    AcceptError, ConnectError, DisconnectError, ListenError, PollError, RecvError, SendError,
+    StatusError,
+};
 
 pub enum Status {
     Connected,
     Disconnected,
-}
```
peterschwarz: I can't tell if this is a github problem or an accidental deletion.

peterschwarz: From the looks of it, the rest of the enum is deleted in a later commit

PR #615, opened 2020-03-27
Implement NetworkMessageSender and NetworkMessagSendQueue

agunde406:
Instead of having a Sender<SendRequest>, dispatch handlers and
other components that need to send a message over the network
will get a NetworkMessageSender. The sender will hide the
implementation details about sending a message to the network
object. As a result SendRequest is now private.

This change is also in support of removing atomic bools used in
shutdown. The NeworkMessageSendQueue now provides a ShutdownSignaler
that will notify the sending thread to shutdown.

This change required that all handler test were rewritten to use
networks and inproc connections. The authorization handlers required
tcp connections because inproc connections are automatically authorized.

Also fixes private_counter and privater_xo examples

### DIFF
--- a/libsplinter/src/network/sender.rs
+++ b/libsplinter/src/network/sender.rs
@@ -11,93 +11,121 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
-use std::time::Duration;
+use std::sync::mpsc::{channel, Sender};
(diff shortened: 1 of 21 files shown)


libsplinter/src/network/sender.rs
```diff
@@ -120,21 +144,20 @@ mod tests {
     use crate::transport::socket::TcpTransport;
     use crate::transport::Transport;
 
-    // Test that a message can successfully be sent by passing it to the sender end of the
```
peterschwarz, line 123: Could you update the comments instead of deleting them?

peterschwarz:
Otherwise, tested locally both with gameroom and `cargo test`

PR #618, opened 2020-03-30
Stabilize circuit-read

ltseeley:
Stabilizes the `circuit-read` feature by removing the feature from the
libsplinter and splinterd Cargo.toml files and removing the feature
guards in the code.


ltseeley:
The commit that fixed the build was removed from this PR because it was merged in #643

PR #619, opened 2020-03-30
Remove arcs in rest api

RyanLassigBanks:
This PR refactors `BiomeRestResourceManager` to take generic types for all of its stores and its SecretManager. As part of this process we are also removing all the uses of `Arc` by BiomeRestResourceManager. Most of it's uses were to allow the stores to be cloned, as such we are having the Clone be a trait bound for stores being used by BiomeRestResourceManager.


libsplinter/src/biome/rest_api/actix/authorize.rs
```diff
@@ -23,9 +21,9 @@ use crate::rest_api::secrets::SecretManager;
 use crate::rest_api::sessions::Claims;
 
 /// Verifies the user has the correct permissions
-pub(crate) fn authorize_user(
+pub(crate) fn authorize_user<SM: SecretManager>(
```
vaporos: The commit message has typo and grammar issues.

PR #622, opened 2020-03-30
Move database-migrate-biome feature into stable

RyanLassigBanks:


cli/Cargo.toml
```diff
@@ -68,11 +70,7 @@ experimental = [
 circuit = ["reqwest", "serde_json", "splinter/sawtooth-signing-compat", "dirs"]
 circuit-auth-type = ["circuit"]
 circuit-template = ["splinter/circuit-template", "circuit"]
-database-migrate-biome = [
-    "splinter/biome-credentials",
-    "splinter/biome-key-management",
```
agunde406: This change should be in a separate commit from the stabilization commit, with some explanation in the commit message about this change.

cli/Cargo.toml
```diff
@@ -69,12 +69,8 @@ experimental = [
...
-    "splinter/biome-notifications"
-]
-
+database-migrate-biome = ["splinter/biome"]
+node-alias = ["dirs", "reqwest"]
```
peterschwarz, line 73: Looks like an extra addition

PR #625, opened 2020-03-31
Stabilize proposal-read

ltseeley:
Stabilizes the `proposal-read` feature by removing the feature from the
libsplinter and splinterd Cargo.toml files and removing the feature
guards in the code.


ltseeley:
Rebased on master to pick up build fix from #643

PR #627, opened 2020-03-31
Credentials store cleanup

vaporos:
Removes generics from UserCredentialsStore and renames SplinterUserCredentialsStore to DieselUserCredentialsStore.


vaporos:
> The update_credentials doc comment no longer makes sense.
> 
> ```
> /// Replaces a credential of a certain type for a user in the underlying storage with new
> /// credentials. This assumes that the user has only one credential of a certain type in
> /// storage
> ```

Fixed

PR #628, opened 2020-03-31
Move database feature to stable

RyanLassigBanks:
So we can stabilize the biome features


vaporos:
This will not be required if #635 is accepted.

RyanLassigBanks:
Closing in favor of [#635](http://github.com/Cargill/splinter/pull/635)

PR #630, opened 2020-03-31
Reject dup node IDs/endpoints in circuit propose

ltseeley:
Updates the `CreateCircuitMessageBuilder::add_node` method used by the
`splinter circuit propose` CLI command to check for duplicate node IDs
and endpoints. If a duplicate is detected, an error is returned to the
user.

### DIFF
--- a/cli/src/action/circuit/builder.rs
+++ b/cli/src/action/circuit/builder.rs
@@ -182,6 +182,22 @@ impl CreateCircuitMessageBuilder {
 
     pub fn add_node(&mut self, node_id: &str, node_endpoint: &str) -> Result<(), CliError> {
         let node = make_splinter_node(node_id, node_endpoint)?;
+
+        for node in &self.nodes {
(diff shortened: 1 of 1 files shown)


cli/src/action/circuit/builder.rs
```diff
@@ -177,13 +177,25 @@ impl CreateCircuitMessageBuilder {
...
+                "Duplicate node ID detected: {}",
+                node_id
+            )));
+        }
+        if self.nodes.iter().any(|node| node.endpoint == node_endpoint) {
+            return Err(CliError::ActionError(format!(
+                "Duplicate node endpoint detected: {}",
+                node_endpoint
+            )));
+        }
```
eloaverona: It's not a huge difference, buy you could probably refactor this to require a single iteration over nodes?

ltseeley: Updated

PR #632, opened 2020-03-31
Credentials fixes

RyanLassigBanks:
Miscellaneous fixes to address comments regarding credential feature stabilization.

### DIFF
--- a/libsplinter/src/biome/credentials/store/mod.rs
+++ b/libsplinter/src/biome/credentials/store/mod.rs
@@ -138,6 +138,11 @@ pub trait CredentialsStore<T> {
     /// # Arguments
     ///
     ///  * `credentials` - Credentials to be added
+    ///
+    /// # Errors
+    ///
+    /// Returns a CredentialsStoreError if the implementation cannot add a new
+    /// credential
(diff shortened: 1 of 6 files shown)


libsplinter/src/biome/credentials/store/mod.rs
```diff
@@ -187,8 +187,16 @@ pub trait CredentialsStore<T> {
     ///
     fn fetch_credential_by_username(&self, username: &str) -> Result<T, CredentialsStoreError>;
 
+    /// Fetches the username for a user by user_id
+    ///
+    /// # Arguments
+    ///
+    ///  * `user_id` - The unique identifier of the user the credential belongs to
+    ///
```
peterschwarz: The should include a block that has
```rust
/// # Errors
///
/// Something about the error that may occur
```
and no blank line at the end.

RyanLassigBanks: All the other doc comments in this file have a blank line at the end.

peterschwarz: Sounds like a good thing to add a commit for

PR #635, opened 2020-03-31
Replace "database" feature with mix of diesel, postgres

vaporos:
Users of libsplinter should use the "postgres" feature (or in the
future, the "sqlite" feature) instead of an intermediate layer.

Some code that was protected by "database" is now protected by "diesel"
instead. This code should be enabled indirectly by enabling "postgres".

In serveral places in the biome module, diesel-specific code was not
guarded by a feature; this fixes that as well.


cli/Cargo.toml
```diff
@@ -79,7 +79,7 @@ node-alias = ["dirs", "reqwest"]
 health = ["reqwest", "serde_json"]
 keygen = ["dirs", "whoami"]
 
-database = ["splinter/database", "diesel", "postgres"]
+database = ["splinter/postgres", "diesel", "postgres"]
```
peterschwarz, line 82: Does the cli still need to have diesel and postgres as direct dependencies?

vaporos: I will investigate this separately.

PR #636, opened 2020-04-01
Edit circuit related help statements

shannynalayna:

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -35,7 +35,7 @@ const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
 #[cfg(feature = "circuit")]
-const CIRCUIT_CREATE_AFTER_HELP: &str = r"DETAILS:
+const CIRCUIT_PROPOSE_AFTER_HELP: &str = r"DETAILS:
     One or more nodes must be specified using the --node and/or --node-file arguments. These
     arguments can be used on their own or together, but at least one of them is required.
 
@@ -344,7 +344,7 @@ fn run() -> Result<(), CliError> {
                     .short("n")
                     .help("Print circuit definition without submitting the proposal"),
             )
-            .after_help(CIRCUIT_CREATE_AFTER_HELP);
+            .after_help(CIRCUIT_PROPOSE_AFTER_HELP);
 
         #[cfg(feature = "circuit-auth-type")]
         let propose_circuit = propose_circuit.arg(
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -263,10 +263,9 @@ fn run() -> Result<(), CliError> {
                     .takes_value(true)
                     .required_unless("node_file")
                     .multiple(true)
-                    .long_help(
-                        "Node that is part of the circuit. \
-                         Format: <node_id>::<endpoint>. \
-                         Endpoint is optional if node alias has been set.",
+                    .help(
+                        "Node that is part of a circuit \
+                         (<node_id>::<endpoint>)",
                     ),
             )
             .arg(
@@ -276,9 +275,9 @@ fn run() -> Result<(), CliError> {
                     .multiple(true)
                     .min_values(2)
                     .required_unless("template")
-                    .long_help(
-                        "Service ID and allowed nodes. \
-                         Format <service-id>::<allowed_nodes>",
+                    .help(
+                        "Service ID and allowed nodes \
+                         (<service-id>::<allowed_nodes>)",
                     ),
             )
             .arg(
@@ -286,9 +285,9 @@ fn run() -> Result<(), CliError> {
                     .long("service-arg")
                     .takes_value(true)
                     .multiple(true)
-                    .long_help(
-                        "Special arguments to be passed to the service. \
-                         Format <service_id>::<key>=<value>",
+                    .help(
+                        "Pass arguments to a service \
+                         (<service_id>::<key>=<value>)",
                     ),
             )
             .arg(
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -39,6 +39,8 @@ const CIRCUIT_CREATE_AFTER_HELP: &str = r"DETAILS:
     One or more nodes must be specified using the --node and/or --node-file arguments. These
     arguments can be used on their own or together, but at least one of them is required.
 
+    If the --node argument is specified, 'endpoint' is optional if node alias has been set.
```
ltseeley, line 42: Could be a little clearer: `..., the 'endpoint' portion of the argument is optional if an alias for the specified node has been set using the 'splinter node alias' command.`

cli/src/main.rs
```diff
@@ -39,6 +39,8 @@ const CIRCUIT_CREATE_AFTER_HELP: &str = r"DETAILS:
     One or more nodes must be specified using the --node and/or --node-file arguments. These
```
ltseeley, line 39: Could you rename `CIRCUIT_CREATE_AFTER_HELP` -> `CIRCUIT_PROPOSE_AFTER_HELP` in another commit while you're here? I missed that when I renamed the subcommand.

cli/src/main.rs
```diff
@@ -264,9 +266,8 @@ fn run() -> Result<(), CliError> {
                     .required_unless("node_file")
                     .multiple(true)
                     .long_help(
-                        "Node that is part of the circuit. \
-                         Format: <node_id>::<endpoint>. \
-                         Endpoint is optional if node alias has been set.",
+                        "Node that is part of the circuit \
+                         Format: <node_id>::<endpoint>",
```
ltseeley, line 270: These `Format` and `Default value` help strings are not printed with new lines. This one shows up as: `Node that is part of the circuit Format: <node_id>::<endpoint>`. Maybe a better format is `Description of argument (format: xxx::yyy)`.

cli/src/main.rs
```diff
@@ -398,7 +405,7 @@ fn run() -> Result<(), CliError> {
                             .value_name("circuit-id")
                             .takes_value(true)
                             .required(true)
-                            .help("The circuit id of the proposed circuit"),
+                            .help("Circuit id of the proposed circuit"),
```
ltseeley, line 408: Could even just be "ID of the proposed circuit" (same below)

cli/src/main.rs
```diff
@@ -474,24 +481,24 @@ fn run() -> Result<(), CliError> {
...
                             .short("U")
                             .long("url")
-                            .help("The URL of the Splinter daemon REST API")
+                            .help("URL of the Splinter daemon REST API")
                             .takes_value(true),
                     )
                     .arg(
                         Arg::with_name("management_type")
                             .long("management-type")
                             .long_help(
```
ltseeley, line 490: I think most or all of these can be changed to just `help` instead of `long_help` since they're all short now.

cli/src/main.rs
```diff
@@ -580,7 +587,7 @@ fn run() -> Result<(), CliError> {
                                 .required(true)
                                 .takes_value(true)
                                 .value_name("name")
-                                .help("The name of the template"),
+                                .help("Template name"),
```
ltseeley, line 590: This isn't consistent with the help strings above (those are "Name of X")

cli/src/main.rs
shannynalayna: Reformatting to @vaporos's suggestion `(xxx::yyy)`

cli/src/main.rs
shannynalayna: Getting rid of this sentence, per @vaporos's comment below

cli/src/main.rs
shannynalayna: Addressed

cli/src/main.rs
shannynalayna: Addressed

cli/src/main.rs
shannynalayna: Addressed

cli/src/main.rs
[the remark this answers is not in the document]
shannynalayna: Excluded this line from the after_help

cli/src/main.rs
```diff
@@ -264,9 +266,8 @@ fn run() -> Result<(), CliError> {
                     .required_unless("node_file")
                     .multiple(true)
                     .long_help(
-                        "Node that is part of the circuit. \
-                         Format: <node_id>::<endpoint>. \
-                         Endpoint is optional if node alias has been set.",
+                        "Node that is part of the circuit \
```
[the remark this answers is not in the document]
shannynalayna: Addressed throughout

cli/src/main.rs
```diff
@@ -287,7 +288,7 @@ fn run() -> Result<(), CliError> {
                     .takes_value(true)
                     .multiple(true)
                     .long_help(
-                        "Special arguments to be passed to the service. \
+                        "Arguments to be passed to the service \
```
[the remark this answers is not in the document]
shannynalayna: Addressed

cli/src/main.rs
```diff
@@ -320,27 +319,30 @@ fn run() -> Result<(), CliError> {
...
             .arg(
                 Arg::with_name("metadata_encoding")
                     .long("metadata-encoding")
                     .takes_value(true)
                     .possible_values(&["json", "string"])
                     .requires("metadata")
-                    .help("Set the encoding for the application metadata. Default value: string"),
+                    .help(
+                        "Set encoding of application metadata \
+                           Default value: string",
```
ltseeley, line 332: Can this line be put in parentheses or brackets? Currently it displays as `Set encoding of application metadata Default value: string [possible values: json, string]` Maybe `... metadata (default value: string) [possible values: ...]` would look best?

shannynalayna: Took out `value` just to cut down on characters but went with your formatting suggestion

PR #639, opened 2020-04-01
Shutdown REST API on test completion

peterschwarz:
When testing REST API endpoints, it is necessary to shutdown the REST API server, so that the all the in-use sockets are closed.  Without this change, tests will fail due to "Too many open files", particularly on MacOS, but may begin to occur on linux, etc, with the addition of more
REST API tests.

### DIFF
--- a/libsplinter/src/admin/rest_api/actix/circuits.rs
+++ b/libsplinter/src/admin/rest_api/actix/circuits.rs
@@ -169,7 +169,7 @@ mod tests {
     #[test]
     /// Tests a GET /admin/circuits request with no filters returns the expected circuits.
     fn test_list_circuits_ok() {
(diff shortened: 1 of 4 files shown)


libsplinter/src/admin/rest_api/actix/circuits.rs
```diff
@@ -169,7 +169,7 @@ mod tests {
     #[test]
     /// Tests a GET /admin/circuits request with no filters returns the expected circuits.
     fn test_list_circuits_ok() {
-        let (_shutdown_handle, _join_handle, bind_url) =
+        let (shutdown_handle, join_handle, bind_url) =
```
ltseeley, line 172: Typo in commit message: "the all the"

PR #642, opened 2020-04-01
Configure RefreshTokenStore with connection pool

peterschwarz:
Configure the RefreshTokenStore like the other stores in the resource builder: via a connection pool.

Additionally:

- if the "biome-refresh-tokens" feature is enabled, configure it in the   splinter daemon.
- make refresh::store::diesel pub(crate) to support this usage.


peterschwarz:
Confirmed that this runs in gameroom with experimental features

PR #645, opened 2020-04-01
Biome rest api fixes

RyanLassigBanks:
Various fixes for biome rest api that were requested as part of stabilization.

### DIFF
--- a/libsplinter/src/biome/rest_api/config.rs
+++ b/libsplinter/src/biome/rest_api/config.rs
@@ -39,19 +39,27 @@ pub struct BiomeRestConfig {
 }
 
 impl BiomeRestConfig {
+    /// Returns token issuer string. Defaults to "self-issued".
     pub fn issuer(&self) -> String {
         self.issuer.to_owned()
     }
 
+    /// Returns duration that the access token is valid.
+    /// Defaults to 90 minutes.
     pub fn access_token_duration(&self) -> Duration {
         self.access_token_duration.to_owned()
     }
 
+    /// Returns durations the refresh token is valid.
+    /// Defaults to 60 days.
     #[cfg(feature = "biome-refresh-tokens")]
     pub fn refresh_token_duration(&self) -> Duration {
         self.refresh_token_duration.to_owned()
     }
 
+    /// Returns the password encryption cost. This roughly equates to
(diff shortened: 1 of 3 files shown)


libsplinter/src/biome/credentials/store/diesel/mod.rs
```diff
@@ -39,7 +39,7 @@ impl SplinterCredentialsStore {
     /// # Arguments
     ///
     ///  * `connection_pool`: connection pool to the database
-    #[cfg(feature = "biome-rest-api")]
+    #[allow(dead_code)]
```
vaporos, line 42: This was correct before -- the guard should just be removed when we remove biome-rest-api feature.

libsplinter/src/biome/rest_api/config.rs
```diff
@@ -39,19 +39,27 @@ pub struct BiomeRestConfig {
 }
 
 impl BiomeRestConfig {
+    /// Returns token issuer string. Defaults to "self-issued"
```
vaporos, line 46: I feel like if it has one period, then everything has to end in periods. Maybe: "Return token issuer string (default: "self-issued")."

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -127,15 +127,10 @@ use crate::rest_api::sessions::AccessTokenIssuer;
 /// * `GET /biome/user/{id}` - Retrieve user with specified ID
 /// * `DELETE /biome/user/{id}` - Remove user with specified ID
 pub struct BiomeRestResourceManager {
-    // Disable lint warning, for now this is only used if the biome-credentials feature is enabled
```
vaporos, line 109: These should be changed to #[cfg(feature = "biome-rest-api")] until that feature is removed.

libsplinter/src/biome/rest_api/config.rs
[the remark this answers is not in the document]
RyanLassigBanks: The other doc comments in Biome are written in an [active voice](https://github.com/Cargill/splinter/blob/935d412c3f3c8370d18dfaefaca1eba597e542fd/libsplinter/src/biome/key_management/store/mod.rs#L34), so I'm trying to be consistent with those.

PR #646, opened 2020-04-01
Refactor DatabaseError into ConnectionError

RyanLassigBanks:
The enum DatabaseError is now the struct ConnectionError, this is
because to only variant of DatabaseError that was being used was
ConnectionError.

### DIFF
--- a/libsplinter/src/biome/refresh_tokens/store/error.rs
+++ b/libsplinter/src/biome/refresh_tokens/store/error.rs
@@ -16,7 +16,7 @@ use std::error::Error;
 use std::fmt;
 
 #[cfg(feature = "diesel")]
-use crate::database::error::DatabaseError;
+use crate::database::error;
 
 #[derive(Debug)]
 pub enum RefreshTokenError {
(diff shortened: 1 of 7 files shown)


libsplinter/src/database/mod.rs
```diff
@@ -26,10 +26,13 @@ use diesel::{
...
 };
 
-pub use super::database::error::DatabaseError;
+pub use super::database::error::ConnectionError;
 
-pub fn run_migrations(conn: &PgConnection) -> Result<(), DatabaseError> {
-    embedded_migrations::run(conn).map_err(|err| DatabaseError::ConnectionError(Box::new(err)))?;
+pub fn run_migrations(conn: &PgConnection) -> Result<(), ConnectionError> {
+    embedded_migrations::run(conn).map_err(|err| ConnectionError {
+        context: "Failed to embed migreations".to_string(),
```
vaporos, line 33: Typo - migreations

libsplinter/src/biome/refresh_tokens/store/error.rs
```diff
@@ -36,7 +36,7 @@ pub enum RefreshTokenError {
         source: Box<dyn Error>,
     },
     /// Represents an issue connecting to the database
-    ConnectionError(Box<dyn Error>),
+    ConnectionError(error::ConnectionError),
```
vaporos, line 39: The switch from Box\<dyn Error\> to error::ConnectionError here is unnecessary and results in a compilation error.

PR #647, opened 2020-04-01
Move database migrations into biome/migrations

RyanLassigBanks:


vaporos:
This PR will need to purge the migrations stuff out of crate::database as well (thus the compile error).

PR #651, opened 2020-04-02
Proto conversion traits

peterschwarz:
This PR takes the various proto<->native traits from transact, updates the code based on the latest Rust (1.42 as of this PR), as well as makes some modifications to support more auto trait implementations. 

The largest change is `To-/FromBytes` which adds a generic parameter, `ViaProtocol`.  This allows for the auto-trait implementation for traits where the `ViaProtocol` is a `protobuf::Message`.

Additionally, `IntoNative/-Proto` have also been turned into auto trait implementations, vs default function implementations.

With these updates, consumers of the api only need to implement `FromNative` and `FromProto` for their structs.

A simplification for users of these traits is to also re-export the traits and error via a prelude module.  Use can now do:

```
use splinter::protos::prelude::*;
```

and gain all the benefits of the traits.

### DIFF
--- a/libsplinter/src/protos/mod.rs
+++ b/libsplinter/src/protos/mod.rs
@@ -48,12 +48,37 @@ pub trait FromNative<N>: Sized {
     fn from_native(other: N) -> Result<Self, ProtoConversionError>;
 }
 
-pub trait FromBytes<N>: Sized {
-    fn from_bytes(bytes: &[u8]) -> Result<N, ProtoConversionError>;
+/// Convert from bytes into a native struct, via the given protocol.
+pub trait FromBytes<ViaProtocol>: Sized {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, ProtoConversionError>;
 }
 
-pub trait IntoBytes: Sized {
+impl<P, N> FromBytes<P> for N
+where
+    P: protobuf::Message,
+    N: FromProto<P>,
+{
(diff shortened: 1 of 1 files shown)


libsplinter/src/protos/mod.rs
```diff
@@ -12,4 +13,110 @@
...
+where
+    P: FromNative<N>,
+{
+    fn into_proto(self) -> Result<P, ProtoConversionError> {
+        FromNative::from_native(self)
+    }
+}
+
+pub mod prelude {
+    //! Allows for he convenient requirement of the proto convertions traits.
```
agunde406, line 111: typo "he" -> "the"

libsplinter/src/protos/mod.rs
```diff
@@ -81,20 +81,28 @@ where
     }
 }
 
-pub trait IntoNative<T>: Sized
+pub trait IntoNative<T>: Sized {
```
vaporos, line 84: Typo in the commit message "implemetnations"

libsplinter/src/protos/mod.rs
peterschwarz: Fixed

libsplinter/src/protos/mod.rs
peterschwarz: Updated

peterschwarz:
Added the commit info to the first commit.

PR #653, opened 2020-04-02
Make credentials store mandatory

RyanLassigBanks:
When the biome-credentials feature is enabled a credentials store needs
to be supplied, this so it is consistent with how other stores are
handled.


libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -125,7 +125,7 @@ pub struct BiomeRestResourceManager {
     #[cfg(feature = "biome-refresh-tokens")]
     refresh_token_store: Arc<dyn RefreshTokenStore>,
     #[cfg(feature = "biome-credentials")]
-    credentials_store: Option<Arc<DieselCredentialsStore>>,
+    credentials_store: Arc<DieselCredentialsStore>,
```
ltseeley, line 128: Typo in commit message: "this so it is"

PR #656, opened 2020-04-02
Remove json-web-tokens feature

RyanLassigBanks:
Required in order to move credentials, key management, rest-api, and
refresh token features into stable.


libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -132,13 +116,12 @@ use crate::rest_api::sessions::AccessTokenIssuer;
 pub struct BiomeRestResourceManager {
     #[cfg(feature = "biome-rest-api")]
     user_store: DieselUserStore,
-    #[cfg(all(feature = "biome-key-management", feature = "json-web-tokens"))]
+    #[cfg(all(feature = "biome-key-management"))]
```
agunde406, line 119: This all can be removed

RyanLassigBanks: updated

PR #660, opened 2020-04-03
Add some splinter-circuit-* man pages

shannynalayna:
Adds man pages for `splinter-circuit-proposals`, `splinter-circuit-list`, and `splinter-circuit-propose`.

### DIFF
--- a/cli/man/splinter-circuit-propose.1.md
+++ b/cli/man/splinter-circuit-propose.1.md
@@ -1,0 +1,188 @@
+% SPLINTER-CIRCUIT-PROPOSE(1) Cargill, Incorporated | Splinter Commands
+
+NAME
+====
+
+**splinter-circuit-propose** — Propose that a new circuit is created
+
+SYNOPSIS
+========
+**splinter circuit propose** \[**FLAGS**\] \[**OPTIONS**\]
+
+DESCRIPTION
+===========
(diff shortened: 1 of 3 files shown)


cli/man/splinter-circuit-propose.1.md
```diff
@@ -0,0 +1,187 @@
...
+: Specifies the service ID and allowed nodes, using the format
+  `<service-id>::<allowed_nodes>`. The <allowed_nodes> specifies the node which
+  the service will run on, currently only one node ID is allowed.
+
+`--service-arg <service_argument>` ...
+: Passes key/value arguments to the specified service (as defined by
+  `--service`), using the format `<service_id>::<key>=<value>`. Service arguments
+  provided must match those required to create the service. The glob operator,
+  `*`, may be used in place of the <service_id> to match all or certain parts
+  of the <service_id>. For instance, `alpha*::<key>=<value>` to match all service
```
eloaverona, line 83: `alpha` is not a valid `service_id`. Service IDs have a max of 4 char in length and must be ASCII alphanumeric characters.   All the examples of service IDs in the man page should show service IDs in the correct format. This is a good example to look at for this case
https://github.com/Cargill/splinter/blob/master/cli/src/main.rs#L50-L54

PR #661, opened 2020-04-03
Even more circuit CLI improvements

ltseeley:
A few small improvements to the circuit CLI and one big change that removes the `splinter circuit default` subcommands.


cli/src/action/circuit/payload.rs
```diff
@@ -71,21 +74,21 @@ where
...
-        CliError::ActionError(format!("Unable to serialize header to bytes: {}", err))
-    })?;
+    let header_bytes = header
+        .write_to_bytes()
+        .map_err(|_| CliError::ActionError("Failed to serialize payload header".into()))?;
 
     let header_signature = signer
         .sign(&header_bytes)
-        .map_err(|err| CliError::ActionError(format!("Unable to sign payload header: {}", err)))?;
+        .map_err(|_| CliError::ActionError("Failed to sign payload header".into()))?;
```
peterschwarz, line 83: Since we have logging setup, should we log the error to `debug!` so it would show-up during verbose logging?

ltseeley: Changed the commit so it doesn't remove the internal errors at the request of @vaporos

PR #662, opened 2020-04-04
Add MeshShutdownSignaler and update Mesh to shutdown

agunde406:
This commit updates InternalEnvelope to be an Enum with
Message and Shutdown options. Mesh will now has a function
that will return a MeshShutdownSignaler that can be used
to shutdown Mesh. 

A function has also been added to Network to return a
MeshShutdownSignaler as well as Network is just a wrapper
around Mesh.

### DIFF
--- a/libsplinter/src/mesh/mod.rs
+++ b/libsplinter/src/mesh/mod.rs
@@ -65,11 +65,12 @@ use std::time::Duration;
 
 #[cfg(feature = "matrix")]
 pub use crate::matrix::Envelope;
-pub use crate::mesh::control::{AddError, Control, RemoveError};
-pub use crate::mesh::incoming::Incoming;
+use crate::mesh::control::Control;
+pub use crate::mesh::control::{AddError, RemoveError};
+use crate::mesh::incoming::Incoming;
 #[cfg(feature = "matrix")]
 pub use crate::mesh::matrix::{MeshLifeCycle, MeshMatrixSender};
-pub use crate::mesh::outgoing::Outgoing;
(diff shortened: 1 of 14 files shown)


libsplinter/src/mesh/mod.rs
```diff
@@ -558,4 +582,65 @@ mod tests {
         let tls = create_test_tls_transport(true);
         test_add_remove_connections(tls, "127.0.0.1:0");
     }
+
+    #[test]
+    // Test mesh can be shutdown
```
vaporos, line 587: Need more descriptive test comments - see PBFT tests for examples.

agunde406: updated

libsplinter/src/mesh/mod.rs
```diff
@@ -76,28 +76,10 @@ use crate::mesh::reactor::Reactor;
...
-
-    pub fn payload(&self) -> &[u8] {
-        &self.payload
-    }
-
-    pub fn take_payload(self) -> Vec<u8> {
-        self.payload
-    }
+#[derive(Debug, PartialEq)]
+pub enum InternalEnvelope {
```
peterschwarz, line 80: Should this be pub? Also, with the changes where the signalling is via the `Control` struct, is this still needed to be an enum?

agunde406: I kept InternalEnvelope an enum so the recv would still get the shutdown notification. It only needs to be available in with the mesh mod but no where else. There are also a couple other components that are public but do not need to be. I added a commit to reduce their visibility to mesh.

PR #663, opened 2020-04-06
Log to stdout

peterschwarz:
Flexi Logger's default output is stderr.  This commit changes all the ouput targets to stdout.


peterschwarz:
Missed a couple of additional spots

PR #664, opened 2020-04-06
Rename key routes /biome/users/keys to /biome/keys

RyanLassigBanks:
/biome/users/keys conflicted with /biome/user/{id} causing attempts to
create new keys to return a 404.


vaporos:
This looks good - please prep an entry in the upgrade doc.

PR #665, opened 2020-04-06
Add more `splinter-circuit` man pages

shannynalayna:
Adds man pages for the following splinter-circuit commands:
* splinter circuit show
* splinter circuit vote
* splinter circuit

### DIFF
--- a/cli/man/splinter-circuit-show.1.md
+++ b/cli/man/splinter-circuit-show.1.md
@@ -1,0 +1,106 @@
+% SPLINTER-CIRCUIT-SHOW(1) Cargill, Incorporated | Splinter Commands
+
+NAME
+====
+
+**splinter-circuit-show** — Displays information about a circuit
+
+SYNOPSIS
+========
+**splinter circuit show** \[**FLAGS**\] \[**OPTIONS**\] <circuit>
+
+DESCRIPTION
+===========
+Display the entire definition of a circuit or proposal that the node is a member
+or proposed member of. All members, or proposed members, may view the circuit
+definition. Viewing a proposed circuit enables nodes to view information pertaining
+to other proposed member nodes as well as the status of a node’s vote regarding
+the circuit. The proposed circuit will be viewable unless any proposed member nodes
+reject the circuit proposal.
+
+FLAGS
+=====
+`-h`, `--help`
+: Prints help information
+
+`-q`, `--quiet`
+: Decrease verbosity (the opposite of -v). When specified, only errors or
+  warnings will be output.
+
+`-V`, `--version`
+: Prints version information
+
+`-v`
+: Increases verbosity (the opposite of -q). Specify multiple times for more
+  output.
+
+OPTIONS
+=======
+`-f`, `--format` <format>
+: Specifies the output format of the circuit proposal. (default `human`).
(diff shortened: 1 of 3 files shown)


cli/man/splinter-circuit-show.1.md
```diff
@@ -0,0 +1,106 @@
...
+**splinter-circuit-show** — Displays information about a circuit
+
+SYNOPSIS
+========
+**splinter circuit show** \[**FLAGS**\] \[**OPTIONS**\] <circuit>
+
+DESCRIPTION
+===========
+Display the entire definition of a circuit or proposal that the node is a member
+or proposed member of. All  members, or proposed members, may view the circuit
```
vaporos, line 15: two spaces between "All" and "members"

cli/man/splinter-circuit-show.1.md
```diff
@@ -0,0 +1,106 @@
...
+        Service: AA01
+            admin_keys:
+                <alpha-public-key>
+            peer_services:
+                AA01
+```
+
+ENVIRONMENT VARIABLES
+=====================
+**SPLINTER_REST_API_URL_ENV**
```
vaporos, line 97: I think this should be **SPLINTER_REST_API_URL** without the _ENV.

cli/man/splinter-circuit-vote.1.md
```diff
@@ -0,0 +1,93 @@
...
+: Specify the circuit ID of the circuit to be voted on.
+
+EXAMPLES
+========
+* The proposed circuit has ID `1234-ABCDE`.
+
+The following command displays a member node voting to accept the circuit proposal:
+```
+$ splinter circuit vote \
+  --key <proposed-member-node-private-key> \
```
vaporos, line 67: All examples should try and avoid shell characters such as "<" and ">" because a user cut-n-pasting these will have a bad day.

shannynalayna: The intention is that the user would replace `<...>` with their own private key, so what would you suggest we use instead of the <> characters to display that? I think these characters are used pretty widely across the man pages where it is necessary

cli/man/splinter-circuit-show.1.md
```diff
@@ -0,0 +1,106 @@
...
+
+OPTIONS
+=======
+`-f`, `--format` <format>
+: Specifies the output format of the circuit proposal. (default `human`).
+  Possible values for formatting are `human` and `csv`.
+
+`-U`, `--url` <url>
+: Specifies the URL for the `splinterd` REST API. The URL is required unless
+  `$SPLINTER_REST_API_URL_ENV` is set.
```
chenette, line 45: No `_ENV` extension.

cli/man/splinter-circuit-vote.1.md
chenette: Possible approaches that I've used in the past:
- Curly braces instead of angle brackets. They're still special chars, but they have less horrible effects when used.

- No brackety characters, but put the replaceable item in ALL-CAPS to make it extremely obvious. 

Best practice is to use a short string for the replaceable item, then explain the details in the text above. For example, "Replace MEMBER-KEY with  the private key of the proposed member node."

PR #667, opened 2020-04-06
Add initial network test for splinter

rbuysse:
This test simulates a brief network interruption between nodes with a created
circuit which causes the splinter nodes to disconnect. When the interruption is
resolved, a second circuit is proposed and created to verify reconnection.

The setting "net.ipv4.tcp_retries2=3" in the Compose file is used to limit the
number of retries before the TCP stack will inform the layers above that the
connection has been broken. tcp_retries2 uses an exponential backoff starting at
approximately 200ms, so a tcp_retries2 value of "3" should cause a timeout
after about 1.4 seconds.

More detailed information about the network settings can be found at
https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt

### DIFF
--- a/tests/test-splinter-network.yaml
+++ b/tests/test-splinter-network.yaml
@@ -1,0 +1,98 @@
+# Copyright 2018-2020 Cargill Incorporated
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


rbuysse:
Run `docker-compose -f tests/test-splinter-network.yaml up --abort-on-container-exit` to test.

tests/test-splinter-network.yaml
```diff
@@ -0,0 +1,98 @@
...
+# limitations under the License.
+
+version: "3.7"
+
+volumes:
+  key-registry:
+
+services:
+
+  key-registry:
```
vaporos, line 22: The name of this service is a bit confusing given that it does more now than create the registry. Maybe "orchestrator".

PR #669, opened 2020-04-07
Remove biome-rest-api feature

RyanLassigBanks:
PR removing `biome-rest-api` feature

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -96,7 +96,6 @@ experimental = [
     "biome-key-management",
     "biome-notifications",
     "biome-refresh-tokens",
-    "biome-rest-api",
     "biome-user",
     "circuit-template",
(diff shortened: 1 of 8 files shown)


libsplinter/Cargo.toml
```diff
@@ -41,7 +41,7 @@ diesel = { version = "1.0", features = ["r2d2", "serde_json"], optional = true }
 diesel_migrations = { version = "1.4", optional = true }
 futures = { version = "0.1", optional = true }
 hyper = { version = "0.12", optional = true }
-jsonwebtoken = { version = "6.0", optional = true }
+jsonwebtoken = "6.0"
```
agunde406, line 44: Should this still be optional, and only included if rest-api is enabled.

RyanLassigBanks: Updated

PR #675, opened 2020-04-08
Simplify admin's RestResourceProvider impls

ltseeley:
Simplifies the admin service's `RestResourceProvider` implementations to
be easier to read and avoid lint.

### DIFF
--- a/libsplinter/src/admin/rest_api/mod.rs
+++ b/libsplinter/src/admin/rest_api/mod.rs
@@ -23,19 +23,6 @@ use crate::admin::service::AdminService;
 use crate::circuit::store;
 use crate::rest_api::{Resource, RestResourceProvider};
 
-#[cfg(feature = "rest-api-actix")]
-use self::actix::circuits::make_list_circuits_resource;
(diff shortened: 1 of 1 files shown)


libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -49,20 +36,19 @@ use self::actix::ws_register_type::make_application_handler_registration_route;
 ///
 /// * `rest-api-actix`
 impl RestResourceProvider for AdminService {
+    #[cfg(feature = "rest-api-actix")]
```
vaporos, line 39: I like the older pattern better (creating a vec, then filling it), because when we have more than one backend things are always stated in the positive, not "not x, y, z" as the empty implementation is now. A compromise would be to do:
```
#[cfg(feature = "rest-api-actix")]
{
    resources.push(actix::ws_register_type::make_application_handler_registration_route(self.commands()));
    ...
}
```

Thus reducing the cfg statements but leaving one resources() function.

ltseeley: Updated

PR #676, opened 2020-04-08
Add Dispatch Source

peterschwarz:
Handlers may have varied source information.  That is, the messages might either be on the connection level, where the source is identified by a connection id, or on the peer level, where the source is identified as a peer id (and the connection id is irrelevant).

This change introduces a new generic parameter on dispatcher-related structs, `Source` that specifies this aspect of the context.  It introduces the `Source` associated type to the Handler trait to specify this type on specific handlers.

This type parameter should be one of two kinds of values: `ConnectionId` or `PeerId`.

Source defaults to `PeerId` in the dispatcher-related structs, where possible.

Handlers have also been changed from using generics to using associated types.

### DIFF
--- a/libsplinter/src/circuit/handlers/admin_message.rs
+++ b/libsplinter/src/circuit/handlers/admin_message.rs
@@ -29,11 +29,14 @@ pub struct AdminDirectMessageHandler {
     state: SplinterState,
 }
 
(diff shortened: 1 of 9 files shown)


peterschwarz:
rebased on latest

PR #682, opened 2020-04-09
Remove the use of concrete stores from Biome RestApi

RyanLassigBanks:
Replaces uses of `DieselUserStore`, `DieselCredentialStore`, and
`PostgresKeyStore` in BiomeRestResourceManager with
`Arc<dyn UserStore>`, `Arc<dyn CredentialsStore>`, and `Arc<dyn
KeyStore<Key>>`. This makes BiomeRestResourceManager testable since it
can now be backed by mock versions of the stores or versions of the
stores that store objects in memory.

Additionally, the signatures of `with_user_store`, `with_credentials_store`,
`with_key_store` and `with_refresh_token_store` have been changed to take
trait objects of their respective stores as arguments instead of
`ConnectionsPool`s to account for store implementations that are not
backed by `ConnectionPool`.

### DIFF
--- a/libsplinter/src/biome/credentials/store/mod.rs
+++ b/libsplinter/src/biome/credentials/store/mod.rs
@@ -140,7 +140,7 @@ impl CredentialsBuilder {
 
 /// Defines methods for CRUD operations and fetching a user’s
 /// credentials without defining a storage strategy
-pub trait CredentialsStore {
+pub trait CredentialsStore: Send + Sync {
     /// Adds a credential to the underlying storage
     ///
     /// # Arguments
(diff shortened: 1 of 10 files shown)


libsplinter/src/biome/credentials/store/mod.rs
```diff
@@ -140,7 +140,7 @@ impl CredentialsBuilder {
 
 /// Defines methods for CRUD operations and fetching a user’s
 /// credentials without defining a storage strategy
-pub trait CredentialsStore {
```
agunde406, line 143: typo in commit message: contrete -> concrete

PR #686, opened 2020-04-10
Add release notes for v0.3.15

RyanLassigBanks:

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,189 @@
 # Release Notes
 
+## Changes in Splinter 0.3.15
+
+### Highlights
+
+* Command name change: Use `splinter circuit propose` to propose a new circuit
+  (instead of  `splinter circuit create`).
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,189 @@
...
+
+* The REST API tests now shut down after the tests finish.
+
+#### Biome
+
+* Change two Biome routes so that the user ID is not required in the route
+    - `biome/users/{USER-ID}/keys` has changed to to `biome/keys`
+    - `biome/users/{USER-ID}/keys/{public_key}` has changed to
+      `biome/keys/{public_key}`
+   The user ID is now derived from the provided access token.
```
chenette, line 75: This is supposed to be on its own line, but it appears that github markdown won't do that. As a work-around, I propose moving this line up with the first one: 

> * Change two Biome routes so that the user ID is not required in the route. The user ID is now derived from the provided access token.

PR #687, opened 2020-04-10
Stabilize biome-key-management feature

RyanLassigBanks:

### DIFF
--- a/libsplinter/src/biome/rest_api/mod.rs
+++ b/libsplinter/src/biome/rest_api/mod.rs
@@ -119,7 +119,7 @@ use crate::rest_api::sessions::AccessTokenIssuer;
 /// * `GET /biome/user/{id}` - Retrieve user with specified ID
 /// * `DELETE /biome/user/{id}` - Remove user with specified ID
 pub struct BiomeRestResourceManager {
-    #[cfg(any(feature = "biome-key-management", feature = "biome-credentials",))]
+    #[cfg(feature = "biome-credentials")]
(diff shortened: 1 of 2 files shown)


libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -117,7 +117,7 @@ use crate::rest_api::sessions::AccessTokenIssuer;
 /// * `GET /biome/user/{id}` - Retrieve user with specified ID
 /// * `DELETE /biome/user/{id}` - Remove user with specified ID
 pub struct BiomeRestResourceManager {
-    #[cfg(any(feature = "biome-key-management", feature = "biome-credentials",))]
+    #[cfg(feature = "biome-credentials")]
```
vaporos, line 120: These guards would presumably still be necessary when we have the feature turned off?

vaporos:
In the other store modules, key_management::store::diesel::postgres::DieselKeyStore is up a level (we don't have a postgres-specific module in that location. Probably that should be moved.

libsplinter/src/biome/rest_api/mod.rs
RyanLassigBanks: Based on the way the routes are set up, the user store only appears to be required by the routes that are enabled by the `biome-credentials` feature

vaporos: Ok, please move that fix to another PR.

PR #690, opened 2020-04-10
Add new splinter-keygen man page

chenette:

### DIFF
--- a/cli/man/splinter-keygen.1.md
+++ b/cli/man/splinter-keygen.1.md
@@ -1,0 +1,77 @@
+% SPLINTER-KEYGEN(1) Cargill, Incorporated | Splinter Commands
+
+NAME
+====
+
+**splinter-keygen** — Generates user and daemon keys for Splinter
+
+SYNOPSIS
+========
+
+**splinter keygen** \[**FLAGS**\] \[**OPTIONS**\] \[*KEY-NAME*\]
+
+DESCRIPTION
+===========
+
+This command generates secp256k1 public/private keys for Splinter.
+
+If no option is specified, this command generates user keys that are stored in
+the directory `$HOME/splinter/keys`. The `--system` flag generates keys for the
+Splinter daemon (`splinterd`) that are stored in `/etc/splinter/keys`.
+
+The file names are determined by the user name, unless the `*KEY-NAME*` argument
+is used.
+
+FLAGS
+=====
+
+`-h`, `--help`
+: Prints help information
+
+`-q`, `--quiet`
+: Decreases verbosity (the opposite of -v). When specified, only errors or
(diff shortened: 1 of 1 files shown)


chenette:
Pushed a fixup commit to clarify the --system description: the default key file names are root.priv and root.pub, because this option runs as root.

Also added two examples: One for a user with the default file names, and one for --system (root) with a different basename.

cli/man/splinter-keygen.1.md
```diff
@@ -0,0 +1,78 @@
...
+: Prints help information
+
+`-q`, `--quiet`
+: Decreases verbosity (the opposite of -v). When specified, only errors or
+  warnings will be output.
+
+`--system`
+: Generates system keys for `splinterd` in `/etc/splinter/keys`.
+  By default, the key files are named `root.priv` and `root.pub`, because
+  this command runs as root when the `--system` option is used.
```
vaporos, line 37: This isn't a fundamental attribute of the command - just how we normally use it. The command doesn't escalate to root, and this might suggest that it does. Perhaps change this to an example - "The key filenames are determined by username, so if running as root the filenames will be 'root.priv' and 'root.pub'."

cli/man/splinter-keygen.1.md
```diff
@@ -0,0 +1,78 @@
...
+`paulbunyan`.
+
+```
+$ splinter keygen
+writing file: "/Users/paulbunyan/splinter/keys/paulbunyan.priv"
+writing file: "/Users/paulbunyan/splinter/keys/paulbunyan.pub"
+```
+
+The next example generates system keys for the Splinter daemon, but specifies
+`splinterd` as the basename for the files (instead of the default `root`).
```
vaporos, line 66: This is misleading as the default is the username, not 'root'.

cli/man/splinter-keygen.1.md
chenette: > ... so if running as root the filenames will be 'root.priv' and 'root.pub'."

Is there a requirement to run --system as root?  Will the system (daemon) keys be functional if you run this option as any-old-user?

chenette: ^ (above question is for @vaporos )

vaporos: You could run splinterd as another user, it does not require root permission. Within docker files, it is often running as root.

chenette: Fixed by adding a clarification to the DESCRIPTION and removing all mentions of the command running as root.

cli/man/splinter-keygen.1.md
chenette: Changed to "(instead of the user name").

PR #694, opened 2020-04-13
Update splinterd registry config

ltseeley:
Updates the way node registries are configured and how splinterd creates them:
- Use the `UnifiedNodeRegistry` in splinterd
- Allow configuring read-only registries (currently just YAML files) for splinterd with the `registries` config option
- Disable configuring the local, writable node registry for splinterd in favor of using a hard-coded file in the splinter data directory


splinterd/src/config/builder.rs
```diff
@@ -228,6 +228,14 @@ impl ConfigBuilder {
                     None => None,
                 })
                 .ok_or_else(|| ConfigError::MissingValue("registry file".to_string()))?,
+            registries: self
```
agunde406, line 215: typo in commit message `URIs with be supported` -> `URIs will be supported`

ltseeley: Fixed

splinterd/src/main.rs
```diff
@@ -247,20 +242,22 @@ fn start_daemon(matches: ArgMatches) -> Result<(), UserError> {
         }
     };
 
+    let local_node_registry_location = format!("{}{}", config.state_dir(), "nodes.yaml");
```
vaporos, line 245: This should probably use Path stuff to do the join. If state_dir() doesn't have a "/", this would result in an unintended path.

splinterd/src/main.rs
```diff
@@ -247,20 +242,22 @@ fn start_daemon(matches: ArgMatches) -> Result<(), UserError> {
...
     #[cfg(feature = "database")]
     let db_url = config.database();
 
-    let registry_backend = config.registry_backend();
-
     let admin_service_coordinator_timeout = config.admin_service_coordinator_timeout();
 
     config.log_as_debug();
 
+    #[allow(unused_mut)]
```
vaporos, line 256: A much better approach is to use it always mutable.

```
let mut daemon_builder = SplinterDaemonBuilder::new();

daemon_builder
    .with_storage_location(storage_location)
    .with_key_registry_location(key_registry_location)
    ...
```

splinterd/src/main.rs
ltseeley: Updated all of these path constructors to use `Path` in a new commit

splinterd/src/main.rs
ltseeley: Updated

PR #699, opened 2020-04-14
splinter-dev image updates

rbuysse:


ci/splinter-dev
```diff
@@ -119,6 +123,18 @@ RUN rm \
     target/release/deps/pcounter* \
     target/release/deps/private_counter* \
     target/release/deps/private_xo* \
+    target/release/deps/scabbard* \
     target/release/deps/splinter_cli* \
     target/release/deps/splinterd*
 
+# Clean up leftover files
```
vaporos, line 130: This misses the additional scabbard files. Would it be better just to make this another stage and copy the dependencies explicitly? Then you don't have to maintain an rm list and might be less error prone.

PR #704, opened 2020-04-14
Rest api unit tests

RyanLassigBanks:
Unit tests for biome rest api

### DIFF
--- a/libsplinter/src/biome/rest_api/mod.rs
+++ b/libsplinter/src/biome/rest_api/mod.rs
@@ -413,3 +413,845 @@ impl BiomeRestResourceManagerBuilder {
         })
     }
 }
+
+#[cfg(test)]
+#[cfg(all(
+    feature = "biome-key-management",
+    feature = "biome-credentials",
+    feature = "biome-refresh-tokens"
+))]
+mod tests {
+    use super::*;
+
+    use std::{panic, thread};
+
+    use reqwest::blocking::Client;
+
+    use crate::biome::{
+        MemoryCredentialsStore, MemoryKeyStore, MemoryRefreshTokenStore, MemoryUserStore,
+    };
+    use crate::rest_api::{RestApiBuilder, RestApiShutdownHandle};
+
+    #[derive(Serialize)]
+    struct UsernamePassword {
+        pub username: String,
+        pub hashed_password: String,
+    }
+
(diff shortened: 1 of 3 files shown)


libsplinter/src/biome/user/store/memory.rs
```diff
@@ -76,6 +89,14 @@ impl UserStore for MemoryUserStore {
             })?;
 
         if inner.remove(id).is_some() {
+            #[cfg(feature = "biome-credentials")]
+            self.credentials_store
+                .remove_credentials(id)
+                .map_err(|err| UserStoreError::QueryError {
+                    context: "Cannot delete credentials store".to_string(),
```
shannynalayna, line 96: I think this error message is slightly off, maybe try rewording to `"Cannot delete {} from credentials store", id`

RyanLassigBanks: Changed to `Cannot delete user {} from credentials store`

libsplinter/src/biome/rest_api/mod.rs
```diff
@@ -413,3 +413,793 @@ impl BiomeRestResourceManagerBuilder {
...
+
+    use reqwest::blocking::Client;
+
+    use crate::biome::{
+        MemoryCredentialsStore, MemoryKeyStore, MemoryRefreshTokenStore, MemoryUserStore,
+    };
+    use crate::rest_api::RestApiBuilder;
+
+    static INIT: Once = Once::new();
+    const REST_API_URL: &'static str = "http://127.0.0.1:5000";
```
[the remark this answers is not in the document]
RyanLassigBanks: Updated to bind to port 0

vaporos:
"Add unit tests for biome rest api" commit could have additional summary of what is being tested.

The comments on the tests should have a blurb which describes at a high-level what is being tested. Ideally it would also state the "why" for the test, though that is sometimes hard to define for happy path. For example:

```
    /// Happy path test for POST /biome/register
    ///
    /// When registering a user with valid inputs, we expect the request to succeed and for the
    /// response to contain the username of the user which was just created.
    ///
    /// Procedure
    ///
    /// 1) Register user via POST /biome/register
    /// 2) Verify that the request was successful
    /// 3) Verify that the username of the user created matches
```

PR #705, opened 2020-04-14
Additional splinterd config

ltseeley:
- Make network endpoints for splinter nodes a list instead of a single value
- Add `advertised_endpoints` configuration to splinterd
- Add `display_name` configuration to splinterd
- Update the response of splinterd's `GET /status` endpoint

### DIFF
--- a/cli/man/splinter-circuit-propose.1.md
+++ b/cli/man/splinter-circuit-propose.1.md
@@ -65,8 +65,8 @@ OPTIONS
 
 `--node NODE-STRING` ...
 : Specifies a node that should be part of the circuit, using the format
-  `NODE-ID::ENDPOINT`. This node ID must match the node ID and endpoint entry
-  in the node registry. The proposer must also specify its own node, if it is to
+  `NODE-ID::ENDPOINT1,ENDPOINT2`. All endpoints must be in the node's entry in
+  the node registry. The proposer must also specify its own node, if it is to
   be included on the circuit proposal. Repeat this option to specify multiple
   nodes.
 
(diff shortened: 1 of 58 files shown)


splinterd/src/daemon.rs
```diff
@@ -252,26 +255,28 @@ impl SplinterDaemon {
...
-                        )));
-                    }
-                };
-                debug!("Received connection from {}", connection.remote_endpoint());
-                match network_clone.add_connection(connection) {
-                    Ok(peer_id) => debug!("Added connection with ID {}", peer_id),
-                    Err(err) => error!("Failed to add connection to network: {}", err),
-                };
+            for mut network_listener in network_listeners {
+                for connection_result in network_listener.incoming() {
```
agunde406, line 261: Doesn't this block? so you are only going to be listening to one network listener?

libsplinter/src/transport/mod.rs
```diff
@@ -107,6 +107,10 @@ pub trait Transport: Send {
     fn accepts(&self, address: &str) -> bool;
     fn connect(&mut self, endpoint: &str) -> Result<Box<dyn Connection>, ConnectError>;
     fn listen(&mut self, bind: &str) -> Result<Box<dyn Listener>, ListenError>;
+
+    fn listen_all(&mut self, binds: &[String]) -> Result<Vec<Box<dyn Listener>>, ListenError> {
```
vaporos, line 111: This should not be added as the it will not properly correlate ListenError and is unnecessary.

cli/man/splinter-circuit-propose.1.md
```diff
@@ -123,14 +123,15 @@ EXAMPLES
 This command proposes a simple circuit with one other node.
 
 * The proposing node has ID `alpha001` and endpoint `tls://splinterd-node-acme001:8044`.
-* The other node has ID `beta001` and endpoint `tls://splinterd-node-beta001:8044`.
+* The other node has ID `beta001` and endpoints `tls://splinterd-node-beta001:8044`
+  and `wss://splinterd-node-beta001:8045`.
```
vaporos, line 127: Should use tcp:// and tcps:// as examples here since those are implemented and stable.

splinterd/src/daemon.rs
ltseeley: Updated so each listener has its own thread

libsplinter/src/transport/mod.rs
ltseeley: Removed

cli/man/splinter-circuit-propose.1.md
ltseeley: Updated

PR #708, opened 2020-04-16
Biome REST API fixes

RyanLassigBanks:
A handful of small fixes for the biome rest api

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/user.rs
+++ b/libsplinter/src/biome/rest_api/actix/user.rs
@@ -150,6 +150,7 @@ fn add_modify_user_method(
     secret_manager: Arc<dyn SecretManager>,
     key_store: Arc<dyn KeyStore>,
 ) -> HandlerFunction {
+    let ecryption_cost = rest_config.password_encryption_cost();
     Box::new(move |request, payload| {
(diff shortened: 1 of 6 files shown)


libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -153,6 +153,7 @@ fn add_modify_user_method(
     Box::new(move |request, payload| {
         let credentials_store = credentials_store.clone();
         let key_store = key_store.clone();
+        let rest_config = rest_config.clone();
```
peterschwarz, line 156: You could probaly just clone the encryption cost, instead of the whole configuration.  Additionally, you could make the encryption struct `Copy` and avoid the use of clone.

libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -153,6 +153,8 @@ fn add_modify_user_method(
     Box::new(move |request, payload| {
         let credentials_store = credentials_store.clone();
         let key_store = key_store.clone();
+        let ecryption_cost = rest_config.password_encryption_cost();
```
peterschwarz, line 156: This could move outside the closure

PR #709, opened 2020-04-16
Add versioning to config objects

shannynalayna:
Adds a required `version` field to all config objects and verifies that each config object added to the final `ConfigBuilder` object has the correct version. 

This also updates the gameroom example to conform with this change.

### DIFF
--- a/splinterd/sample_configs/splinterd-node-0-docker.toml
+++ b/splinterd/sample_configs/splinterd-node-0-docker.toml
@@ -13,6 +13,9 @@
 # limitations under the License.
 # ------------------------------------------------------------------------------
 
+# Config file format version
+version = "1"
+
 node_id = "012"
 
 # Endpoint used for service to daemon communication.
--- a/splinterd/src/config/error.rs
+++ b/splinterd/src/config/error.rs
@@ -25,6 +25,7 @@ pub enum ConfigError {
     TomlParseError(TomlError),
     InvalidArgument(clap::Error),
     MissingValue(String),
+    InvalidVersion(String),
 }
 
 impl From<TomlError> for ConfigError {
@@ -46,6 +47,7 @@ impl Error for ConfigError {
             ConfigError::TomlParseError(source) => Some(source),
             ConfigError::InvalidArgument(source) => Some(source),
             ConfigError::MissingValue(_) => None,
+            ConfigError::InvalidVersion(_) => None,
         }
     }
 }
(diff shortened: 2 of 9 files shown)


splinterd/sample_configs/splinterd-node-0-docker.toml
```diff
@@ -39,3 +39,6 @@ bind = "0.0.0.0:8080"
 ca_certs = "/etc/splinter/certs/generated_ca.pem"
 # db address
 database = "postgres://splinter_admin:splinter_test@127.0.0.1:5432/splinter"
+
+# Toml config version
```
vaporos, line 43: Should put this as the first entry in the file (because its a common placement) and change the comment to "Config file format version"

splinterd/src/config/error.rs
```diff
@@ -59,6 +61,9 @@ impl fmt::Display for ConfigError {
                 write!(f, "Unable to parse command line argument: {}", source)
             }
             ConfigError::MissingValue(msg) => write!(f, "Configuration value must be set: {}", msg),
+            ConfigError::InvalidVersion(msg) => {
+                write!(f, "Config object has invalid version: {}", msg)
```
vaporos, line 65: This should probably pass through with no additional message text.

splinterd/src/config/toml.rs
```diff
@@ -67,6 +70,20 @@ impl PartialConfigBuilder for TomlPartialConfigBuilder {
             },
         };
 
+        if let Some(version) = self.toml_config.version {
+            debug!("Toml config has version {}", version);
+            if version == TOML_VERSION {
+                ()
+            } else {
+                return Err(ConfigError::InvalidVersion(format!(
+                    "{:?}: {} must be {}",
```
vaporos, line 79: Make this the entire message (see earlier pass-through comment). This should allow the message to be less generic (than prefixing more generic text).

examples/gameroom/docker-compose.yaml
```diff
@@ -246,6 +246,8 @@ services:
         - ./splinterd-config:/configs
         - ./node_registry:/node_registry
         - bubba-var:/var/lib/splinter
+      environment:
```
vaporos, line 249: This should be unnecessary here.

splinterd/sample_configs/splinterd-node-0-docker.toml
shannynalayna: Addressed

splinterd/src/config/toml.rs
shannynalayna: Addressed, in conjunction with that earlier comment

examples/gameroom/docker-compose.yaml
shannynalayna: Yep, missed this one earlier 👍

PR #713, opened 2020-04-17
Remove user from credential store when deleting them

RyanLassigBanks:
`DELETE /biome/users/{id}` now deletes users from the credentials
store in addition to removing them from the user store. Without doing
this, the user's data is still accessible via `GET /biome/users`.

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/user.rs
+++ b/libsplinter/src/biome/rest_api/actix/user.rs
@@ -83,10 +83,13 @@ pub fn make_user_routes(
                 key_store,
             ),
         )
-        .add_method(Method::Get, add_fetch_user_method(credentials_store))
+        .add_method(
+            Method::Get,
+            add_fetch_user_method(credentials_store.clone()),
+        )
         .add_method(
             Method::Delete,
-            add_delete_user_method(rest_config, secret_manager, user_store),
+            add_delete_user_method(rest_config, secret_manager, user_store, credentials_store),
         )
 }
 
@@ -292,6 +295,7 @@ fn add_delete_user_method(
     rest_config: Arc<BiomeRestConfig>,
     secret_manager: Arc<dyn SecretManager>,
     user_store: Arc<dyn UserStore>,
+    credentials_store: Arc<dyn CredentialsStore>,
 ) -> HandlerFunction {
     Box::new(move |request, _| {
         let user_store = user_store.clone();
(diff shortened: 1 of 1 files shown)


libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -314,27 +318,54 @@ fn add_delete_user_method(
...
+                debug!("User not found: {}", msg);
+                Box::new(
+                    HttpResponse::NotFound()
+                        .json(ErrorResponse::not_found(&format!(
+                            "User ID not found: {}",
+                            &user_id
+                        )))
+                        .into_future(),
+                )
+            }
```
ltseeley, line 360: Is it a valid state for a user to exist in the user store but not the credential store?

RyanLassigBanks: Currently, the credentials feature needs to be enabled for the register and user endpoints to be created, so it shouldn't be a valid state. Are you suggesting this should be an internal error instead?

ltseeley: Maybe; if biome should always have a corresponding entry in the credentials store for each user in the users store, and the REST API actions don't let the consumer mess that up, then it would be internal error (something's wrong with the code or the stores).

RyanLassigBanks: I agree with that. Updated that block to only return internal server errors in error case

libsplinter/src/biome/rest_api/actix/user.rs
```diff
@@ -314,27 +318,43 @@ fn add_delete_user_method(
...
+        };
+
+        match credentials_store.remove_credentials(&user_id) {
+            Ok(()) => Box::new(
+                HttpResponse::Ok()
+                    .json(json!({ "message": "User deleted sucessfully" }))
+                    .into_future(),
+            ),
+            Err(err) => {
+                error!("Failed to delete user in database {}", err);
```
ltseeley, line 351: This error message should probably be different than the one for the users store so they're easily distinguishable in the logs.

RyanLassigBanks: Changed to "Failed to delete credentials in database"

RyanLassigBanks:
The credentials table is being updated via `CASCADE delete` in the sql definition of the tables. Closing.

PR #714, opened 2020-04-17
Fix missed column rename in gameroom migration

ltseeley:
Updates the gameroom database migration for changing member `endpoint`
column to `endpoints` to actually change the name of the column; this
was overlooked in the original PR.


rbuysse:
Tested locally by upgrading nodes through all versions from 0.3.11 to master+this PR in docker.

PR #717, opened 2020-04-20
New splinterd man page (now with actual contents!)

chenette:

### DIFF
--- a/splinterd/man/splinterd.1.md
+++ b/splinterd/man/splinterd.1.md
@@ -1,39 +1,80 @@
-% COMMAND(1) Cargill, Incorporated | Splinter Commands
+% SPLINTERD(1) Cargill, Incorporated | Splinter Commands
 
-<!--
-Template for a man page source file in Markdown format. The file should have the
-name of the command plus any subcommands (with dashes) with an extension of
-`1.md`. For example `splinter-cert-generate.1.md` is the Markdown file that
-generates the man page for `splinter cert generate`.
--->
 NAME
 ====
 
-**command-subcommand** — About blurb from command help
+**splinterd** — Starts the Splinter daemon
 
 SYNOPSIS
 ========
-| **command** \[**FLAGS**\] \[**OPTIONS**\]
+
+**splinterd** \[**FLAGS**\] \[**OPTIONS**\] <arguments>
 
 DESCRIPTION
 ===========
-Paragraph or two that describes what the command does, explains how it fits in
-the Splinter environment, provides big-picture information on who uses it and
-why (or when), and lists prerequisites or other considerations.
+
+This command configures and runs the Splinter daemon, `splinterd`. This daemon
+provides core Splinter functionality such as circuit management, a REST API,
+and an admin service. It can also include application-specific services.
+
+The `splinterd` command requires a new ID for the node. Other settings are
+optional.
+
+**Configuration File**
+
+This command provides the `-c` and `--config` options to specify an optional
+TOML configuration file containing `splinterd` settings, instead of using
+command-line options. (Any options on the command line will override those in
+the configuration file.) The file name must end with a `.toml` extension, such
+as `splinterd.toml`). For examples, see
+[splinterd/packaging/splinterd.toml.example](https://github.com/Cargill/splinter/blob/master/splinterd/packaging/splinterd.toml.example)
+and
+[splinterd/sample_configs](https://github.com/Cargill/splinter/tree/master/splinterd/sample_configs)
+in the `splinter` repository.
+
+**Connection Types**
+
+The Splinter daemon supports transport-level connections with raw (TCP),
+Transport Layer Security (TLS), and WebSocket protocols, which are also called
+"transport types". A node can use multiple transport types at once. By default,
+`splinterd` configures all transport types. If the `--no-tls` flag is set, no
+TLS connections are configured.
+
+
+TLS requires certificate authority (CA) certificates and keys, which are stored
+in `/etc/splinter/certs/` by default. If necessary, you can change the
+directory, specify individual file paths and names, or use self-signed
+(insecure) certificates in a development environment. For more information, see
+the certificate-related options and CERTIFICATE FILES, below.
 
 FLAGS
 =====
-<!--
-List of flags (in alphabetical order) that describes what each flag does
-and the default value/behavior. If appropriate, explain
-how/when it's used, with tips, gotchas, and very short examples.
--->
+
+`--enable-biome`
+: Enable the Biome subsystem, which provides user management functions for
+  Splinter applications. The `--database` option is required when this flag is
+  used.
 
 `-h`, `--help`
-: Prints help information
+: Prints help information.
+
+`--insecure`
+: Turns off certificate authority validation for TLS connections; all peer
+  certificates are accepted. This flag is intended for development environments
+  using self-signed certificates.
+
+`--no-tls`
+: Turns off TLS configuration and restricts `splinterd` to TCP (`raw`)
+  connections. This flag allows `splinterd` to start without the certificates
+  and keys that TLS requires. Without `--no-tls`, if `splinterd` cannot find the
+  certificates and keys required by TLS, it exits with an error.
+
+`-q`, `--quiet`
+: Decreases verbosity (the opposite of `-v`). When specified, only errors or
+  warnings will be displayed.
 
 `-V`, `--version`
-: Prints version information
+: Prints version information.
 
 `-v`
 : Increases verbosity (the opposite of -q). Specify multiple times for more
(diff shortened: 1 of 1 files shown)


splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,278 @@
...
 ===========
-Paragraph or two that describes what the command does, explains how it fits in
-the Splinter environment, provides big-picture information on who uses it and
-why (or when), and lists prerequisites or other considerations.
+
+This command configures and runs the Splinter daemon, `splinterd`. This daemon
+provides core Splinter functionality such as circuit management, a REST API,
+and an admin service. It can also include application-specific services.
+
+The `splinterd` command requires a new unique ID for the node. Other settings
```
chenette, line 20: How "unique" must the node ID be?  Across all possible node IDs on the network (aka The Internet)?  Just for the local Splinter node?  Across all current and future circuits?  

Also, what happens if the node ID is NOT unique?

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,278 @@
...
+  (Default: 30000 ms (30 seconds).)
+
+  This setting affects consensus-related activities for pending circuit changes
+  (functions that use the two-phase commit agreement protocol in the Scabbard
+  service).
+
+`--advertised-endpoint` `ADVERTISED-ENDPOINT`
+: Specifies the public network endpoint for daemon-to-daemon communication
+  between Splinter nodes, if the network endpoint is not public.
+  (Default: The network endpoint; see `-n`, `--network-endpoint.)
```
RyanLassigBanks, line 80: Missing closing "`"

chenette: Fixed.

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,278 @@
...
+see CERTIFICATE FILES, below.
 
 FLAGS
 =====
-<!--
-List of flags (in alphabetical order) that describes what each flag does
-and the default value/behavior. If appropriate, explain
-how/when it's used, with tips, gotchas, and very short examples.
--->
+
```
agunde406, line 40: Please add ` --no-tls      Turn off tls configuration`.  If this command is not provided and the cert and keys for tls are not available splinterd will error. This will limit splinterd to only running over tcp.

chenette:
This difference has been noted and might be addressed in a later PR.

splinterd/man/splinterd.1.md
agunde406: Unique among all nodes on a network. If using trust auth, two nodes could become confused for each other. (Hence why trust is insecure)

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,278 @@
...
+`--enable-biome`
+: (Feature: "biome" and "database") Enable the Biome subsystem, which provides
```
vaporos, line 42: Calling out the compilation features like this isn't going to make sense to sysadmins that are not intimately knowledgeable of our development practices.

chenette: @vaporos  -  Is the `--whitelist` version better?  It has a separate sentence that says "This option is available when CORS is enabled with the `rest-api-cores` feature."

splinterd/man/splinterd.1.md
eloaverona: What is a network? I thought there was no concept of a "splinter network"?

agunde406: I am using network to mean every possible splinterd you can connect to from your splinterd. It is not a "real concept". Im sorry for the confusion.

chenette:
Force-pushed an update with several corrections and wording improvements.

Summary:

- Add "subheaders" in DESCRIPTION for config file and connection type info.

- Update the transport-related info in DESCRIPTION to mention `--no-tls` and WebSocket.

- Add `--no-tls` under FLAGS.

- Change the marker for flags/options that require a specific feature: `--enable-biome`, `--database`, and `--whitelist`. Now, the description starts with "(Optional feature)" and ends with "Note: This flag/option is available when the XXX feature is enabled."

- Remove incorrect "node ID format" info (a node ID has no required format).

- Fix examples:  The simplest case requires only a node ID (the endpoints have defaults). As a result, the config file example needed to change. It's now an example of using a config file to set the node ID and change the heartbeat interval (a "harmless" setting, I hope).

- Make the --{*}endpoint descriptions consistent . 

- Edit the --whitelist description.

- Fix font problems throughout (add backticks).

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,298 @@
...
+in `/etc/splinter/certs/` by default. If necessary, you can change the
+directory, specify individual file paths and names, or use insecure self-signed
+certificates in a development environment. For more information, see CERTIFICATE
+FILES, below.
+
+A node can use multiple transport types at once. By default, `splinterd`
+configures all transport types. The following flags restrict the connection
+type:
+
+* If `--insecure` is set (when using self-signed certificates), only TLS
```
agunde406, line 51: This is not true, it just means that TLS is not secure (aka does not verify the certificates) It can still run TLS, TCP and websockets.

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,298 @@
...
+  connections are configured.
+
+* If `--no-tls` is set, no TLS connection are configured; `splinterd` is
+  restricted to TCP connections only.
```
agunde406, line 55: Might be better to word this as in what it takes away, instead of what is still avaialbe so it doesnt need to be updated when websockets are on by default.

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,298 @@
...
+`--cert-dir CERT-DIR`
+: Specifies the directory that contains the trusted CA certificates and
+  associated key files. (Default: `/etc/splinter/certs/`.)
+
+  This option overrides the `SPLINTER_CERT_DIR` environment variable, if set,
+  and any setting in the `splinterd` configuration file.
+
+`--client-cert CERT-FILE`
+: Specifies the path and file name for the client certificate, which is
+  used by `splinterd` when it is sending messages.
```
agunde406, line 126: this should specifiy "sending messages over TLS"

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,298 @@
...
+  nodes, using the format `tcp://ip:port`. (Default: 127.0.0.1:8044.)
+
+  `--node-id NODE-ID`
+: (Required) Sets a new ID for the node. The node ID must be unique across the
+  network (for all Splinter nodes that could participate on the same circuit).
+
+`--peer PEER-URL` `[,...]`
+: Specifies one or more Splinter nodes that `splinterd` will automatically
+  connect to when it starts. The *PEER-URL* argument must specify another node's
+  network endpoint, using the format `ip:port`.
```
agunde406, line 170: technically this should now include the protocol prefix `protocol_prefix://ip:port` as well, ill add a story to fix this

splinterd/man/splinterd.1.md
```diff
@@ -1,81 +1,298 @@
...
+
+  Specify multiple nodes in a comma-separated list or by repeating the
+  `--peer` option.
+
+`--registry REGISTRY-FILE` `[,...]`
+: Specifies one or more read-only node registry files.
+
+`--server-cert SERVER-CERT`
+: Specifies the path and file name for the server certificate, which is used by
+  `splinterd` when it is receiving messages.
```
agunde406, line 180: Again this should clarify over TLS

chenette:
Force-pushed a commit with review corrections:

- Correct the `--no-tls` flag information in the DESCRIPTION and move the `--insecure` info to the flag description

- Remove "(Optional feature)" and feature names from `--enable-biome`, `--database`, and `--whitelist`

- Clarify "when sending messages over TLS" for `--client-cert` and `--server-cert`

- In DESCRIPTION, change "new unique ID for the node" to "new ID for the node", because the option's description does a better job of explaining what "unique" means

splinterd/man/splinterd.1.md
chenette: Addressed - removed "unique" from this phrase, and expanded the `--node-id` description to clarify what "unique" means.

splinterd/man/splinterd.1.md
chenette: Removed all markers for optional features, because I learned that they'll be enabled by default.

splinterd/man/splinterd.1.md
chenette: Reworded to correct and clarify (I hope).

splinterd/man/splinterd.1.md
chenette: Fixed.

splinterd/man/splinterd.1.md
chenette: Done.

splinterd/man/splinterd.1.md
chenette: Done.

PR #719, opened 2020-04-20
Authorize connections within the connection manager

peterschwarz:
This PR introduces authorization at the connection level, instead of at the peer level.  This ensures that any connection requested or added to the connection manager is authorized before the connection is usable by the layers above (like the peer manager or the service connection manager).

It copies (for the time being) the existing authorization code from network and reworks the the authorization manager into pool that owns the connections until they have finished the authorization process. Once done, it returns the connection to the connection manager, so it can proceed as usual.

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -37,6 +37,41 @@ const DEFAULT_HEARTBEAT_INTERVAL: u64 = 10;
 const INITIAL_RETRY_FREQUENCY: u64 = 10;
 const DEFAULT_MAXIMUM_RETRY_FREQUENCY: u64 = 300;
 
+pub type AuthorizerCallback =
+    Box<dyn Fn(AuthorizationResult) -> Result<(), Box<dyn std::error::Error>> + Send>;
+
+pub trait Authorizer {
+    fn authorize_connection(
+        &self,
+        connection_id: String,
+        connection: Box<dyn Connection>,
+        on_complete: AuthorizerCallback,
+    ) -> Result<(), AuthorizerError>;
+}
+
+pub enum AuthorizationResult {
+    Authorized {
+        connection_id: String,
+        identity: String,
+        connection: Box<dyn Connection>,
+    },
+    Unauthorized {
+        connection_id: String,
+        connection: Box<dyn Connection>,
+    },
+}
+
+#[derive(Debug)]
+pub struct AuthorizerError(pub String);
+
+impl std::error::Error for AuthorizerError {}
+
+impl std::fmt::Display for AuthorizerError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
 pub type SubscriberId = usize;
 type Subscriber =
     Box<dyn Fn(ConnectionManagerNotification) -> Result<(), Box<dyn std::error::Error>> + Send>;
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -16,7 +16,6 @@ mod error;
 mod notification;
 mod pacemaker;
 
-use std;
 use std::cmp::min;
 use std::collections::HashMap;
 use std::sync::mpsc::{channel, Sender};
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -119,6 +119,7 @@ impl SubscriberMap {
 enum CmMessage {
     Shutdown,
     Request(CmRequest),
+    AuthResult(AuthResult),
     SendHeartbeats,
 }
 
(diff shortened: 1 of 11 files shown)


libsplinter/src/network/connection_manager/auth/handlers.rs
```diff
@@ -0,0 +1,626 @@
...
+    use std::collections::VecDeque;
+    use std::sync::{Arc, Mutex};
+
+    use protobuf::Message;
+
+    use crate::protos::authorization;
+    use crate::protos::network::{NetworkMessage, NetworkMessageType};
+
+    #[test]
+    fn connect_request_dispatch() {
```
vaporos, line 420: Tests should have a) a short description; b) a description of the purpose of the test; c) the procedure to achieve (b).

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -529,57 +569,217 @@ where
...
+        reply_sender: Sender<Result<(), ConnectionManagerError>>,
+        internal_sender: Sender<CmMessage>,
+        auth_pool: &AuthorizationPool,
+        subscribers: &mut SubscriberMap,
+    ) {
         let endpoint = connection.remote_endpoint();
-
         let id = Uuid::new_v4().to_string();
 
+        if endpoint.contains("inproc") {
```
vaporos, line 580: Why is inproc special?

libsplinter/src/network/connection_manager/auth/mod.rs
```diff
@@ -11,3 +11,5 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+pub mod protocol;
```
vaporos, line 17: This should probably go in crate::protocol and follow Transact's lead.

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: Inproc was previously special, in that it didn't require auth (it's compiled in, after all).  This has just shifted to let the connection manager decide, versus the auth layer, which should have to be concerned about what type of connection it is.

libsplinter/src/network/connection_manager/auth/handlers.rs
peterschwarz: As this was taken from the previous code, it did not originally have a comment, but will add one

peterschwarz: Fixed

vaporos: This comment was intended to apply to all the tests in the PR.

libsplinter/src/network/connection_manager/auth/pool.rs
```diff
@@ -0,0 +1,196 @@
...
+    //     self
+    // }
+
+    pub fn with_prefix(mut self, prefix: String) -> Self {
+        self.prefix = Some(prefix);
+        self
+    }
+
+    pub fn build(self) -> Result<ThreadPool, ThreadPoolBuildError> {
+        let size = self.size.unwrap_or_else(|| num_cpus::get());
```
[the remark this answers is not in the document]
peterschwarz: I've added back the below commented code, and utilize this aspect of the builder at the callsite.

libsplinter/src/network/connection_manager/auth/pool.rs
```diff
@@ -0,0 +1,196 @@
...
+    size: Option<usize>,
+    prefix: Option<String>,
+}
+
+impl ThreadPoolBuilder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    // pub fn with_size(mut self, size: usize) -> Self {
```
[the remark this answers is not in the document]
peterschwarz: Fixed.

libsplinter/src/network/connection_manager/auth/mod.rs
peterschwarz: Moved.

libsplinter/src/network/connection_manager/auth/handlers.rs
peterschwarz: These were the original comments, but I've updated them all to be more verbose. I probably wasn't clear in the commit message that this is the same code from the existing network auth code moved here.

libsplinter/src/network/connection_manager/auth/handlers.rs
```diff
@@ -0,0 +1,626 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use protobuf::Message;
+
+use crate::network::dispatch::{
```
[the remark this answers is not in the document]
peterschwarz: I agree, but I think, in order to have a running system, we should make this change in a future PR (possible as part of, or after integrating it into splinter as a whole).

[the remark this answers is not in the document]
peterschwarz: Updated as part of this PR.  The connection manager takes an `Authorizer` trait

libsplinter/src/network/connection_manager/mod.rs
peterschwarz: It probably would be better to add a feature where a connection could be indicate that it is authorizable, but that would require some API changes at the transport level.  Not hard - the default implementation of the function could return true.

PR #722, opened 2020-04-20
Support ./ and ../ in config file paths

shannynalayna:
Adds support to './' and '../' in file paths for files necessary
to the splinterd config. This ensures that these file paths will
not be prepended with the default certificate directory in case
'./' or '../' are present in the file path.

### DIFF
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -24,12 +24,15 @@ pub trait PartialConfigBuilder {
 }
 
 fn get_file_path(cert_dir: &str, file: &str) -> String {
-    let cert_dir_path = Path::new(&cert_dir);
-    let cert_file_path = cert_dir_path.join(file);
-    cert_file_path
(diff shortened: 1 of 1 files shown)


splinterd/src/config/builder.rs
```diff
@@ -24,12 +24,15 @@ pub trait PartialConfigBuilder {
 }
 
 fn get_file_path(cert_dir: &str, file: &str) -> String {
-    let cert_dir_path = Path::new(&cert_dir);
-    let cert_file_path = cert_dir_path.join(file);
-    cert_file_path
-        .to_str()
-        .map(ToOwned::to_owned)
-        .unwrap_or_else(|| String::from(file))
+    if file.contains("./") || file.contains("../") {
```
vaporos, line 27: Should this be startswith()?

shannynalayna: Oh, yeah, updated!

PR #725, opened 2020-04-21
Move rest-api-cors into stable

shannynalayna:


peterschwarz:
This should wait for #728

PR #726, opened 2020-04-21
Remove keygen feature from splinter cli

RyanLassigBanks:
Stabilize keygen feature by removing feature and related feature guards.
`splinter keygen` will now be available when the cli is compiled with
no feature flags.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -33,7 +33,7 @@ use flexi_logger::FlexiLoggerError;
 use flexi_logger::{DeferredNow, LogSpecBuilder, Logger};
 use log::Record;
 
-use action::{admin, certs, circuit, Action, SubcommandActions};
+use action::{admin, certs, circuit, keygen, Action, SubcommandActions};
 use error::CliError;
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -109,62 +109,63 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
-        SubCommand::with_name("cert")
-            .about("Generates certificates that can be used for development")
-            .setting(AppSettings::SubcommandRequiredElseHelp)
-            .subcommand(
-                SubCommand::with_name("generate")
-                    .long_about(
-                        "Generates test certificates and keys for running splinterd with \
+    app = app
+        .subcommand(
+            SubCommand::with_name("cert")
```
peterschwarz, line 114: This reformatting is odd - did that happen automatically?

RyanLassigBanks: Yeah that's what `cargo fmt` gave me.

PR #731, opened 2020-04-22
Stabilize registry-remote feature

ltseeley:
Stabilizes the `registry-remote` feature in libsplinter, and removes the
corresponding feature in splinterd in favor of always using the feature.

NOTE: the remote registry is not fully tested yet; tests will be added as part of "stabilizing" the entire node registry.


peterschwarz:
Mobile GitHub did not submit my line comment: I think this should still be a feature in libsplinter. As a library user, I may wish to not use this type of registry

PR #734, opened 2020-04-22
Switch two phase coordinator timeout to seconds

agunde406:
This is done in an effort for consistency with other
splinterd options. Heartbeat takes seconds while
coordinator timeout used to take milliseconds.
Now both will take seconds.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -55,7 +55,7 @@ use self::shared::AdminServiceShared;
 pub use self::error::AdminServiceError;
 pub use self::error::AdminSubscriberError;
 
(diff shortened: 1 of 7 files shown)


libsplinter/src/admin/service/mod.rs
```diff
@@ -55,7 +55,7 @@ use self::shared::AdminServiceShared;
 pub use self::error::AdminServiceError;
 pub use self::error::AdminSubscriberError;
 
-const DEFAULT_COORDINATOR_TIMEOUT_MILLIS: u64 = 30000; // 30 seconds
+const DEFAULT_COORDINATOR_TIMEOUT_MILLIS: u64 = 30; // 30 seconds
```
vaporos, line 58: s/MILLIS//

PR #739, opened 2020-04-22
splinterd man page fixes: No -q; multi. endpoints

chenette:

- Remove -q flag (splinterd doesn't have one) and update -v blurb

- Clarify that you can specify multiple endpoints for --advertised-endpoint and --network-endpoint

### DIFF
--- a/splinterd/man/splinterd.1.md
+++ b/splinterd/man/splinterd.1.md
@@ -64,10 +64,6 @@ FLAGS
   and keys that TLS requires. Without `--no-tls`, if `splinterd` cannot find the
   certificates and keys required by TLS, it exits with an error.
 
-`-q`, `--quiet`
-: Decreases verbosity (the opposite of `-v`). When specified, only errors or
-  warnings will be displayed.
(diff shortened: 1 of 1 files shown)


splinterd/man/splinterd.1.md
```diff
@@ -149,6 +147,9 @@ OPTIONS
 : Specifies the endpoint for daemon-to-daemon communication between Splinter
   nodes, using the format `tcp://ip:port`. (Default: 127.0.0.1:8044.)
```
agunde406, line 148: `tcp://` should be changed to `protocol_prefix://`. we should also explain how tcps:// creates a TLS connection and tcp:// connection a TCP connection. This is important because if you want to accept both types you need to provide an endpoint for each (to create a listener for)

chenette: I'd like to make this change in a separate PR.  Is that okay?

agunde406: yes

PR #741, opened 2020-04-23
Add SPLINTER_HOME env variable

shannynalayna:
Adds a SPLINTER_HOME environment variable, to the EnvConfig object, as well as the CLI. This will check if the SPLINTER_HOME env variable is set, and if it is, will be appended to a default directory value. If not, the default directory value is used. 

In terms of the config objects, the `splinter_home` value is not added to the final config object, but evaluated while the final config object is being built.


splinterd/src/config/builder.rs
```diff
@@ -35,6 +35,21 @@ fn get_file_path(cert_dir: &str, file: &str) -> String {
     }
 }
 
+fn build_with_splinter_home_dir(
```
vaporos, line 38: We should be able to make this change entirely within the env partial config builder, without changes to anywhere else in the config system.

vaporos:
Should use the same directory structure as SAWTOOTH_HOME uses for consistency.

splinterd/src/config/builder.rs
shannynalayna: So, you're thinking that in the case that the splinter_home is set in the env partial config builder, the other values (if unset) are set to "splinter_home + default_directory_path"?

shannynalayna: I updated it assuming that my understanding outlined in that comment, was the desired approach.

chenette:
Should this PR include adding (and describing) the SPLINTER_HOME environment variable on the splinterd(1) man page and applicable splinter-*(1) man pages?

cli/src/action/admin.rs
```diff
@@ -34,8 +34,9 @@ use crate::error::CliError;
 
 use super::{chown, Action};
 
-const DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
+const DEFAULT_STATE_DIR: &str = "/var/lib/splinter";
```
vaporos, line 37: This change look unnecessary.

cli/src/action/admin.rs
```diff
@@ -34,8 +34,9 @@ use crate::error::CliError;
...
 const STATE_DIR_ENV: &str = "SPLINTER_STATE_DIR";
+const SPLINTER_HOME: &str = "SPLINTER_HOME";
```
vaporos, line 39: SPLINTER_HOME_ENV for consistency.

cli/src/action/certs.rs
```diff
@@ -63,7 +64,16 @@ impl Action for CertGenAction {
             .value_of("cert_dir")
             .map(ToOwned::to_owned)
             .or_else(|| env::var(CERT_DIR_ENV).ok())
-            .or_else(|| Some(DEFAULT_CERT_DIR.to_string()))
+            .or_else(|| {
+                if let Ok(splinter_home) = env::var(SPLINTER_HOME) {
+                    Path::new(&splinter_home)
```
vaporos, line 69: This approach doesn't match the one in splinterd.

cli/src/action/admin.rs
```diff
@@ -81,7 +82,16 @@ impl Action for KeyRegistryGenerationAction {
             .value_of("target_dir")
             .map(ToOwned::to_owned)
             .or_else(|| env::var(STATE_DIR_ENV).ok())
-            .or_else(|| Some(DEFAULT_STATE_DIR.to_string()))
+            .or_else(|| {
+                if let Ok(splinter_home) = env::var(SPLINTER_HOME) {
+                    Path::new(&splinter_home)
```
vaporos, line 87: This approach doesn't match the one in splinterd.

splinterd/man/splinterd.1.md
```diff
@@ -111,10 +119,10 @@ OPTIONS
   configuration file.
 
 `--config-dir CONFIG-DIR`
-: Specifies the directory containing configuration files.
-  (Default: `/etc/splinter`.)
-
-  This option overrides the `SPLINTER_CONFIG_DIR` environment variable, if set.
+: Specifies the directory containing Splinter configuration files. (Default:
+  `/etc/splinter`, unless `SPLINTER_CONFIG_DIR` or `SPLINTER_HOME` is set.)
+  This option overrides any value set by the environment variables or
```
vaporos, line 124: Omit this statement, as it is always true of all CLI options. (Same for instances that follow.)

splinterd/man/splinterd.1.md
```diff
@@ -237,29 +245,83 @@ self-signed certificates and keys (which can be generated by the
...
+
+* Splinter configuration directory: `/etc/splinter`
+
+* State directory: `/var/lib/splinter/`
+
+* TLS certificate directory: `/etc/splinter/certs/`
+
+For the configuration and certificate directories, the directory paths can be
+changed individually with a `splinterd` option, a setting in a TOML config file,
+or an environment variable. (The state directory location is controlled only by
```
vaporos, line 261: The state directory config/option should be added.

vaporos:
When SPLINTER_HOME is set, the appropriate sub-directories should automatically be created.

I got this error when testing this:

[2020-05-04 10:46:36.128] T["main"] ERROR [splinterd] Failed to start daemon, unable to start the Splinter daemon: the admin service encountered an error: unable to create admin service: service was unable to be created: Unable to write to persisted storage: File write failed: No such file or directory (os error 2): /var/lib/splinter//circuit_proposals.yaml

I also ran into:

[2020-05-04 10:45:03.282] T["main"] ERROR [splinterd] Failed to start daemon, unable to get transport: unable to retrieve certificate: Must provide a valid file containing ca certs: /Users/amundson/projects/review/splinter-pr-741/home/certs/ca.pem

Because the file is generated_ca.pem. We should probably create a symlink ca.pem->generated_ca.pem or find another solution to this issue so it doesn't require a manual step.

splinterd/man/splinterd.1.md
shannynalayna: Updated

splinterd/man/splinterd.1.md
shannynalayna: Will make a story as part of stabilization clean up

PR #744, opened 2020-04-23
Add newline to end of local YAML registry file

ltseeley:
Updates the `LocalYamlNodeRegistry` implementation to append a newline
to the end of its backing file. This allows the file to be edited with
standard text editors.

### DIFF
--- a/libsplinter/src/node_registry/yaml/local.rs
+++ b/libsplinter/src/node_registry/yaml/local.rs
@@ -21,6 +21,7 @@
 //! [`RwNodeRegistry`]: ../../trait.RwNodeRegistry.html
 
 use std::fs::File;
+use std::io::Write;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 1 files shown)


libsplinter/src/node_registry/yaml/local.rs
```diff
@@ -135,7 +136,17 @@ impl LocalYamlNodeRegistry {
...
+        let mut file = File::create(&file_backend.file_path).map_err(|err| {
+            NodeRegistryError::general_error_with_source(
+                &format!(
+                    "Failed to open YAML registry file '{}'",
+                    file_backend.file_path
+                ),
+                Box::new(err),
+            )
+        })?;
+        file.write_all(&output).map_err(|err| {
```
peterschwarz, line 149: Since you're cleaning this up a bit, you might use `serde_yaml::to_writer` to write to the file you created, instead of creating the bytes.

ltseeley: Unfortunately, `serde_yaml::to_writer` consumes the `Write` and `File` isn't easily cloneable, so that isn't a clean replacement.

PR #747, opened 2020-04-24
Add config_dir option to config objects

shannynalayna:
Adds a `config_dir` option to the DefaultPartialConfigBuilder, EnvPartialConfigBuilder, and the ClapPartialConfigBuilder. The `config_dir`, default `/etc/splinter`, is appended to the beginning of the `tls_cert_dir` if the `tls_cert_dir` does not begin with "/" (in which case we assume this is the full path)

This pr also adds another step to building the config file paths to check if the file starts with "/", in which case we assume this is the full path. If it does not start with "/", the default / set tls cert directory will be appended to the beginning of the file. 

This also updates the tests to reflect these changes.

### DIFF
--- a/splinterd/man/splinterd.1.md
+++ b/splinterd/man/splinterd.1.md
@@ -112,6 +112,12 @@ OPTIONS
   Any options on the command line will override the settings in the
   configuration file.
 
+`--config-dir CONFIG-DIR`
+: Specifies the directory containing configuration files.
+  (Default: `/etc/splinter`.)
+
+  This option overrides the `SPLINTER_CONFIG_DIR` environment variable, if set.
+
 `--display-name DISPLAY-NAME`
 : Specifies a human-readable name for the node (Default: "Node NODE-ID")
 
@@ -241,6 +247,10 @@ ENVIRONMENT VARIABLES
 : Specifies where to store the circuit state YAML file, if `--storage` is
   set to `yaml`. (See `--storage`.) By default, this file is stored in
   `/var/lib/splinter`.
+
+**SPLINTER_CONFIG_DIR**
+: Specifies the directory containing configuration files.
+  (See: `--config-dir`.)
 
 FILES
 =====
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -63,6 +63,14 @@ impl ConfigBuilder {
     /// Builds a Config object by incorporating the values from each PartialConfig object.
     ///
     pub fn build(self) -> Result<Config, ConfigError> {
+        let config_dir = self
+            .partial_configs
+            .iter()
+            .find_map(|p| match p.config_dir() {
+                Some(v) => Some((v, p.source())),
+                None => None,
+            })
+            .ok_or_else(|| ConfigError::MissingValue("config directory".to_string()))?;
         let tls_cert_dir = self
             .partial_configs
             .iter()
@@ -160,6 +168,7 @@ impl ConfigBuilder {
         // Iterates over the list of PartialConfig objects to find the first config with a value
         // for the specific field. If no value is found, an error is returned.
         Ok(Config {
+            config_dir,
             storage: self
                 .partial_configs
                 .iter()
@@ -324,11 +333,11 @@ mod tests {
 
     /// Example configuration values.
     static EXAMPLE_STORAGE: &str = "yaml";
-    static EXAMPLE_CA_CERTS: &str = "certs/ca.pem";
-    static EXAMPLE_CLIENT_CERT: &str = "certs/client.crt";
-    static EXAMPLE_CLIENT_KEY: &str = "certs/client.key";
-    static EXAMPLE_SERVER_CERT: &str = "certs/server.crt";
-    static EXAMPLE_SERVER_KEY: &str = "certs/server.key";
+    static EXAMPLE_CA_CERTS: &str = "/etc/splinter/certs/ca.pem";
+    static EXAMPLE_CLIENT_CERT: &str = "/etc/splinter/certs/client.crt";
+    static EXAMPLE_CLIENT_KEY: &str = "/etc/splinter/certs/client.key";
+    static EXAMPLE_SERVER_CERT: &str = "/etc/splinter/certs/server.crt";
+    static EXAMPLE_SERVER_KEY: &str = "/etc/splinter/certs/server.key";
     static EXAMPLE_SERVICE_ENDPOINT: &str = "127.0.0.1:8043";
     static EXAMPLE_NETWORK_ENDPOINT: &str = "127.0.0.1:8044";
     static EXAMPLE_ADVERTISED_ENDPOINT: &str = "localhost:8044";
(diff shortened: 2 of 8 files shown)


splinterd/src/config/mod.rs
```diff
@@ -425,6 +425,24 @@ impl Config {
...
+            self.tls_cert_dir(),
+            self.tls_cert_dir_source()
+        );
+        debug!(
+            "Config: state_dir: {} (source: {:?})",
+            self.state_dir(),
+            self.state_dir_source()
+        );
+    }
 }
```
ltseeley, line 446: Seems unnecessary to split this out from the main function

shannynalayna: There was a clippy error when I added the whitelist debug statement about the complexity of the function being too high. This was an attempt to not run into that clippy error again.

ltseeley: Hmm, this might actually be a case where ignoring the lint is the right thing to do. The cognitive complexity of the `log_as_debug` function is actually very low since it only does one thing; it just does it many times. IMO splitting parts of it out just makes it more complex, because you have to look at the various sub-functions to get the full picture.

shannynalayna: Mmkay, ya I would agree. I'll fix that

splinterd/src/config/env.rs
```diff
@@ -16,6 +16,7 @@ use std::env;
 
 use crate::config::{ConfigError, ConfigSource, PartialConfig, PartialConfigBuilder};
 
+const CONFIG_DIR: &str = "CONFIG_DIR";
```
vaporos, line 19: Following the convention of the other two environment variables - the name of the const should be CONFIG_DIR_ENV and the environment variable should be SPLINTER_CONFIG_DIR.

splinterd/src/config/default.rs
```diff
@@ -14,8 +14,9 @@
 
 use crate::config::{ConfigError, ConfigSource, PartialConfig, PartialConfigBuilder};
 
-const DEFAULT_CERT_DIR: &str = "/etc/splinter/certs/";
-const DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
+pub(in crate::config) const CONFIG_DIR: &str = "/etc/splinter";
```
vaporos, line 17: These should remain private to this module.

shannynalayna: Fixed

splinterd/src/config/builder.rs
```diff
@@ -23,15 +23,23 @@ pub trait PartialConfigBuilder {
     fn build(self) -> Result<PartialConfig, ConfigError>;
 }
 
-fn get_file_path(cert_dir: &str, file: &str) -> String {
+fn get_tls_file_path(config_dir: &str, cert_dir: &str, file: &str) -> String {
     if file.starts_with("./") || file.starts_with("../") {
```
vaporos: Should this check for startswith('/') too? I think yes, in that if the user specified the full path "/tmp/foobar" then we don't want to prefix it with anything.

shannynalayna: That is addressed in the second commit

splinterd/man/splinterd.1.md
```diff
@@ -112,6 +112,13 @@ OPTIONS
   Any options on the command line will override the settings in the
   configuration file.
 
+`--config-dir CONFIG_DIR`
+: Specifies the directory containing configuration files. This is appended to
+  the directory containing certificates and associated key files if an absolute
```
vaporos, line 117: Technically "absolute path" is one that starts with "/", so this should be stated in a different way.

splinterd/src/config/builder.rs
```diff
@@ -23,15 +23,23 @@ pub trait PartialConfigBuilder {
...
         String::from(file)
-    } else {
+    } else if cert_dir.starts_with('/') || cert_dir.starts_with("./") || cert_dir.starts_with("../")
```
vaporos, line 29: A couple thoughts:

1. The absolute path check should be done with std::path::Path's absolute() check so it is cross-platform compatible.
2. The check for "./" and "../" should also check for the exact strings "." and ".." which would also be relative paths that we do not want to prefix. Really, then, we should split the path and (using stuff in std::path) and do this comparison on the first element of the path instead of start_with().

splinterd/src/config/default.rs
```diff
@@ -14,8 +14,9 @@
 
 use crate::config::{ConfigError, ConfigSource, PartialConfig, PartialConfigBuilder};
 
-const DEFAULT_CERT_DIR: &str = "/etc/splinter/certs/";
-const DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
+pub const CONFIG_DIR: &str = "/etc/splinter";
```
vaporos, line 17: This hsould be DEFAULT_CONFIG_DIR.

The pub should be removed here and on the following two lines.

splinterd/src/config/mod.rs
```diff
@@ -287,8 +296,24 @@ impl Config {
...
 
+    #[allow(clippy::cognitive_complexity)]
     /// Displays the configuration value along with where the value was sourced from.
     pub fn log_as_debug(&self) {
+        debug!(
+            "Config: config_dir: {:?} (source: {:?})",
+            self.config_dir(),
+            self.config_dir_source()
+        );
+        debug!(
```
vaporos, line 307: These were previously sorted alphabetically. That is probably still the best approach here?

splinterd/src/config/mod.rs
```diff
@@ -412,17 +427,14 @@ impl Config {
...
-
-    #[cfg(feature = "rest-api-cors")]
-    fn log_whitelist(&self) {
-        if let Some(list) = self.whitelist() {
-            debug!(
-                "Config: whitelist: {:?} (source: {:?})",
-                list,
-                self.whitelist_source()
-            );
+        {
```
vaporos, line 430: This change looks unrelated.

splinterd/src/config/mod.rs
shannynalayna: I don't think these were previously sorted alphabetically, it seems like they've always been sorted in the order the entry is in the Config object. I've added the log for the config_dir first, as this is now the first entry in the Config objects.

splinterd/man/splinterd.1.md
```diff
@@ -112,6 +112,10 @@ OPTIONS
   Any options on the command line will override the settings in the
   configuration file.
 
+`--config-dir CONFIG_DIR`
+: Specifies the directory containing configuration files.
+  (Default: `/etc/splinter`.)
```
chenette, line 117: Make consistent with other options:
- `CONFIG-DIR` (with a hyphen)
- Mention the `SPLINTER_CONFIG_DIR` environment variable here, following the model of `--tls-ca-file` (precede with a blank line and indent so it lines up with the existing description).

shannynalayna: Updated.

PR #748, opened 2020-04-24
Node registry REST API refactor

ltseeley:
Cleans up and refactors the node registry REST API code to be up to the standards of the other REST API code that has undergone stabilization.


libsplinter/src/node_registry/mod.rs
```diff
@@ -254,6 +254,12 @@ pub trait RwNodeRegistry: NodeRegistryWriter + NodeRegistryReader {
     ///  }
     ///```
     fn clone_box(&self) -> Box<dyn RwNodeRegistry>;
+
```
agunde406, line 261: Commit message says "Also fixes a typo in the `clone_box` doc string for `RwNodeRegistry`." but it doesnt look like that is actually included in this commit.

splinterd/api/static/openapi.yml
```diff
@@ -401,9 +401,13 @@ paths:
 
   /admin/nodes:
     post:
+      summary: Add a node to the registry
+      description: |
+        This endpoint can be used to a new node to the node registry. The node
```
agunde406, line 406: Typo: "a new node -> "add a new node"

libsplinter/src/node_registry/mod.rs
ltseeley: Removed from the commit message

splinterd/api/static/openapi.yml
ltseeley: Fixed

PR #749, opened 2020-04-24
Only use authorized connections and handle incoming connections

peterschwarz:
This PR does two things with the PeerManager, with respect to authorized connections:

* Validate identity returned is same as requested

    If the peer requested doesn't return the identity requested, the peer should be considered invalid and dropped.

* Accept temporary peers

    Accept incoming connections as temporary peers.  This state for the peer means that an remote node has peered with the local node, but that remote peer is not yet referenced locally for any reason.

    Once the local peer has referenced the peer id, the peer is promoted into the reference map and can participate in the fallback endpoint cascade.

    For example, it is not yet in any circuit relationship yet - either via negotiating a circuit or an existing circuit. In either of those circumstances, the admin service will add a peer reference, which would promote the peer connection.

Additionally,

- In order to enable this, it removes the special handling of inproc connections.  This should be left to the authorization code to determine what sort of conversation needs to be had with the remote connection. Most likely, they should still provide a valid identity.
- peers can no longer change their id, as this is determined via the authorization process.

### DIFF
--- a/libsplinter/src/network/peer_manager/mod.rs
+++ b/libsplinter/src/network/peer_manager/mod.rs
@@ -347,7 +347,19 @@ fn add_peer(
     // If new, try to create a connection
     for endpoint in endpoints.iter() {
         match connector.request_connection(&endpoint, &connection_id) {
-            Ok(_identity) => {
+            Ok(identity) => {
+                if peer_id != identity {
+                    if let Err(err) = connector.remove_connection(&endpoint) {
+                        error!("Unable to clean up mismatched identity connection: {}", err);
+                    }
+                    // remove the reference created above
+                    ref_map.remove_ref(&peer_id);
+                    return Err(PeerRefAddError::AddError(format!(
+                        "Peer {} (via {}) presented a mismatched identity {}",
+                        peer_id, endpoint, identity
+                    )));
+                }
+
                 debug!("Peer {} connected via {}", peer_id, endpoint);
                 peers.insert(
                     peer_id.clone(),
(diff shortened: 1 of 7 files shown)


libsplinter/src/network/peer_manager/mod.rs
```diff
@@ -109,6 +113,12 @@ impl Drop for PeerRef {
     }
 }
 
+/// An entry of temporary peers, that may connected externally, but not yet requested locally.
+pub struct TemporaryPeer {
```
agunde406, line 117: I think this is going to cause an issue with PeerInterconnect since this information is not returned with the Connection id request. So if the PeerInterconnect gets a message from a temporary peer, for example a circuit creation request, PeerInterconnect will not know who the request is from. 

The temporary peers information should be returned in ConnectionIds request as if it was a promoted peer.

peterschwarz: The temporary status is different from the old idea of temporary peers. In the peer manager, temporary peers are only visible internally (the one function that list them is more for tools to have some visibility into the process - like the status endpoint, for example). 

The temporary status is also only for incoming connections. Admin service will still request a peer ref to make a circuit, which is the point at which the temp peer will be promoted.

peterschwarz: It probably should be pub, though... I might have left that in.

peterschwarz: It probably should be pub, though... I might have left that in.

agunde406: After discussion, there will be a follow up PeerInterconnect PR.

peterschwarz: Removed the `pub` from the struct

PR #750, opened 2020-04-26
 Negotiate protocol version between Admin Services

agunde406:
This PR adds a step after authorization but before
processing a circuit request that requires the admin
services to agree on a protocol version. After a
peer is authorized, the local admin service will send
a request for a protocol version, providing its own
min and max supported protocol versions. The remote
admin service will respond with the agreed protocol,
sending 0 if there is not a matching protocol version.

This is a breaking change because an older AdminService
will not be able to respond with the required response
and therefore will never enter consensus for a proposal.

Finally, Renames UnpeeredPendingPayload to PendingPaylaod
because it now includes tracking for which services need to
agree on a protocol version.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -38,7 +38,10 @@ use crate::network::{
     peer::PeerConnector,
 };
 use crate::orchestrator::ServiceOrchestrator;
-use crate::protos::admin::{AdminMessage, AdminMessage_Type, CircuitManagementPayload};
+use crate::protocol::{ADMIN_PROTOCOL_VERSION, ADMIN_SERVICE_PROTOCOL_MIN};
+use crate::protos::admin::{
+    AdminMessage, AdminMessage_Type, CircuitManagementPayload, ServiceProtocolVersionResponse,
+};
 #[cfg(feature = "service-arg-validation")]
 use crate::service::validation::ServiceArgValidator;
 use crate::service::{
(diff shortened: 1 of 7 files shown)


libsplinter/src/admin/service/mod.rs
```diff
@@ -553,6 +634,36 @@ mod tests {
             .propose_circuit(payload, "test".to_string())
             .expect("The proposal was not handled correctly");
 
+        // wait up to 1 second for the service protocol version request
+        let message;
+        let start = Instant::now();
+        loop {
+            if Instant::now().duration_since(start) > Duration::from_secs(1) {
```
vaporos, line 641: 1 second might be too short of a duration on very bad networks.

peterschwarz: Since this is in a test, I would suggest 60 seconds, as that's where cargo starts to complain about a test taking too long anyway.

agunde406: Updated the test to wait for 60 seconds

libsplinter/src/admin/service/shared.rs
```diff
@@ -841,46 +929,131 @@ impl AdminServiceShared {
...
+                        .missing_protocol_ids
+                        .iter()
+                        .any(|missing_protocol_id| missing_protocol_id == service_id)
+                    {
+                        warn!(
+                            "Dropping circuit request including service {}, \
+                             due to protocol mismatch",
+                            service_id
+                        );
+                        pending_protocol_payload.missing_protocol_ids.clear();
```
peterschwarz, line 1012: Doesn't this delete all of the missing protocol ids?

agunde406: yes

PR #758, opened 2020-04-28
scabbard-get-state fixes

ltseeley:
A variety of fixes and improvements for stabilizing the `scabbard-get-state` feature.

This PR addresses all open issues with the feature except testing, which will be the subject of a future PR.


libsplinter/src/service/scabbard/state.rs
```diff
@@ -834,91 +840,229 @@ impl Iterator for ChannelBatchInfoIter {
...
+        );
+    }
 
-        let test_result = std::panic::catch_unwind(|| {
-            let transaction_receipt_store =
-                Arc::new(RwLock::new(TransactionReceiptStore::new(Box::new(
-                    LmdbOrderedStore::new(&temp_db_path, Some(TEMP_DB_SIZE))
-                        .expect("Failed to create LMDB store"),
-                ))));
+    /// Verify that the event iterator works as expected.
```
vaporos, line 879: Tests need more description, this is not enough information to determine in the future whether modifications to the test would invalidate the original intent of the test.

ltseeley: Added more verbose descriptions for the tests

PR #761, opened 2020-04-28
DefaultConfig updates

shannynalayna:
Adds a way to construct a default node_id, as well as saving the node id to a file and verifying the node_id in the config matches the node_id saved (if present)

### DIFF
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -131,14 +131,21 @@ impl ConfigBuilder {
                 None => None,
             })
             .ok_or_else(|| ConfigError::MissingValue("network endpoints".to_string()))?;
-        let node_id = self
+        let node_id = self.partial_configs.iter().find_map(|p| match p.node_id() {
+            Some(v) => Some((v, p.source())),
+            None => None,
+        });
+        let display_name = self
             .partial_configs
             .iter()
-            .find_map(|p| match p.node_id() {
+            .find_map(|p| match p.display_name() {
                 Some(v) => Some((v, p.source())),
                 None => None,
             })
-            .ok_or_else(|| ConfigError::MissingValue("node id".to_string()))?;
+            .unwrap_or_else(|| match &node_id {
+                Some((id, _)) => (format!("Node {}", id), ConfigSource::Default),
+                None => (String::from("Node"), ConfigSource::Default),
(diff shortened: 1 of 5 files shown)


splinterd/src/config/mod.rs
```diff
@@ -459,12 +459,20 @@ mod tests {
...
     static EXAMPLE_DISPLAY_NAME: &str = "Node 1";
 
+    /// Default configuration values
     static DEFAULT_CLIENT_CERT: &str = "client.crt";
     static DEFAULT_CLIENT_KEY: &str = "client.key";
     static DEFAULT_SERVER_CERT: &str = "server.crt";
     static DEFAULT_SERVER_KEY: &str = "server.key";
     static DEFAULT_CA_CERT: &str = "ca.pem";
 
+    static DEFAULT_STATE_DIR: &str = "/var/lib/splinter/";
```
ltseeley, line 469: The changes here seem to be unrelated to the commit

shannynalayna: I added a bit to the commit message to explain why the defaults were added. Mainly, because the node_id is now in the DefaultPartialConfigBuilder, a Config object can be successfully built from just that. Therefore, the test which previously errored because of a missing node_id, now passes and is able to create a Config object. So, adding all of these defaults is a way to verify the contents of that Config object.

splinterd/src/config/partial.rs
```diff
@@ -24,6 +24,12 @@ pub enum ConfigSource {
     CommandLine,
 }
 
+impl Default for ConfigSource {
```
vaporos, line 27: Commit message doesn't say why.

splinterd/src/config/builder.rs
```diff
@@ -131,11 +154,29 @@ impl ConfigBuilder {
...
+            })
+            .ok_or_else(|| ConfigError::MissingValue("state directory".to_string()))?;
+
         let node_id = self
             .partial_configs
             .iter()
             .find_map(|p| match p.node_id() {
-                Some(v) => Some((v, p.source())),
+                Some(v) => {
+                    if p.source() == ConfigSource::Default {
```
vaporos, line 171: If the node id file exists, it must always match the configuration or we should produce a fatal error.

PR #763, opened 2020-04-29
Rewrite WsTransport in tungstenite and add SSL support

vaporos:
The websocket crate had serious limitations which manifested when attempting to implement TLS support. After more analysis, Tungstenite looks better overall, so re-implemented with that dependency instead.

Added TLS support with wss:// URLs.

To test, you can run two nodes:

Node1:

```
#!/bin/bash

set -e

export SPLINTER_HOME=$(pwd)/home/node1
cargo run \
    --manifest-path cli/Cargo.toml \
    cert generate \
    --skip
cargo run \
    --manifest-path splinterd/Cargo.toml \
    --features experimental \
    -- \
    -vv \
    --tls-insecure \
    --network-endpoints tcp://127.0.0.1:18044 \
    --network-endpoints tcps://127.0.0.1:18045 \
    --network-endpoints ws://127.0.0.1:18046 \
    --network-endpoints wss://127.0.0.1:18047 \
    --service-endpoint tcp://127.0.0.1:18048 \
    --rest-api-endpoint 127.0.0.1:18080 \
    --peers wss://127.0.0.1:28047
```

Node 2:

```
#!/bin/bash

set -e

export SPLINTER_HOME=$(pwd)/home/node2
cargo run \
    --manifest-path cli/Cargo.toml \
    cert generate \
    --skip
cargo run \
    --manifest-path splinterd/Cargo.toml \
    --features experimental \
    -- \
    -vv \
    --tls-insecure \
    --network-endpoints tcp://127.0.0.1:28044 \
    --network-endpoints tcps://127.0.0.1:28045 \
    --network-endpoints ws://127.0.0.1:28046 \
    --network-endpoints wss://127.0.0.1:28047 \
    --service-endpoint tcp://127.0.0.1:28048 \
    --rest-api-endpoint 127.0.0.1:28080
```

### DIFF
--- a/libsplinter/src/transport/ws/connection.rs
+++ b/libsplinter/src/transport/ws/connection.rs
@@ -12,63 +12,55 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io::{self, Read, Write};
-use std::os::unix::io::AsRawFd;
+use std::io::{Read, Write};
 
-use mio::{unix::EventedFd, Evented, Poll, PollOpt, Ready, Token};
-use websocket::{
-    client::sync::Client,
-    message::{Message, OwnedMessage::Binary},
-    result::WebSocketError,
-    stream::sync::{AsTcpStream, Stream},
-};
+use mio::Evented;
+use tungstenite::{protocol::WebSocket, Message};
 
 use crate::transport::{Connection, DisconnectError, RecvError, SendError};
 
-pub(super) struct WsClientConnection<S>
+pub(super) struct WsConnection<S>
 where
-    S: Read + Write + Send,
+    S: Read + Write + Send + Evented,
 {
-    client: Client<S>,
+    websocket: WebSocket<S>,
     remote_endpoint: String,
     local_endpoint: String,
 }
 
-impl<S> WsClientConnection<S>
+impl<S> WsConnection<S>
 where
-    S: Read + Write + Send,
+    S: Read + Write + Send + Evented,
 {
-    pub fn new(client: Client<S>, remote_endpoint: String, local_endpoint: String) -> Self {
-        WsClientConnection {
-            client,
+    pub fn new(websocket: WebSocket<S>, remote_endpoint: String, local_endpoint: String) -> Self {
+        WsConnection {
+            websocket,
             remote_endpoint,
             local_endpoint,
         }
     }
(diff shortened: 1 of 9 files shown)


libsplinter/src/transport/ws/connection.rs
```diff
@@ -12,63 +12,55 @@
...
-                        websocket::message::OwnedMessage::Binary"
-                        .to_string(),
+                    "message received was not binary".to_string(),
                 )),
             },
-            Err(WebSocketError::IoError(ref e)) if e.kind() == io::ErrorKind::WouldBlock => {
-                Err(RecvError::WouldBlock)
-            }
-            Err(WebSocketError::NoDataAvailable) => Err(RecvError::WouldBlock),
+            Err(tungstenite::error::Error::Io(e)) => Err(RecvError::from(e)),
```
peterschwarz, line 63: I think your `From<tungstenite::error::Error>` covers all of these errors. You may be able to convert the error part of this match statement to a `?`.

vaporos: This piece was reworked a bit to cut down on the duplication.

libsplinter/src/transport/ws/listener.rs
```diff
@@ -16,58 +16,92 @@ use std::net::TcpListener;
...
-            {
-                |mut handshake_err| loop {
-                    match handshake_err {
-                        HandshakeError::Interrupted(mid_handshake) => {
-                            thread::sleep(Duration::from_millis(100));
-                            match mid_handshake.handshake() {
-                                Ok(ok) => break Ok(ok),
-                                Err(err) => handshake_err = err,
+        if let Some(acceptor) = &self.acceptor {
+            let remote_endpoint = format!("{}{}", WS_PROTOCOL_PREFIX, stream.peer_addr()?);
```
agunde406, line 53: This is reversed, should be WSS_PROTOCOL_PREFIX ( and WS_PROTOCOL_PREFIX below)

vaporos: fixed

vaporos:
> Seen in testing - handshake appears to take 1+ minute

I believe this is not the handshake but rather the duration of time the peer manager is using to retry peer connections (60 seconds by default).

PR #771, opened 2020-05-01
Move scabbard

ltseeley:
Moves all scabbard code out of libsplinter and into scabbard-specific crates


vaporos:
Seems like this should be either:

* scabbard (libscabbard/) - service + client
* scabbard-cli (cli/)

or:

* scabbard-service (service/)
* scabbard-client (client/)
* scabbard-cli (cli/)

Trade-off is that the first approach potentially requires more features, and the second approach requires publishing another crate.

Currently we have the 2nd approach, and if we keep that, I propose renaming the service code to scabbard-service (service/).

services/scabbard/libscabbard/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+name = "scabbard"
+version = "0.3.17"
+authors = ["Cargill Incorporated"]
+edition = "2018"
+license = "Apache-2.0"
+description = """\
+    Splinter is a privacy-focused platform for distributed applications that \
+    provides a blockchain-inspired networking environment for communication \
+    and transactions between organizations.
+"""
```
chenette, line 25: Here, I would expect a description of the scabbard package, not the crate splinter blurb. Information on related packages and separate crates (such as crate splinter) would be helpful.

services/scabbard/client/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+name = "scabbard-client"
+version = "0.3.17"
+authors = ["Cargill Incorporated"]
+edition = "2018"
+license = "Apache-2.0"
+description = """\
+    Splinter is a privacy-focused platform for distributed applications that \
+    provides a blockchain-inspired networking environment for communication \
+    and transactions between organizations.
+"""
```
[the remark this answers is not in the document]
ltseeley: The client crate has been removed

services/scabbard/libscabbard/Cargo.toml
ltseeley: Updated to be a description of scabbard (taken from the scabbard module documentation)

PR #783, opened 2020-05-05
 Stabilize config-toml feature

shannynalayna:


shannynalayna:
Since this feature is already in `default`, this pr doesn't necessarily make sense

PR #790, opened 2020-05-05
Dispatcher updates in support of PeerManager integration with splinterd/AdminService

agunde406:


libsplinter/src/network/dispatch.rs
```diff
@@ -602,6 +618,37 @@ where
     }
 }
 
+impl<MT, Source> DispatchLoop<MT, Source>
+where
+    MT: Any + Hash + Eq + Debug + Clone + Send,
+    Source: Send + 'static,
+{
+    pub fn builder() -> DispatchLoopBuilder<MT, Source> {
```
peterschwarz, line 610: This should probably be in a separate commit

agunde406: split out

agunde406: split out

libsplinter/src/network/dispatch.rs
```diff
@@ -289,6 +289,8 @@ pub trait MessageSender<R>: Send {
     ///
     /// If an error occurs, return the intended recipient and message bytes.
     fn send(&self, reciptient: R, message: Vec<u8>) -> Result<(), (R, Vec<u8>)>;
+
+    fn box_clone(&self) -> Box<dyn MessageSender<R>>;
```
peterschwarz, line 293: Why is this necessary?

agunde406: I added it for passing it into the create_authorization_dispatcher and thought it would be useful but I will rework that and remove it.

agunde406: auth2::handlers::create_authorization_dispatcher takes a MessageSender<ConnectionId>> and that needs to be wrapped in Box. That function clones the message sender to create two dispatchers.

agunde406: I was able to rework that function. removed box_clone

PR #794, opened 2020-05-07
Config cleanup

shannynalayna:
A few cleanup tasks surrounding the splinterd config, including adding a `state_dir` option to the splinterd CLI, as well as using this value for the AdminService.

Also adds some documentation. And adds a check for the default node diplay name to incorporate the randomly generated node_id, if neither the node_id nor display_name are set.

### DIFF
--- a/splinterd/src/config/builder.rs
+++ b/splinterd/src/config/builder.rs
@@ -139,21 +139,6 @@ impl ConfigBuilder {
                 None => None,
             })
             .ok_or_else(|| ConfigError::MissingValue("network endpoints".to_string()))?;
-        let node_id = self.partial_configs.iter().find_map(|p| match p.node_id() {
-            Some(v) => Some((v, p.source())),
-            None => None,
-        });
-        let display_name = self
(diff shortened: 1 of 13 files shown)


splinterd/src/config/builder.rs
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! ConfigBuilder implementation to construct a finalized Config object.
```
ltseeley, line 15: References to code entities (like "ConfigBuilder") should be between ticks (`) to show up as code in the documentation.

splinterd/src/main.rs
```diff
@@ -421,6 +421,10 @@ fn start_daemon(matches: ArgMatches) -> Result<(), UserError> {
     config.log_as_debug();
 
     let node_id = find_node_id(&config)?;
+    let display_name = match config.display_name_source() {
```
vaporos, line 425: This should just check if it is None.

PR #807, opened 2020-05-08
Remove Connection Manager NotificationIter

peterschwarz:
Remove the `NotificationIter`, as this struct is only used in tests, and has been supplanted by the `subscribe` function.  The iterator behavior can be easily recreated simply by using the receiving end of the channel.

This also removes the `connection-manager-notification-iter-try-next` feature, which is no longer needed with the removal of the iterator.

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -25,7 +25,7 @@ use std::time::Instant;
 use uuid::Uuid;
 
 pub use error::{AuthorizerError, ConnectionManagerError};
-pub use notification::{ConnectionManagerNotification, NotificationIter};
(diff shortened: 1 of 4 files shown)


libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -25,7 +25,7 @@ use std::time::Instant;
 use uuid::Uuid;
 
 pub use error::{AuthorizerError, ConnectionManagerError};
-pub use notification::{ConnectionManagerNotification, NotificationIter};
+pub use notification::ConnectionManagerNotification;
```
agunde406, line 28: Commit message still says it removes connection-manager-notification-iter-try-next  but that has already been removed

PR #810, opened 2020-05-08
Connect/Disconnect handlers for Service Component Messages

peterschwarz:
This PR includes the handlers for the base component `ServiceMessage` envelope, as well as the connect (`SMConnectRequest`) and disconnect (`SMDisconnectRequest`).  The actual operations are abstracted behind a `ServiceInstances` trait.  This will remove the tight coupling between the handlers and `SplinterState` (as the previous "network" version of these message handlers were).

Additionally, 

- Adds native protocol structs for the component and service protobuf messages.
- Adds getters to the ServiceId struct
- Adds PartialEq to the PeerId/ConnectionId structs

Notes:

- Handlers for the `ServiceProcessMessage` will be added in a future PR
- ~Implementation of the `ServiceInstances` will be added in a future PR~

### DIFF
--- a/libsplinter/src/service/network/handlers.rs
+++ b/libsplinter/src/service/network/handlers.rs
@@ -13,3 +13,58 @@
 // limitations under the License.
 
 //! Dispatch handlers for service component messages.
+
+use crate::circuit::service::ServiceId;
+use crate::network::dispatch::{
+    ConnectionId, DispatchError, DispatchMessageSender, Handler, MessageContext, MessageSender,
+};
+use crate::protos::component;
+use crate::protos::service;
+
+/// Dispatch handler for the service message envelope.
+pub struct ServiceMessageHandler {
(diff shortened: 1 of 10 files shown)


libsplinter/src/service/network/handlers.rs
```diff
@@ -68,3 +68,134 @@ impl Handler for ServiceMessageHandler {
             })
     }
 }
+
+/// A service registry stores the service information provide by service components.  This can be
+/// used to add or remove service registration information.
+pub trait ServiceRegistry {
```
vaporos, line 80: Since we have a registry as a top-level component (was node registry, now Splinter registry), this probably needs a new name to avoid confusion.

peterschwarz: Renamed to ServiceInstances

peterschwarz:
Rebased on master, after the 0.3.17 release

peterschwarz:
Added commits to implement the `ServiceInstances` trait over `SplinterState`

PR #811, opened 2020-05-08
Removing blocking on authorization from ConnectionManager and PeerManager

agunde406:
Before add_peer_ref would block until the underlying
peer connection completed authorization. This will be a
problem in the AdminService when all peers must be connected
before the payload can be processed.

This commit changes add_peer_ref to return a peer_ref
before the peer is ready to be used and the peer is set as
Pending. When the underlying connection has completed
authorization, a PeerMangerNotfication::PeerReady is
sent to subscribers and the peer is set to Connected. If
authorization is failed, a BadPeer notification is sent
instead and the peer is set to Invalid.

add_unidentified_peer requests a connection and the resulting
ConnectionReady notification adds the peer as a UnreferencedPeers.

To handle the case where a Pending peer cannot be connected to at
the time of request, a PeerRef is still returned and then the peer
will be retried with an exponential back off. A Pacemaker was added
to drive this retry with the new PeerManagerMessage::RetryPending
message.

### DIFF
--- a/libsplinter/src/network/connection_manager/mod.rs
+++ b/libsplinter/src/network/connection_manager/mod.rs
@@ -228,9 +228,12 @@ where
                         Ok(CmMessage::AuthResult(auth_result)) => {
                             handle_auth_result(auth_result, &mut state, &mut subscribers);
                         }
-                        Ok(CmMessage::SendHeartbeats) => {
-                            send_heartbeats(&mut state, &mut subscribers)
-                        }
+                        Ok(CmMessage::SendHeartbeats) => send_heartbeats(
+                            &mut state,
+                            &mut subscribers,
+                            &*authorizer,
+                            resender.clone(),
+                        ),
                         Err(_) => {
                             warn!("All senders have disconnected");
                             break;
@@ -764,6 +767,8 @@ where
         &mut self,
         endpoint: &str,
         subscribers: &mut SubscriberMap,
+        authorizer: &dyn Authorizer,
+        internal_sender: Sender<CmMessage>,
     ) -> Result<(), ConnectionManagerError> {
         let mut meta = if let Some(meta) = self.connections.get_mut(endpoint) {
             meta.clone()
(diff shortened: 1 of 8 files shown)


libsplinter/src/network/peer_manager/notification.rs
```diff
@@ -21,6 +21,10 @@ use std::sync::mpsc::{Receiver, TryRecvError};
 pub enum PeerManagerNotification {
     Connected { peer: String },
     Disconnected { peer: String },
+    PeerReady { peer: String },
+    // returned if the peer has failed for some reason, such as authorization failure and any
+    // peer refs should be dropped
+    BadPeer { peer: String },
```
peterschwarz, line 27: This should be a doc comment.  `BadPeer` isn't a great name, but I don't yet have better name.  We can fix rename that in followup PR.

agunde406: changed to doc comments, ya i don't love that name either

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -687,10 +722,19 @@ where
...
-                Ok(identity)
+                subscribers.broadcast(ConnectionManagerNotification::Connected {
+                    endpoint,
+                    connection_id,
+                    identity,
+                });
             }
             AuthorizationResult::Unauthorized { connection_id, .. } => {
-                Err(ConnectionManagerError::Unauthorized(connection_id))
+                // If the connection is unathorized, notify subscriber this is a bad connection
```
vaporos, line 732: Typo in comment - 'unathorized'

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -730,11 +782,14 @@ where
                     connection_id,
                     identity,
                 });
-
-                Ok(())
             }
             AuthorizationResult::Unauthorized { connection_id, .. } => {
-                Err(ConnectionManagerError::Unauthorized(connection_id))
+                // If the connection is unathorized, notify subscriber this is a bad connection
```
vaporos, line 787: Same typo

libsplinter/src/network/peer_manager/peer_map.rs
```diff
@@ -151,6 +151,12 @@ impl PeerMap {
             .values()
             .find(|meta| meta.connection_id == connection_id)
     }
+
```
vaporos, line 154: Typo in comment "Add get_pending method to PeerMap " - "ha" -- also, no period after the last sentence.

PR #812, opened 2020-05-09
Remove storage_type and use storage location in AdminService

agunde406:
The AdminService will now take the path to proposals
location directly, instead of configuring it based
an storage locations.

Storage locations are now configured in only 1 place
instead of passing around storage type and locations.


agunde406:
Closing in favor of https://github.com/Cargill/splinter/pull/813

PR #815, opened 2020-05-10
Integrate PeerManager/Connection Manager into Splinterd

agunde406:
This PR removes the use of Network and the old PeerConnector with the new ConnectionManager and PeerManager. 

This change is to support better re connection and connection stability. 

To test, start up gameroom and create a gameroom. 

There are a couple know bugs around reconnection/authorization that will be seen:
 
InboundConnections are not properly clean up on reconnection if after a gameroom is created you shutdown and restart 1 side. 

During normal connection you will see a message that there is not a Dispatcher for Authorization message types. 

Also some errors can been seen on shutdown due to  shutdown ordering.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -239,7 +239,9 @@ impl AdminService {
     /// circuits should be initialized using the service orchestrator. This may not include all
     /// services if they are not supported locally. It is expected that some services will be
     /// started externally.
-    fn restart_services(&self) -> Result<(), ServiceStartError> {
+    ///
+    /// Also adds peer references for members of the circuits and proposals.
+    fn re_initialize_circuits(&self) -> Result<(), ServiceStartError> {
         let circuits = self
             .admin_service_shared
             .lock()
@@ -249,12 +251,40 @@ impl AdminService {
             .get_circuits()
             .map_err(|err| ServiceStartError::Internal(Box::new(err)))?;
 
+        let nodes = self
+            .admin_service_shared
+            .lock()
+            .map_err(|_| {
+                ServiceStartError::PoisonedLock("the admin shared lock was poisoned".into())
+            })?
+            .get_nodes()
+            .map_err(|err| ServiceStartError::Internal(Box::new(err)))?;
+
         let orchestrator = self.orchestrator.lock().map_err(|_| {
             ServiceStartError::PoisonedLock("the admin orchestrator lock was poisoned".into())
         })?;
-
+        let mut peer_refs = vec![];
         // start all services of the supported types
         for (circuit_name, circuit) in circuits.iter() {
+            // restart all peer in the circuit
+            for member in circuit.members() {
+                if member != &self.node_id {
+                    if let Some(node) = nodes.get(member) {
+                        let peer_ref = self
+                            .peer_connector
+                            .add_peer_ref(member.to_string(), node.endpoints().to_vec());
+
+                        if let Ok(peer_ref) = peer_ref {
+                            peer_refs.push(peer_ref);
+                        } else {
+                            info!("Unable to peer with {} at this time", member);
+                        }
+                    } else {
+                        error!("Missing node information for {}", member);
+                    }
+                }
+            }
+
             // Get all services this node is allowed to run and the orchestrator has a factory for
             let services = circuit
                 .roster()
@@ -294,6 +324,37 @@ impl AdminService {
             }
         }
 
+        let proposals = self
+            .admin_service_shared
+            .lock()
+            .map_err(|_| {
+                ServiceStartError::PoisonedLock("the admin shared lock was poisoned".into())
+            })?
+            .get_proposals();
+
+        for (_, proposal) in proposals.iter() {
+            // restart all peer in the circuit
+            for member in proposal.circuit.members.iter() {
+                if member.node_id != self.node_id {
+                    let peer_ref = self
+                        .peer_connector
+                        .add_peer_ref(member.node_id.to_string(), member.endpoints.to_vec());
+
+                    if let Ok(peer_ref) = peer_ref {
+                        peer_refs.push(peer_ref);
+                    } else {
+                        info!("Unable to peer with {} at this time", member.node_id);
+                    }
+                }
+            }
+        }
+
+        self.admin_service_shared
+            .lock()
+            .map_err(|_| {
+                ServiceStartError::PoisonedLock("the admin shared lock was poisoned".into())
+            })?
+            .add_peer_refs(peer_refs);
         Ok(())
     }
 }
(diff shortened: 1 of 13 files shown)


libsplinter/src/admin/service/mod.rs
```diff
@@ -23,7 +23,8 @@ mod shared;
 use std::any::Any;
 #[cfg(feature = "service-arg-validation")]
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, MutexGuard};
```
peterschwarz, line 26: This might be in the wrong commit

libsplinter/src/transport/multi.rs
```diff
@@ -31,6 +31,10 @@ impl MultiTransport {
     pub fn new(transports: Vec<SendableTransport>) -> Self {
         Self { transports }
     }
+
```
peterschwarz, line 34: Can't you just pass a clone of the inproc transport before creating the multi transport?

libsplinter/src/admin/service/shared.rs
```diff
@@ -765,53 +861,49 @@ impl AdminServiceShared {
...
-                            let mut msg = AdminMessage::new();
-                            msg.set_message_type(
-                                AdminMessage_Type::SERVICE_PROTOCOL_VERSION_REQUEST,
-                            );
-                            msg.set_protocol_request(request);
-
-                            let envelope_bytes = msg.write_to_bytes()?;
-                            network_sender
-                                .send(&admin_service_id(node.get_node_id()), &envelope_bytes)?;
+                debug!("Connecting to node {:?}", node);
```
peterschwarz, line 874: Should we use the phrase "Connecting to" here?  Perhaps "Referencing", so that the actual connection state isn't confusing in the logs.

libsplinter/src/admin/service/mod.rs
```diff
@@ -221,7 +221,9 @@ impl AdminService {
     /// circuits should be initialized using the service orchestrator. This may not include all
     /// services if they are not supported locally. It is expected that some services will be
     /// started externally.
-    fn restart_services(&self) -> Result<(), ServiceStartError> {
+    ///
+    /// Also creates the requests the connection for members of the circuits and proposals
```
peterschwarz, line 225: This is a strange sentence.  I'm not sure what the intent is.

libsplinter/src/admin/service/mod.rs
```diff
@@ -221,7 +221,9 @@ impl AdminService {
...
+    fn restart_circuits(&self) -> Result<(), ServiceStartError> {
```
peterschwarz, line 226: Perhaps this should be `re_initialize_circuits`, as this is doing more than restarting (and re-referencing isn't a restart operation, really).

libsplinter/src/admin/service/mod.rs
```diff
@@ -524,6 +524,24 @@ impl Service for AdminService {
...
 }
 
+fn handle_peer_manager_notification(
+    notification: PeerManagerNotification,
+    mut admin_shared: MutexGuard<AdminServiceShared>,
+) {
+    match notification {
+        PeerManagerNotification::Connected { peer } => {
+            if let Err(err) = admin_shared.on_peer_connected(&peer) {
+                debug!("Peer {} has connected", peer);
```
vaporos, line 595: Should this be outside the if statement?

libsplinter/src/admin/service/mod.rs
agunde406: whoops yep

libsplinter/src/transport/multi.rs
agunde406: We decided to leave this

libsplinter/src/admin/service/shared.rs
agunde406: updated

libsplinter/src/admin/service/mod.rs
agunde406: whoops, that was a combo of two diff sentences. fixed

libsplinter/src/admin/service/mod.rs
agunde406: updated

libsplinter/src/admin/service/mod.rs
agunde406: fixed

libsplinter/src/admin/service/mod.rs
```diff
@@ -360,6 +335,32 @@ impl Service for AdminService {
...
+                };
+
+                if let Ok(admin_shared) = peer_admin_shared.lock() {
+                    handle_peer_manager_notification(notification, admin_shared);
+                } else {
+                    error!("the admin shared lock was poisoned");
+                    break;
+                }
+            })
+            .map_err(|err| ServiceStartError::Internal(Box::new(err)))?;
```
vaporos, line 423: This thread should probably get joined on shutdown.

agunde406: update to return join handle from the admin service. Also updated the shutdown in splinterd to shutdown cleanly.

libsplinter/src/admin/service/mod.rs
```diff
@@ -523,6 +515,24 @@ impl Service for AdminService {
     }
 }
 
+fn handle_peer_manager_notification(
+    notification: PeerManagerNotification,
+    mut admin_shared: MutexGuard<AdminServiceShared>,
```
peterschwarz, line 581: You should be able to just have this be `&mut AdminServiceShared`, and not need the mutex guard.  Might need to pass in `&mut *admin_shared` at the call site.

libsplinter/src/admin/service/shared.rs
```diff
@@ -400,8 +418,12 @@ impl AdminServiceShared {
                     }
                     Ok(CircuitProposalStatus::Rejected) => {
                         // remove circuit
-                        self.remove_proposal(&circuit_id)?;
-
+                        let proposal = self.remove_proposal(&circuit_id)?;
+                        if let Some(proposal) = proposal {
+                            for member in proposal.get_circuit_proposal().members.iter() {
+                                self.remove_peer_ref(member.get_node_id());
```
peterschwarz, line 424: Is there a peer ref per proposal?  Looking at the add_peer_ref, there's a unique peer ref per peer in this state.  Shouldn't we only remove the peer ref if no other proposal has that peer?

agunde406: add_peer_ref adds the peer_ref to a list of peer_refs for that peer_id. remove_peer_ref pops one peer ref off, only keeping the list around if its not empty. This does not remove the peer simply drops one of the peer refs (which will remove a 1 reference from the peer manager)

PR #821, opened 2020-05-12
Wait for notification in  test_peer_interconnect

agunde406:
With the new async behavior of the connection manager and
the peer manager, before running the interconnect test
we need to wait until the peer comes back as connected.

Also ups the timeout time to 60 seconds instead of 2.

### DIFF
--- a/libsplinter/src/network/peer_manager/interconnect.rs
+++ b/libsplinter/src/network/peer_manager/interconnect.rs
@@ -397,7 +397,7 @@ pub mod tests {
         dispatch_channel, DispatchError, DispatchLoopBuilder, Dispatcher, Handler, MessageContext,
         MessageSender, PeerId,
     };
-    use crate::network::peer_manager::PeerManager;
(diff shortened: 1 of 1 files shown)


libsplinter/src/network/peer_manager/interconnect.rs
```diff
@@ -523,14 +523,22 @@ pub mod tests {
...
+
         let peer_ref = peer_connector
             .add_peer_ref("test_peer".to_string(), vec!["test".to_string()])
             .expect("Unable to add peer");
 
         assert_eq!(peer_ref.peer_id, "test_peer");
 
-        // wait to be told to shutdown, timeout after 2 seconds
-        let test_timeout = std::time::Duration::from_secs(2);
+        let notification = subscriber.next().expect("Unable to notification");
```
vaporos, line 534: Message should be "Unable to get notification" if the intent was the same as the prior message.

agunde406: fixed

PR #826, opened 2020-05-12
Modify authorization state machine

peterschwarz:
This change require tracking the states more closely. The states of remote acceptance and remote identified have been added, so that these states can be properly transitioned through to the final Authorized state.  

It also allows for the use of the state value to indicate whether or not authorization is complete, as well as hold the identity.

### DIFF
--- a/libsplinter/src/network/auth/handlers.rs
+++ b/libsplinter/src/network/auth/handlers.rs
@@ -26,8 +26,8 @@ use crate::protos::network::{NetworkMessage, NetworkMessageType};
 use crate::protos::prelude::*;
 
 use super::{
-    AuthorizationAction, AuthorizationManagerStateMachine, AuthorizationMessageSender,
-    AuthorizationState,
+    AuthorizationAction, AuthorizationActionError, AuthorizationManagerStateMachine,
+    AuthorizationMessageSender, AuthorizationState,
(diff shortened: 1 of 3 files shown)


libsplinter/src/network/auth/handlers.rs
```diff
@@ -120,10 +128,31 @@ impl Handler for AuthorizedHandler {
...
+        debug!(
+            "Received authorize message from {}",
+            context.source_connection_id()
+        );
+        match self.auth_manager.next_state(
+            context.source_connection_id(),
+            AuthorizationAction::RemoteAuthorizing,
+        ) {
+            Err(err) => {
+                debug!(
```
vaporos, line 140: should this be warn?

vaporos: should this cause the connection to close?

peterschwarz: The connection manager will get the `Unauthorized` result, and close the connection.

peterschwarz: Possibly those should all be `warn`. I'll add a commit for that.

peterschwarz:
Added a commit for the log levels, and specificity.

PR #829, opened 2020-05-12
Always shutdown scheduler in scabbard

ltseeley:
Fixes a bug where an invalid transaction resulted in the scheduler not
being shutdown by scabbard.


services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -241,6 +241,8 @@ impl ScabbardState {
             .map_err(|_| ScabbardStateError("failed to receive result in reasonable time".into()))?
             .ok_or_else(|| ScabbardStateError("no result returned from executor".into()))?;
 
+        scheduler.shutdown();
```
peterschwarz, line 244: Couldn't any one of lines 239-242 still create the same problem?

ltseeley: Good catch; fixed

PR #833, opened 2020-05-13
Fix scabbard crates publish

rbuysse:


ci/publish-crates
```diff
@@ -19,6 +19,7 @@ export VERSION=AUTO_STRICT
 export REPO_VERSION=$($top_dir/bin/get_version)
 
 docker build -f ci/publish-splinter-crates -t publish-splinter-crates ci/
+sed -i'' -e "0,/version.*$/ s/version.*$/version\ =\ \"$REPO_VERSION\"/" libsplinter/Cargo.toml
```
vaporos, line 22: Shouldn't this be the same already? If so, it would be better here just to confirm that they are the same and error.

rbuysse: I have it replacing the value on the assumption that at some point we'll want to publish from an untagged commit.

Having it replace also guards against someone accidentally running this and publishing a crate from an untagged commit with a non-dev version number.

rbuysse: I dropped the changes for changing crate versions. This can be revisited once the problem is better understood.

PR #837, opened 2020-05-14
 Improve PeerManager handling of duplicate connections

agunde406:
This PR improves the handling the case where both
peers are simultaneously trying to connect.

The peers need to both choose to close the same connection.
To handle this, the peers will compare their identities.
The peer with the greater id will close the incoming connection
while the peer with the lesser id will close the outbound
connection.

The PR also fixes outbound reconnection and improves
logging to include the peer_id.

### DIFF
--- a/libsplinter/src/network/peer_manager/mod.rs
+++ b/libsplinter/src/network/peer_manager/mod.rs
@@ -150,6 +150,7 @@ pub struct PeerManager {
     shutdown_handle: Option<ShutdownHandle>,
     max_retry_attempts: u64,
     retry_interval: u64,
+    identity: String,
 }
 
 impl PeerManager {
@@ -157,6 +158,7 @@ impl PeerManager {
         connector: Connector,
         max_retry_attempts: Option<u64>,
         retry_interval: Option<u64>,
+        identity: String,
     ) -> Self {
         PeerManager {
             connection_manager_connector: connector,
@@ -165,6 +167,7 @@ impl PeerManager {
             shutdown_handle: None,
             max_retry_attempts: max_retry_attempts.unwrap_or(DEFAULT_MAXIMUM_RETRY_ATTEMPTS),
             retry_interval: retry_interval.unwrap_or(DEFAULT_PACEMAKER_INTERVAL),
+            identity,
         }
     }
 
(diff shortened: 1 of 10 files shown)


libsplinter/src/network/peer_manager/mod.rs
```diff
@@ -667,134 +672,239 @@ fn handle_notifications(
...
-                            "Unable to remove connection for {}: {}",
-                            peer_metadata.active_endpoint, err
-                        );
+                info!(
+                    "Disconnected peer {} connected via {}",
+                    peer_metadata.id, endpoint
+                );
+            }
+            PeerStatus::Connected => {
+                // Compare identities, if local identity is , close incoming connection
```
jsmitchell, line 808: if local identity is...

libsplinter/src/network/peer_manager/mod.rs
```diff
@@ -667,134 +672,239 @@ fn handle_notifications(
...
-                    // tell subscribers this Peer is currently disconnected
-                    let notification = PeerManagerNotification::Disconnected {
-                        peer: peer_metadata.id.to_string(),
-                    };
-
-                    subscribers.retain(|sender| sender.send(notification.clone()).is_ok());
-                    error!(
-                        "Peer {} (via {}) presented a mismatched identity {}",
-                        identity, endpoint, peer_metadata.id
+                info!("Adding inbound connection to pending peer: {}", identity);
```
jsmitchell, line 712: Be consistent with capitalization of the status describing the peer (compare with line 707)

libsplinter/src/network/peer_manager/mod.rs
agunde406: fixed, reworded

libsplinter/src/network/peer_manager/mod.rs
agunde406: fixed

libsplinter/src/network/peer_manager/peer_map.rs
```diff
@@ -20,7 +20,7 @@ use crate::collections::BiHashMap;
 use super::error::PeerUpdateError;
 
 // Default intial value for how long to wait before retrying a peers endpoints
-const INITIAL_RETRY_FREQUENCY: u64 = 10;
+pub const INITIAL_RETRY_FREQUENCY: u64 = 10;
```
peterschwarz, line 23: Making this pub has a bit of a code smell to it.  I think that either: the `PeerMap` should be initialized with this value (and it not being a constant) or the place the place where this is being set in the peer manager, should use a `replace` method, that will re-initialize those internal values.

agunde406: Ah yep i forgot to change that. I am going to have it passed into PeerMap.

agunde406: Added a commit for this

PR #846, opened 2020-05-15
Await response from submit endpoint

shannynalayna:
Waits for the response from the /submit endpoint to more accurately
display error messages that may occur.

To test: Run the gameroom example, then attempt submitting a gameroom proposal with the local node selected. (So, if you're on Acme's UI, select the 'Acme' node in the dropdown and vice-versa for bubba's UI) This should then display the error rather than a success message, as it previously did.

### DIFF
--- a/examples/gameroom/gameroom-app/src/store/api.ts
+++ b/examples/gameroom/gameroom-app/src/store/api.ts
@@ -62,6 +62,7 @@ async function http(
 ): Promise<string> {
   return new Promise((resolve, reject) => {
     const request = new XMLHttpRequest();
+    request.responseType = 'json';
     request.open(method, `/api${url}`);
     if (headerFn) {
       headerFn(request);
@@ -70,7 +71,7 @@ async function http(
       if (request.status >= 200 && request.status < 300) {
         resolve(request.response);
       } else {
-        console.error(request);
+        console.error(request.response.message);
         if (request.status >= 400 && request.status < 500) {
           reject('Failed to send request. Contact the administrator for help.');
         } else {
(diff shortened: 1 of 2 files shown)


examples/gameroom/gameroom-app/src/scss/components/_toast.scss
```diff
@@ -18,7 +18,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  height: 3rem;
+  height: 4rem;
```
vaporos, line 21: These .scss seem unrelated to the fix

shannynalayna: It's to make the toast big enough to hold the error message, otherwise it was being clipped. The scroll change, in addition, is to ensure that the error message is not too big for the toast itself.

shannynalayna: I added a bit to the commit message to explain this change.

examples/gameroom/gameroom-app/src/store/modules/nodes.ts
```diff
@@ -30,9 +30,10 @@ const getters = {
 };
 
 const actions = {
-  async listNodes({ commit }: any) {
-    const nodes = await listNodes();
-    commit('setNodes', nodes);
+  async listNodesWithoutLocal({ commit }: any, localBrand: string) {
```
agunde406, line 33: Was this change supposed to be included in this commit?

shannynalayna: Oh no, it wasn't. Think I accidentally scooped up some unrelated changes when I pushed to update the commit message. Fixed now!

PR #851, opened 2020-05-18
Update peers created by only an endpoint, to be treated like a normal peer

agunde406:
add_unidentified_peer now returns an EndpointPeerRef that will act like a normal PeerRef, sending a remove request on drop. On successful connection the peer is added to the peer map and treated like a normal peer.

### DIFF
--- a/libsplinter/src/peer/mod.rs
+++ b/libsplinter/src/peer/mod.rs
@@ -1013,7 +1013,7 @@ fn handle_inbound_connection(
             }
         }
     } else {
-        debug!("Adding unrefrenced peer for {}", identity);
+        debug!("Adding peer with id: {}", identity);
 
         if let Some(old_peer) = unreferenced_peers.peers.insert(
             identity,
@@ -1022,8 +1022,8 @@ fn handle_inbound_connection(
                 endpoint: endpoint.to_string(),
             },
         ) {
-            debug!("Removing old unreferenced peer connection");
             if old_peer.endpoint != endpoint {
+                debug!("Removing old peer connection for {}", old_peer.endpoint);
                 if let Err(err) = connector.remove_connection(&old_peer.endpoint) {
                     error!("Unable to clean up old connection: {}", err);
                 }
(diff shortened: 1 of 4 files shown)


libsplinter/src/peer/mod.rs
```diff
@@ -524,8 +524,40 @@ fn add_peer(
...
+                            PeerRefAddError::AddError(format!(
+                                "Unable to update peer {}:{}",
+                                peer_id, err
+                            ))
+                        })?
+                    } else {
+                        // remove ref we just added
+                        ref_map.remove_ref(&peer_id);
+                        return Err(PeerRefAddError::AddError(format!(
+                            "Mismatch betwen exisiting and requested peer endpoints: {:?} does not \
```
vaporos, line 556: Typo - 'exisiting'

libsplinter/src/peer/mod.rs
```diff
@@ -524,8 +524,40 @@ fn add_peer(
 
     // if this is not a new peer, return success
     if new_ref_count > 1 {
-        let peer_ref = PeerRef::new(peer_id, peer_remover.clone());
-        return Ok(peer_ref);
+        if let Some(mut peer_metadata) = peers.get_by_peer_id(&peer_id).cloned() {
+            if peer_metadata.endpoints.len() == 1 && endpoints.len() > 1 {
+                // this will always be true
+                if let Some(endpoint) = peer_metadata.endpoints.get(0) {
```
vaporos, line 541: Should add in an error clause for the else then, so if code changes make it not true, its obvious.

libsplinter/src/peer/mod.rs
agunde406: fixed

libsplinter/src/peer/mod.rs
agunde406: added

libsplinter/src/peer/mod.rs
```diff
@@ -524,8 +524,52 @@ fn add_peer(
...
+                    } else {
+                        // remove ref we just added
+                        ref_map.remove_ref(&peer_id);
+                        return Err(PeerRefAddError::AddError(format!(
+                            "Mismatch betwen existing and requested peer endpoints: {:?} does not \
+                            contain {}",
+                            endpoints, endpoint
+                        )));
+                    }
+                }
```
vaporos, line 574: should have "else" here

PR #865, opened 2020-05-22
Scabbard cli fixes

ltseeley:
Some fixes for the stabilizing the scabbard CLI's features

### DIFF
--- a/services/scabbard/cli/src/main.rs
+++ b/services/scabbard/cli/src/main.rs
@@ -86,6 +86,7 @@ fn run() -> Result<(), CliError> {
         app = app.subcommand(
             SubCommand::with_name("contract")
                 .about("List, show, or upload a Sabre smart contract")
+                .setting(AppSettings::SubcommandRequiredElseHelp)
                 .subcommand(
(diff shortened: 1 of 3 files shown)


services/scabbard/cli/src/main.rs
```diff
@@ -86,25 +86,24 @@ fn run() -> Result<(), CliError> {
...
-                                    "The name and version requirement of the .scar to upload, in \
-                                     the format 'name:version_req'. The contract name must not \
-                                     include underscores, '_'. The version requirement can \
-                                     be any valid semver requirement string.",
+                                    "Name and version requirement of the smart contract archive \
+                                     (scar) file to upload",
                                 )
                                 .required(true),
-                            Arg::with_name("path")
+                            Arg::with_name("paths")
```
vaporos, line 100: "path" singular is more common for a list of directories to search (for example, $PATH, $LD_LIBRARY_PATH, etc.)

ltseeley: Dropped the commit

PR #868, opened 2020-05-22
Change RefMap so it returns an Result instead of panic!ing

agunde406:
Also makes PeerManager configurable to either return an error if refmap error or panic if strict_ref_map is set. Also adds env variable to splinterd to configure the PeerManager, SPLINTER_STRICT_REF_COUNT.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -797,7 +797,8 @@ mod tests {
             .start()
             .expect("Unable to start Connection Manager");
         let connector = cm.connector();
-        let mut peer_manager = PeerManager::new(connector, None, Some(1), "test-node".to_string());
+        let mut peer_manager =
+            PeerManager::new(connector, None, Some(1), "test-node".to_string(), true);
         let peer_connector = peer_manager.start().expect("Cannot start PeerManager");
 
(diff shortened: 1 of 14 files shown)


libsplinter/src/peer/mod.rs
```diff
@@ -399,6 +405,9 @@ impl ShutdownHandle {
     }
 }
 
+// Allow clippy errors for too_many_arguments. The arguments are required
+// to avoid needing a lock in the PeerManager.
+#[allow(clippy::too_many_arguments)]
```
peterschwarz, line 410: Not for this PR, but time to introduce a PeerManagerBuilder

libsplinter/src/admin/service/mod.rs
```diff
@@ -797,7 +797,8 @@ mod tests {
             .start()
             .expect("Unable to start Connection Manager");
         let connector = cm.connector();
-        let mut peer_manager = PeerManager::new(connector, None, Some(1), "test-node".to_string());
+        let mut peer_manager =
```
vaporos, line 800: All the test should fail if ref counting is violated, so I think these should all be true?

agunde406: whoops, yep you are correct

PR #883, opened 2020-06-01
Implement Vec::from for ConnectionMatrixEnvelope

agunde406:
This commit also deprecates ConnectionMatrixEnvelope
take_payload() because it is similar to From implementation
but From is cleaner.

### DIFF
--- a/libsplinter/src/mesh/mod.rs
+++ b/libsplinter/src/mesh/mod.rs
@@ -169,13 +169,13 @@ impl Mesh {
     /// Send the envelope on the mesh.
     ///
     /// This is a convenience function and is equivalent to
-    /// `mesh.outgoing(envelope.id()).send(envelope.take_payload())`.
+    /// `mesh.outgoing(envelope.id()).send(Vec::from(envelope))`.
     pub fn send(&self, envelope: Envelope) -> Result<(), SendError> {
         let state = &self.state.read().map_err(|_| SendError::PoisonedLock)?;
         let id = envelope.id().to_string();
         if let Some(mesh_id) = state.unique_ids.get_by_key(&id) {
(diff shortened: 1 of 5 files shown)


libsplinter/src/mesh/mod.rs
```diff
@@ -169,13 +169,13 @@ impl Mesh {
...
     /// This is a convenience function and is equivalent to
-    /// `mesh.outgoing(envelope.id()).send(envelope.take_payload())`.
+    /// `mesh.outgoing(envelope.id()).send(Vec::from(envelope))`.
     pub fn send(&self, envelope: Envelope) -> Result<(), SendError> {
         let state = &self.state.read().map_err(|_| SendError::PoisonedLock)?;
         let id = envelope.id().to_string();
         if let Some(mesh_id) = state.unique_ids.get_by_key(&id) {
             match state.outgoings.get(mesh_id) {
-                Some(ref outgoing) => match outgoing.send(envelope.take_payload()) {
+                Some(ref outgoing) => match outgoing.send(Vec::from(envelope)) {
```
peterschwarz, line 178: `envelope.into()` should work just as well, here

libsplinter/src/transport/matrix.rs
```diff
@@ -72,11 +72,18 @@ impl ConnectionMatrixEnvelope {
     }
 
     /// Returns the bytes of the payload while consuming the `ConnectionMatrixEnvelope`
+    #[deprecated(since = "0.3.19", note = "Please use Vec::from instead")]
     pub fn take_payload(self) -> Vec<u8> {
```
peterschwarz, line 76: Interestingly enough, I am no more inclined to think this method should still exist but be called `into_inner`, which is also common for this style of transformation in the standard library.

PR #888, opened 2020-06-01
Set CLI default REST API port to 8080

peterschwarz:
The default REST API port which is used if no URL is specified was set
to 8085.  This value is an older default and does not match the default
value set by splinterd.

This commit corrects the port to 8080.


vaporos:
This will require an entry in the upgrading doc.

vaporos:
We are waiting to merge this for the moment because its an API change.

PR #907, opened 2020-06-16
BACKPORT: Add 10s sleep during crates publish

rbuysse:
Cargo's index updates are not instantaneous, so a brief sleep is necessary
to ensure that the newly published splinter version is available for use by
libscabbard.

More information about Cargo's index udpates can be found at
https://internals.rust-lang.org/t/changes-to-how-crates-io-handles-index-updates/9608


rbuysse:
Added to master with #897.

PR #908, opened 2020-06-17
 Update subscribing to PeerManagerNotfications to take a sender 

agunde406:
This commit deprecates subscribe() in favor of subscribe_sender
which takes a Sender<T> which can be used to get a
PeerManagerNotfication and adds the ability to convert the
notification to another type. This change updates the PeerManager
to match the ConnectionManager.

Also adds an unsubscribe option to PeerManagerConnector so a
subscriber can unsubscribe from notifications.

Also removes Clone and PartialEq from PeerManager Messages.
These decorators were not required and caused issues with the
adding of the new subscriber messages.

### DIFF
--- a/libsplinter/src/peer/connector.rs
+++ b/libsplinter/src/peer/connector.rs
@@ -20,7 +20,7 @@ use super::error::{
     PeerConnectionIdError, PeerListError, PeerLookupError, PeerManagerError, PeerRefAddError,
     PeerRefRemoveError, PeerUnknownAddError,
 };
-use super::notification::PeerNotificationIter;
+use super::notification::{PeerManagerNotification, PeerNotificationIter, SubscriberId};
 use super::{EndpointPeerRef, PeerRef};
 use super::{PeerManagerMessage, PeerManagerRequest};
 
@@ -194,6 +194,7 @@ impl PeerManagerConnector {
     ///
     /// Returns a `PeerNotificationIter` that can be used to receive notifications about connected
     /// and disconnected peers
+    #[deprecated(since = "0.5.1", note = "please use `subscribe_sender` instead")]
     pub fn subscribe(&self) -> Result<PeerNotificationIter, PeerManagerError> {
         let (send, recv) = channel();
         match self.sender.send(PeerManagerMessage::Subscribe(send)) {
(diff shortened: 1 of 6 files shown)


libsplinter/src/peer/connector.rs
```diff
@@ -203,6 +204,66 @@ impl PeerManagerConnector {
...
     }
+
+    /// Subscribe to notifications for peer events.
+    ///
+    /// `PeerManagerNotification` instances will be transformed via type `T`'s implementation
+    /// of `From<PeerManagerNotification>` and passed to the given sender.
+    ///
+    /// # Returns
+    ///
+    /// The subscriber id that can be used for unsubscribing the given sender.
```
ltseeley, line 215: s/id/ID/

libsplinter/src/peer/connector.rs
```diff
@@ -203,6 +204,66 @@ impl PeerManagerConnector {
...
+            .map_err(|_| {
+                PeerManagerError::SendMessageError("The peer manager is no longer running".into())
+            })?;
+
+        recv.recv().map_err(|_| {
+            PeerManagerError::SendMessageError("The peer manager is no longer running".into())
+        })?
+    }
+
+    /// Unsubscribe to `PeerManagerNotification`.
```
ltseeley, line 245: s/to/from/

libsplinter/src/peer/notification.rs
```diff
@@ -61,6 +63,51 @@ impl Iterator for PeerNotificationIter {
...
+        let mut failures = vec![];
+        for (id, callback) in self.subscribers.iter() {
+            if let Err(err) = (*callback)(notification.clone()) {
+                failures.push(*id);
+                debug!("Dropping subscriber ({}): {}", id, err);
+            }
+        }
+
+        for id in failures {
+            self.subscribers.remove(&id);
```
ltseeley, line 94: Why can't this be done in the `if let Err...`?

libsplinter/src/peer/connector.rs
agunde406: fixed

libsplinter/src/peer/connector.rs
agunde406: fixed

libsplinter/src/peer/notification.rs
agunde406: An immutable borrow occurs above

PR #915, opened 2020-07-06
Replace ShutdownHandler with ShutdownSignaler 

agunde406:
This is in an effort to match the connection manager.

### DIFF
--- a/libsplinter/src/peer/mod.rs
+++ b/libsplinter/src/peer/mod.rs
@@ -128,7 +128,7 @@ pub(crate) enum PeerManagerRequest {
 pub struct PeerManager {
     join_handle: thread::JoinHandle<()>,
     sender: Sender<PeerManagerMessage>,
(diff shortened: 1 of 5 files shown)


libsplinter/src/peer/mod.rs
```diff
@@ -343,20 +349,21 @@ impl PeerManager {
...
+            shutdown_signaler,
             join_handle,
             sender,
         })
     }
 }
 
-/// Handles sending a shutdown message to the `PeerManager` and the `Pacemaker`
+/// Handles sending a shutdown message to the `PeerManager` and the `Pacemaker`.
+/// Deprecated, use ShutdownSignaler instead.
```
ltseeley, line 370: Can the `deprecated` attribute be added to a struct as well?

agunde406: It can be, but then it always raises the deprecation warning because of the Impl then uses the struct.

PR #918, opened 2020-07-16
Add SQLite support for Biome

ltseeley:


libsplinter/src/biome/key_management/store/diesel/mod.rs
```diff
@@ -29,24 +30,76 @@ use operations::{
     KeyStoreOperations,
 };
 
+/// TODO: remove mentions of postgres in comments
```
vaporos, line 33: TODO

libsplinter/src/store/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            let pool = pool_builder.build(connection_manager).map_err(|err| {
+                StoreFactoryCreationError(format!("Failed to build connection pool: {}", err))
+            })?;
+            Ok(Box::new(sqlite::SqliteStoreFactory::new(pool)))
+        }
+    }
+}
+
+pub enum StorageLocation {
```
vaporos, line 78: Don't like this enum name; however, I think it would be better to get rid of this enum completely and just match on the appropriate substring in the url.

libsplinter/src/store/sqlite.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use super::StoreFactory;
+
+/// A `StoryFactory` backed by a SQLite database.
+pub struct SqliteStoreFactory {
+    pool: Pool<ConnectionManager<SqliteConnection>>,
+}
+
+impl SqliteStoreFactory {
+    pub fn new(pool: Pool<ConnectionManager<SqliteConnection>>) -> Self {
```
vaporos, line 28: If we are concerned about this diesel/r2d2 leakage, we could provide ::new() and ::new_with_pool() and then protect ::new_path_pool() with a feature flag "expose-diesel-api" (not that exact feature name though).

vaporos: Maybe sqlite-expose-diesel-pool, sqlite-expose-external-apis, or sqlite-pool-integration

vaporos: If we took the approach of hiding pool-integration behind a more specific feature, we could probably also get away with DieselStoreFactore<C> where C: Connection and only expose DieselStoreFactory if the feature is enabled.

libsplinter/src/store/mod.rs
ltseeley: String matching causes a lot of complexity here. Which storage types are supported is determined by compilation flags; string matching forces us to check if the storage type is supported and handle any errors that may arise with that usage. With an enum, the compiler removes any possibility for errors; you can only use the enum variants that match your compilation options.

This is made even more complicated by SQLite supporting a variety of values (file paths, URIs, the `:memory:` string). It also assumes that every DB we implement support for can be identified by just the string passed in (which may not always be the case).

libsplinter/src/biome/key_management/store/diesel/mod.rs
ltseeley: Removed (and removed the postgres-specific language of the comments)

libsplinter/src/store/sqlite.rs
ltseeley: I don't think we can get away with a generic `DieselStoreFactore<C: Connection>`, because the stores don't have generic implementations for `C: Connection`; they have implementations for  `PgConnection` and `SqliteConnection`. The compiler needs to know which type of connection is to pick the right trait implementations for the stores.

As far as the diesel/r2d2 API leakage, I'm not personally concerned about it as long as we re-export the diesel/r2d2 elements that are exposed (to effectively pin them to the same version splinter uses).

libsplinter/src/store/mod.rs
vaporos: We still need to design the CLI arguments (for daemon and CLIs (at least migration CLI)) -- I think we should reduce them as much as possible (if we have to specify a URL and the database type every time, we've failed). That will drive what we need here, though we could have a String-to-URL-enum function to partition the "unknown database" error a bit.

For now, I'd at least like to see the enum with a better name. How about Url or DatabaseUrl or DatabaseUri.

splinterd/src/daemon.rs
```diff
@@ -699,29 +693,30 @@ fn start_health_service(
...
+fn build_biome_routes(db_url: String) -> Result<BiomeRestResourceManager, StartError> {
     info!("Adding biome routes");
-    let connection_pool: ConnectionPool =
-        database::ConnectionPool::new_pg(db_url).map_err(|err| {
-            StartError::RestApiError(format!(
-                "Unable to connect to the Splinter database: {}",
-                err
-            ))
-        })?;
+    let storage_location = if db_url.starts_with("postgres://") {
```
peterschwarz, line 698: Not sure I see the value of exposing this here, or at least, having a `StorageLocation::from_str` function.  That should return an error if it's an unsupported value. Both the `from_str`function and the create_store_factory can have feature guards for the available store backends

libsplinter/src/store/mod.rs
peterschwarz: What happens when you compile this with just store-factory?  Should there be a Memory option that's always available?

ltseeley: Updated enum name to `ConnectionUri`

libsplinter/src/store/sqlite.rs
peterschwarz: Do these constructors need to be crate public?

ltseeley: They do if we want to allow a single connection pool to be used for stores that aren't internal to Splinter, which is a requirement per @vaporos.

libsplinter/src/store/mod.rs
ltseeley: @peterschwarz added the memory store factory (and variant of the enum)

splinterd/src/daemon.rs
ltseeley: Added a string parsing method (implementation of the `FromStr` trait).

libsplinter/src/biome/migrations/error.rs
```diff
@@ -16,19 +16,19 @@ use std::error::Error;
 use std::fmt;
 
 #[derive(Debug)]
-pub struct ConnectionError {
+pub struct MigrationError {
```
peterschwarz, line 19: This is part of the public stable API.  You could create a type alias for the old name and deprecate it.

ltseeley: Updated

PR #921, opened 2020-07-16
Use gameroom circuit template in Gameroom daemon

shannynalayna:
Updates the `propose_gameroom` REST API endpoint function to use a
CircuitCreateTemplate to generate the CreateCircuit admin message
used to propose a gameroom.

This also updates the module for ApplicationMetadata to remove the
now unused functions that were only used in the `propose_gameroom`
function. Also removes an error variant that is no longer used.

In terms of testing this PR, the best approach would probably be to rebuild `gameroomd` and attempt to submit a gameroom through the UI, which should end successfully on each node. :)


ltseeley:
Tested

PR #925, opened 2020-07-21
`circuit-template` doc additions and small clean-up

shannynalayna:
Adds rust documentation to the `circuit-template` modules, as well as updates the template CLI man pages to mention the default circuit template directory location. 

Also makes a small update to the `list_available_templates` method, to simply log an error and keep going if it encounters an issue with an individual file. If it is unable to read from the directory itself, an error is returned.

### DIFF
--- a/cli/man/splinter-circuit-template-arguments.1.md
+++ b/cli/man/splinter-circuit-template-arguments.1.md
@@ -22,6 +22,9 @@ Circuit templates help simplify the process of creating new circuits with the
 building a circuit from the template. This command lists the arguments that are
 defined in the specified circuit template.
 
+All available templates are located in the circuit templates directory,
+`/usr/share/splinter/circuit-templates`.
+
 FLAGS
 =====
 `-h`, `--help`
@@ -41,7 +44,9 @@ FLAGS
 ARGUMENTS
 =========
 `TEMPLATE-NAME`
-: Circuit template that contains the arguments of interest.
+: Name of the circuit template containing the arguments of interest. The
+  template file must exist in the circuit template directory,
+    `/usr/share/splinter/circuit-templates`.
 
 EXAMPLES
 ========
--- a/libsplinter/src/circuit/template/error.rs
+++ b/libsplinter/src/circuit/template/error.rs
@@ -12,13 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! An error wrapper for the errors generated throughout the module.
+
+/// General error wrapper
 #[derive(Debug)]
 pub struct CircuitTemplateError {
+    /// Human-readable description of the action attempted, which caused the error.
     context: String,
+    /// Optional field which provides a `source` error, in case an exception must be handled.
     source: Option<Box<dyn std::error::Error>>,
 }
 
 impl CircuitTemplateError {
+    /// Builds a `CircuitTemplateError` with a `context`. This sets the `source` field to `None`.
     pub fn new(context: &str) -> Self {
         Self {
             context: context.into(),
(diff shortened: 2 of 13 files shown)


cli/man/splinter-circuit-template-arguments.1.md
```diff
@@ -41,7 +44,9 @@ FLAGS
 ARGUMENTS
 =========
 `TEMPLATE-NAME`
-: Circuit template that contains the arguments of interest.
+: Name of the circuit template containing the arguments of interest. The template
+  name is appended to the default circuit template directory,
+  `/usr/share/splinter/circuit-templates`, to locate the circuit template file.
```
chenette, line 49: Change the second sentence to "The template file must exist in the default circuit template directory, `/usr/share/splinter/circuit-templates`."

Also, if there's no way to change that directory, remove the word "default".

(This comment applies to all man pages with the TEMPLATE-NAME argument.)

cli/man/splinter-circuit-template-arguments.1.md
```diff
@@ -22,6 +22,9 @@ Circuit templates help simplify the process of creating new circuits with the
 building a circuit from the template. This command lists the arguments that are
 defined in the specified circuit template.
 
+All available templates are located in the default circuit templates directory,
```
chenette, line 25: If there's no way to use a different directory for the circuit templates, remove the word "default" (here and in all other places/man pages that reference this directory).

libsplinter/src/circuit/template/mod.rs
```diff
@@ -78,34 +78,33 @@ impl CircuitTemplateManager {
...
+                            Ok(name) => {
+                                available_templates
+                                    .push(name.trim_end_matches(".yaml").to_string());
+                            }
+                            Err(_) => {
+                                error!("Unable to read file name {}", self.path);
+                            }
+                        },
+                        Err(err) => {
+                            error!("Unable to read file {}", err);
```
ltseeley, line 117: These error logs should provide a bit more context, like "Unable to read circuit template {}". Otherwise they're pretty much unidentifiable errors.

libsplinter/src/circuit/template/mod.rs
```diff
@@ -78,34 +78,33 @@ impl CircuitTemplateManager {
...
-                        CircuitTemplateError::new(&format!(
-                            "Failed to read file name {}",
-                            self.path
-                        ))
-                    })?
-                    .trim_end_matches(".yaml")
-                    .to_string())
-            })
-            .collect::<Result<_, CircuitTemplateError>>()
+        let mut available_templates: Vec<String> = vec![];
```
ltseeley, line 102: not a necessary change, but this could be more be done idiomatically with `filter_map`.

libsplinter/src/circuit/template/error.rs
```diff
@@ -12,20 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! An error wrapper for the errors generated throughout the module.
+
+/// General error wrapper,
```
ltseeley, line 17: s/,//

cli/man/splinter-circuit-template-arguments.1.md
shannynalayna: I changed all of the TEMPLATE-NAME argument descriptions, and removed `default` before `circuit template directory` across all mentions in the man pages that I added

libsplinter/src/circuit/template/mod.rs
shannynalayna: Good point, fixed

libsplinter/src/circuit/template/mod.rs
shannynalayna: Also good point, changed it to `filter_map`

PR #928, opened 2020-07-24
Circuit template clean-up

shannynalayna:
* Removes the 'a:' & 'r:' prefix of template arguments. 
* Fixes a few typos and cleans up some error messages
* Changes the `into_builders` method to an `apply_to_builder` method which applies the template rules to the builder provided to the method. Also makes this change in the CLI circuit template implementation.

### DIFF
--- a/libsplinter/src/circuit/template/rules/mod.rs
+++ b/libsplinter/src/circuit/template/rules/mod.rs
@@ -137,15 +137,15 @@ impl TryFrom<v1::RuleArgument> for RuleArgument {
 }
 
 fn is_arg(key: &str) -> bool {
-    key.starts_with("$(a:")
+    key.starts_with("$(")
 }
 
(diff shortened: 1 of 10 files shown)


libsplinter/src/circuit/template/rules/mod.rs
```diff
@@ -63,9 +61,21 @@ impl Rules {
...
 
-        builders.set_create_circuit_builder(circuit_builder);
-        builders.set_service_builders(service_builders);
-        Ok(())
+        let mut services = vec![];
+        for service_builder in service_builders {
+            match service_builder.build() {
+                Ok(service) => services.push(service),
+                Err(err) => {
+                    debug!("Failed to build SplinterService: {}", err);
```
vaporos, line 69: The Error returned will likely end up in the log, so logging here would result in log duplication.

PR #929, opened 2020-07-28
Add database-backed registry

ltseeley:
Adds the `DieselRegistry`, a database-backed registry powered by Diesel.
This implementation is behind the new `registry-database` experimental
feature.


libsplinter/src/registry/diesel/operations/list_nodes.rs
```diff
@@ -0,0 +1,133 @@
...
+            splinter_nodes, splinter_nodes_endpoints, splinter_nodes_keys, splinter_nodes_metadata,
+        },
+    },
+    MetadataPredicate, Node, NodeBuilder, RegistryError,
+};
+
+use super::{exists_statements_from_metadata_predicates, RegistryOperations};
+
+pub(in crate::registry::diesel) trait RegistryListNodesOperation {
+    fn list_nodes(&self, predicates: &[MetadataPredicate]) -> Result<Vec<Node>, RegistryError>;
```
ltseeley, line 32: This currently loads all nodes into a `Vec` in memory. Diesel does not currently support iterators of query results. I found a GitHub issue about this (https://github.com/diesel-rs/diesel/issues/1087) with an example implementation of a buffering iterator, but it is not a complete implementation (it doesn't work with simultaneous connections without a lock). This seems like a difficult problem to solve, so I didn't solve it in this PR.

libsplinter/Cargo.toml
```diff
@@ -95,6 +95,7 @@ experimental = [
     "biome-notifications",
     "biome-user",
     "circuit-template",
+    "registry-database",
```
ltseeley, line 98: When stabilized, this feature should probably be removed and the feature guards replaced with `diesel` (at which point the combination of the `registry` and `postgres` and/or `sqlite` features will guard the implementation).

libsplinter/src/registry/mod.rs
```diff
@@ -245,9 +245,7 @@ pub trait RegistryReader: Send + Sync {
     /// # Arguments
     ///
     ///  * `identity` - The identity of the node.
-    fn has_node(&self, identity: &str) -> Result<bool, RegistryError> {
```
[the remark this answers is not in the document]
ltseeley: I forgot that you could do this with traits; I updated the commit to leave the default implementation and just override the implementations for local YAML and unified registries (since the overriding implementations are more efficient).

PR #930, opened 2020-07-29
Add AdminServiceStore Trait and YAML backend

agunde406:
This PR defines an API for managing the API for writing and
reading circuit state and pending circuit proposals. The
provided `AdminServiceStore` trait makes no assumptions about
the storage backend.

This module also include the structs that will be stored/returned by
the service store.

This is added behind an experimental feature at is not currently used.

The YAML backend is compatible with the current circuit/proposals state files.

### DIFF
--- a/libsplinter/src/admin/store/builders.rs
+++ b/libsplinter/src/admin/store/builders.rs
@@ -1,0 +1,1017 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Structs for building the native representations of circuits, proposals, services and nodes
+
+use crate::admin::messages::{is_valid_circuit_id, is_valid_service_id};
+
+use super::error::BuilderError;
+use super::{
+    AuthorizationType, Circuit, CircuitNode, CircuitProposal, DurabilityType, PersistenceType,
+    ProposalType, ProposedCircuit, ProposedNode, ProposedService, RouteType, Service, VoteRecord,
+};
+
+/// Builder to be used to build a `Circuit`
+#[derive(Default, Clone)]
+pub struct CircuitBuilder {
+    circuit_id: Option<String>,
+    roster: Option<Vec<Service>>,
+    members: Option<Vec<String>>,
+    auth: Option<AuthorizationType>,
+    persistence: Option<PersistenceType>,
+    durability: Option<DurabilityType>,
+    routes: Option<RouteType>,
+    circuit_management_type: Option<String>,
+}
+
+impl CircuitBuilder {
+    /// Creates a new circuit builder
+    pub fn new() -> Self {
+        CircuitBuilder::default()
+    }
+
+    /// Returns the circuit ID in the builder
+    pub fn circuit_id(&self) -> Option<String> {
+        self.circuit_id.clone()
+    }
+
+    /// Returns the list of services in the builder
+    pub fn roster(&self) -> Option<Vec<Service>> {
+        self.roster.clone()
+    }
+
+    /// Returns the list of node IDs in the builder
+    pub fn members(&self) -> Option<Vec<String>> {
+        self.members.clone()
+    }
+
+    /// Returns the authorization type in the builder
+    pub fn auth(&self) -> Option<AuthorizationType> {
+        self.auth.clone()
+    }
+
+    /// Returns the persistence type in the builder
+    pub fn persistence(&self) -> Option<PersistenceType> {
+        self.persistence.clone()
+    }
+
+    /// Returns the durability type in the builder
+    pub fn durability(&self) -> Option<DurabilityType> {
+        self.durability.clone()
+    }
+
+    /// Returns the routing type in the builder
+    pub fn routes(&self) -> Option<RouteType> {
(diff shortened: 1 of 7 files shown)


libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+        })?;
+
+        state.proposal_state = proposals_state;
+        Ok(())
+    }
+
+    /// Read circuit state from the circuit file path and cache the contents in the store and then
+    /// read circuit proposal state from the proposal file path and cache the contents in the
+    /// store
+    fn read_state(&mut self) -> Result<(), YamlAdminStoreError> {
```
ltseeley, line 299: Can this method just call `read_circuit_state` and `read_proposal_state` instead of duplicating them? Same for the write method.

libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+                .proposals
+                .contains_key(&proposal.circuit_id)
+            {
+                state
+                    .proposal_state
+                    .proposals
+                    .insert(proposal.circuit_id.to_string(), proposal);
+            } else {
+                return Err(AdminServiceStoreError::OperationError {
+                    context: format!("A proposal with id {} does not exists", proposal.circuit_id),
```
ltseeley, line 612: s/exists/exist/ (same in later messages)

libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+                        source: None,
+                    })?;
+
+            if state
+                .proposal_state
+                .proposals
+                .contains_key(&proposal.circuit_id)
+            {
+                return Err(AdminServiceStoreError::OperationError {
+                    context: format!("A proposal with id {} already exists", proposal.circuit_id),
```
ltseeley, line 565: "id" should be capitalized (also in later messages)

libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+        proposals.retain(|proposal| {
+            predicates
+                .iter()
+                .all(|predicate| predicate.apply_to_proposals(proposal))
+        });
+
+        Ok(Box::new(proposals.into_iter()))
+    }
+
+    ///
```
ltseeley, line 712: Missing the main part of the description

libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+
+        Ok(Box::new(circuits.into_iter()))
+    }
+
+    // /// Adds a circuit to the underlying storage based on the proposal that is already in state..
+    // /// Also includes the associated Services and Nodes. The associated circuit proposal for
+    // /// the circuit ID is also removed
+    // ///
+    // /// # Arguments
+    // ///
```
ltseeley, line 900: Commented out

libsplinter/src/admin/store/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+impl PartialOrd for ServiceId {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Predicate for filtering the lists of circuits and circuit proposals
+pub enum CircuitPredicate {
+    ManagmentType(String),
+    Members(Vec<String>),
```
ltseeley, line 272: We should probably change these variants to be more specific: `ManagementTypeEq` and `MembersIncludes`. This is more clear, and it gives us flexibility to add more variants in the future like `MembersExcludes` or `ManagementTypeNe`.

libsplinter/src/admin/store/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+    /// # Arguments
+    ///
+    ///  * `circuit` - The circuit to be added to state
+    ///  * `nodes` - A list of nodes that represent the circuit's members
+    ///
+    ///  Returns an error if a `Circuit` with the same ID already exists
+    fn add_circuit(
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
```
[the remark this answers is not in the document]
agunde406: No, nodes are stored separately because they can be referenced by multiple circuits. They are included in the circuit defined in a proposal but not in the actual circuit definition.

libsplinter/src/admin/store/yaml/mod.rs
agunde406: I had seperated them out like this so the state lock would be held during the required read/write. If we just called read_circuit_state/read_proposal_state the lock could be acquired inbetween.

libsplinter/src/admin/store/builders.rs
```diff
@@ -0,0 +1,1065 @@
...
+        let roster = self
+            .roster
+            .ok_or_else(|| CircuitBuilderError::MissingField("roster".to_string()))?;
+
+        let members = self
+            .members
+            .ok_or_else(|| CircuitBuilderError::MissingField("members".to_string()))?;
+
+        let auth = self.auth.unwrap_or_else(|| {
+            debug!(
```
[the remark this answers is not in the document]
agunde406: removing, they are not helpful at this level

libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -0,0 +1,1830 @@
...
+    CircuitPredicate, CircuitProposal, DurabilityType, PersistenceType, RouteType, Service,
+    ServiceId,
+};
+
+/// YAML file specific circuit definition. This circuit definition in the 0.4v YAML stores service
+/// arguments in a map format, which differs from the definition defined in the AdminServiceStore.
+/// To handle this, circuit needs to be converted to the correct format during read/write
+/// operations.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Eq)]
+struct YamlCircuit {
```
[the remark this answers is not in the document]
agunde406: I will move it to the bottom, above the tests

PR #939, opened 2020-08-07
Update CLI man page 'SEE ALSO' link

shannynalayna:
Updates the link in the 'SEE ALSO' section of the CLI man pages to point
to the Splinter website rather than the repository.


ltseeley:
I'm a little concerned that the version on these won't get updated when a 0.6 branch is created and master gets bumped to 0.7. @rbuysse is this something you'd be able to catch when we do the 0.6 release and bump the versions?

PR #945, opened 2020-08-25
Add database backend for AdminServiceStore

shannynalayna:
Implements the agnostic database backend for the AdminServiceStore. That includes a `DieselAdminServiceStore` and `DieselAdminServiceStoreOperations` structs which wrap a generic db connection, as well as operations implemented to implement the `AdminServiceStore` trait for the `DieselAdminServiceStore`.

### DIFF
--- a/libsplinter/src/admin/store/diesel/operations/add_circuit.rs
+++ b/libsplinter/src/admin/store/diesel/operations/add_circuit.rs
@@ -1,0 +1,290 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Provides the "add circuit" operation for the `DieselAdminServiceStore`.
+
+use std::collections::HashMap;
+
+use diesel::{dsl::insert_into, prelude::*};
+
+use super::AdminServiceStoreOperations;
+use crate::admin::store::{
+    diesel::{
+        models::{
+            CircuitMemberModel, CircuitModel, NodeEndpointModel, ServiceAllowedNodeModel,
+            ServiceArgumentModel, ServiceModel,
+        },
+        schema::{
+            circuit, circuit_member, node_endpoint, service, service_allowed_node, service_argument,
+        },
+    },
+    error::AdminServiceStoreError,
+    Circuit, CircuitNode,
+};
+
+pub(in crate::admin::store::diesel) trait AdminServiceStoreAddCircuitOperation {
+    fn add_circuit(
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
+    ) -> Result<(), AdminServiceStoreError>;
+}
+
+#[cfg(feature = "postgres")]
+impl<'a> AdminServiceStoreAddCircuitOperation
+    for AdminServiceStoreOperations<'a, diesel::pg::PgConnection>
+{
+    fn add_circuit(
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
+    ) -> Result<(), AdminServiceStoreError> {
+        self.conn.transaction::<(), _, _>(|| {
+            // Check if the circuit already exists in the `AdminServiceStore`, in which case
+            // an error is returned.
+            if circuit::table
+                .filter(circuit::circuit_id.eq(&circuit.id))
+                .first::<CircuitModel>(self.conn)
+                .optional()
+                .map_err(|err| AdminServiceStoreError::QueryError {
+                    context: String::from("Error occurred fetching Circuit"),
+                    source: Box::new(err),
+                })?
+                .is_some()
+            {
+                return Err(AdminServiceStoreError::OperationError {
+                    context: String::from("Circuit already exists in AdminServiceStore"),
+                    source: None,
+                });
+            }
+
+            // Create a `CircuitModel` from the `Circuit` to add to database
+            let circuit_model: CircuitModel = CircuitModel::from(&circuit);
+            insert_into(circuit::table)
+                .values(circuit_model)
+                .execute(self.conn)
+                .map_err(|err| AdminServiceStoreError::QueryError {
+                    context: String::from("Unable to insert Circuit"),
+                    source: Box::new(err),
+                })?;
+            // Create a list of circuit members from `nodes`
+            let circuit_members: Vec<CircuitMemberModel> = nodes
+                .iter()
+                .map(|node| CircuitMemberModel {
+                    circuit_id: circuit.id.clone(),
+                    node_id: node.id.clone(),
+                })
+                .collect();
+            insert_into(circuit_member::table)
+                .values(circuit_members)
+                .execute(self.conn)
+                .map_err(|err| AdminServiceStoreError::QueryError {
+                    context: String::from("Unable to insert Circuit members"),
+                    source: Box::new(err),
+                })?;
+            // Iterate over the list of `CircuitNodes` to extract the `node_id` and `endpoints`, to
+            // convert them into the `NodeEndpointModel`. Then, verify the `node_id` does not
+            // already have associated `node_endpoint` entries before inserting the list of
+            // `NodeEndpointModel`.
+            for (node_id, endpoints) in nodes
+                .iter()
+                .map(|node| {
+                    (
+                        node.id.clone(),
+                        node.endpoints
+                            .iter()
+                            .map(|endpoint| NodeEndpointModel {
+                                node_id: node.id.clone(),
+                                endpoint: endpoint.clone(),
+                            })
+                            .collect::<Vec<NodeEndpointModel>>(),
+                    )
+                })
+                .collect::<HashMap<String, Vec<NodeEndpointModel>>>()
+                .into_iter()
+            {
+                if let Some(0) = node_endpoint::table
+                    .filter(node_endpoint::node_id.eq(&node_id))
+                    .count()
+                    .first(self.conn)
+                    .optional()
+                    .map_err(|err| AdminServiceStoreError::QueryError {
+                        context: String::from("Error occurred counting CircuitNode endpoints"),
+                        source: Box::new(err),
+                    })?
+                {
+                    insert_into(node_endpoint::table)
+                        .values(endpoints)
+                        .execute(self.conn)
+                        .map_err(|err| AdminServiceStoreError::QueryError {
+                            context: String::from("Unable to insert Circuit members"),
+                            source: Box::new(err),
+                        })?;
+                }
+            }
+
+            // Build `Services` and all associated data from `circuit`
+            let circuit_services: Vec<ServiceModel> = Vec::from(&circuit);
+            insert_into(service::table)
+                .values(&circuit_services)
+                .execute(self.conn)
+                .map_err(|err| AdminServiceStoreError::QueryError {
+                    context: String::from("Unable to insert Services"),
+                    source: Box::new(err),
+                })?;
+            let service_argument: Vec<ServiceArgumentModel> = Vec::from(&circuit);
+            insert_into(service_argument::table)
+                .values(&service_argument)
+                .execute(self.conn)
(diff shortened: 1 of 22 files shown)


libsplinter/src/admin/store/diesel/schema.rs
```diff
@@ -132,3 +132,20 @@ table! {
         endpoint -> Text,
     }
 }
+
+allow_tables_to_appear_in_same_query!(
```
ltseeley, line 136: Are the proposal tables used in any of the same queries as the circuit tables?

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
```diff
@@ -0,0 +1,122 @@
...
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
+    ) -> Result<(), AdminServiceStoreError>;
+}
+
+impl<'a, C> AdminServiceStoreAddCircuitOperation for AdminServiceStoreOperations<'a, C>
+where
+    C: diesel::Connection,
+    <C as diesel::Connection>::Backend: diesel::backend::SupportsDefaultKeyword,
```
ltseeley, line 43: SQLite connections don't implement `SupportsDefaultKeyword`; for other stores, we've had to implement the "add" trait separately for SQLite and Postgres.

libsplinter/src/admin/store/diesel/operations/fetch_circuit.rs
```diff
@@ -0,0 +1,88 @@
...
+        models::{CircuitMemberModel, CircuitModel},
+        schema::{circuit, circuit_member},
+    },
+    error::AdminServiceStoreError,
+    AuthorizationType, Circuit, CircuitBuilder, DurabilityType, PersistenceType, RouteType,
+    Service,
+};
+
+pub(in crate::admin::store::diesel) trait AdminServiceStoreFetchCircuitOperation {
+    fn fetch_circuit(&self, circuit_id: &str) -> Result<Option<Circuit>, AdminServiceStoreError>;
```
ltseeley, line 32: Bit of a change-up, but I think we decided to go with `get_*` instead of `fetch_*` for stores.

shannynalayna: Ok, since that is an admin store API change, perhaps that should go in another PR? I don't think that change would fit with this db implementation (and this pr is big enough as-is)

libsplinter/src/admin/store/diesel/schema.rs
shannynalayna: No, I can separate them

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn add_circuit(
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
+    ) -> Result<(), AdminServiceStoreError> {
+        if circuit::table
+            .filter(circuit::circuit_id.eq(&circuit.id))
+            .first::<CircuitModel>(self.conn)
+            .map(Some)
+            .or_else(|err| if err == NotFound { Ok(None) } else { Err(err) })
```
ltseeley, line 53: This can be accomplished more succinctly with `.optional()`

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
```diff
@@ -0,0 +1,200 @@
...
+            .map_err(|err| AdminServiceStoreError::QueryError {
+                context: String::from("Error occurred fetching Circuit"),
+                source: Box::new(err),
+            })?
+            .is_some()
```
ltseeley, line 58: Instead of `if option.is_some() { ... }`, you can do `option.ok_or_else(|| { ... })`

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
```diff
@@ -0,0 +1,200 @@
...
+#[cfg(feature = "postgres")]
+impl<'a> AdminServiceStoreAddCircuitOperation
+    for AdminServiceStoreOperations<'a, diesel::pg::PgConnection>
+{
+    fn add_circuit(
+        &self,
+        circuit: Circuit,
+        nodes: Vec<CircuitNode>,
+    ) -> Result<(), AdminServiceStoreError> {
+        if circuit::table
```
ltseeley, line 49: This should probably be in the transaction closure below to ensure the whole database interaction is atomic. I just realized that the registry implementation performs this check outside of the transaction as well (I'll update that in my own PR).

libsplinter/src/admin/store/diesel/operations/add_proposal.rs
```diff
@@ -0,0 +1,249 @@
...
+            // Insert `members` of a `ProposedCircuit`
+            let proposed_members: Vec<ProposedNodeModel> = Vec::from(&proposal.circuit);
+            insert_into(proposed_node::table)
+                .values(proposed_members)
+                .execute(self.conn)
+                .map_err(|err| AdminServiceStoreError::QueryError {
+                    context: String::from("Unable to insert ProposedNode"),
+                    source: Box::new(err),
+                })?;
+            // Insert the node `endpoints` the proposed `members` of a `ProposedCircuit`
```
ltseeley, line 92: s/the/and/

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
```diff
@@ -0,0 +1,203 @@
...
+{
+    fn list_circuits(
+        &self,
+        predicates: &[CircuitPredicate],
+    ) -> Result<Box<dyn ExactSizeIterator<Item = Circuit>>, AdminServiceStoreError> {
+        // Collect the `Circuits` and members
+        let mut circuits: HashMap<String, CircuitModel> = HashMap::new();
+        let mut circuit_members: HashMap<String, Vec<String>> = HashMap::new();
+        circuit::table
+            .inner_join(
```
ltseeley, line 55: I'm not sure why these joins are necessary. It seems like it would be simpler to construct each of the HashMaps in their own queries. If it's not that simple, you should add some comments to explain why and how this code works; it's fairly difficult to understand.

libsplinter/src/admin/store/diesel/operations/update_circuit.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn update_circuit(&self, circuit: Circuit) -> Result<(), AdminServiceStoreError>;
+}
+
+#[cfg(feature = "postgres")]
+impl<'a> AdminServiceStoreUpdateCircuitOperation
+    for AdminServiceStoreOperations<'a, diesel::pg::PgConnection>
+{
+    fn update_circuit(&self, circuit: Circuit) -> Result<(), AdminServiceStoreError> {
+        // Check if a `Circuit` exists with the given `circuit_id`
+        if circuit::table
```
ltseeley, line 46: This should be in the transaction

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
shannynalayna: The joins are being used in situations where the tables being joined have mostly duplicate info (like the node tables, the endpoint table only has the endpoint uniquely in that table, otherwise it's the node id, which is present in the node table and the circuit id which is in the node and circuit tables). So, the joins are used in an attempt to cut down on the number of queries used to put the data into hash maps (which would have to be done anyways, regardless of the number of queries in order to sort them). I'll add some more documentation to explain what's happening throughout

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
shannynalayna: The point of the `if option.is_some()` is to check if the query returns a circuit (`is_some()`), in which case we need to return an error (because we're trying to add a circuit, but one already exists). So, it wouldn't make sense, in this scenario, to use `option.ok_or_else`.

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
```diff
@@ -0,0 +1,203 @@
...
+
+            ret_circuits.push(circuit_builder.build().map_err(|err| {
+                AdminServiceStoreError::OperationError {
+                    context: String::from("Unable to build Circuit"),
+                    source: Some(Box::new(err)),
+                }
+            })?);
+        }
+
+        ret_circuits.retain(|circuit| {
```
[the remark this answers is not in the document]
shannynalayna: Updated both list circuits and proposals to filter within the initial query rather than after

libsplinter/src/admin/store/diesel/operations/add_circuit.rs
ltseeley: Ah, I misread the code. 👍

libsplinter/src/admin/store/diesel/operations/remove_circuit.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl<'a, C> AdminServiceStoreRemoveCircuitOperation for AdminServiceStoreOperations<'a, C>
+where
+    C: diesel::Connection,
+    String: diesel::deserialize::FromSql<diesel::sql_types::Text, C::Backend>,
+    i64: diesel::deserialize::FromSql<diesel::sql_types::BigInt, C::Backend>,
+{
+    fn remove_circuit(&self, circuit_id: &str) -> Result<(), AdminServiceStoreError> {
+        // Verify the circuit attempting to be removed exists.
+        self.fetch_circuit(&circuit_id).and_then(|_| {
```
ltseeley, line 35: Because the fetch performs its own queries, this whole thing should be wrapped in a transaction (I need to make the same change for the registry's fetch operation).

shannynalayna: Updated!

ltseeley (approved):

ltseeley (approved):

ltseeley (approved):

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
```diff
@@ -0,0 +1,263 @@
...
+                // Store circuit IDs separately to make it easier to filter following queries
+                let circuit_ids: Vec<String> = circuits.keys().cloned().collect();
+
+                // Collect the `Circuit` members
+                let mut circuit_members: HashMap<String, Vec<String>> = HashMap::new();
+                for member in circuit_member::table
+                    .filter(circuit_member::circuit_id.eq_any(&circuit_ids))
+                    .load::<CircuitMemberModel>(self.conn)
+                    .map_err(|err| AdminServiceStoreError::QueryError {
+                        context: String::from("Unable to load Service information"),
```
agunde406, line 108: Should this say "Unable to load Member information"

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
```diff
@@ -0,0 +1,263 @@
...
+                        source: Box::new(err),
+                    })?
+                {
+                    if let Some(members) = circuit_members.get_mut(&member.node_id) {
```
agunde406, line 112: I think this should be using circuit_id not member.node_id, some for the insertions below.

libsplinter/src/admin/store/diesel/operations/fetch_proposal.rs
```diff
@@ -0,0 +1,283 @@
...
+
+            // Create HashMap of `service_id` to a `ProposedServiceBuilder` to collect
+            // `ProposedService` information
+            let mut proposed_services: HashMap<String, ProposedServiceBuilder> = HashMap::new();
+            // Create HashMap of `service_id` to the associated argument values
+            let mut arguments_map: HashMap<String, Vec<(String, String)>> = HashMap::new();
+            // Create HashMap of `service_id` to the associated allowed nodes
+            let mut allowed_nodes_map: HashMap<String, Vec<String>> = HashMap::new();
+            // Collect all 'proposed_service' entries and associated data using `inner_join`, as
+            // `proposed_service` has a one-to-many relationship to both `proposedservice_argument`
```
agunde406, line 147: "proposedservice_argument" -> "proposed_service_argument"

agunde406 (changes requested):

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
shannynalayna: Yep, updated!

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
shannynalayna: Yes, updated this error message!

agunde406 (approved):

ltseeley (approved):

PR #946, opened 2020-09-01
Add circuit template environment variable

shannynalayna:
Adds a circuit template environment variable to allow users to change the directory circuit templates are stored in. This also affects the circuit templates packaged within Gameroom and the Splinter CLI.

### DIFF
--- a/libsplinter/src/circuit/template/mod.rs
+++ b/libsplinter/src/circuit/template/mod.rs
@@ -25,10 +25,12 @@ mod rules;
 mod yaml_parser;
 
 use std::convert::TryFrom;
+use std::env;
 use std::path::Path;
 
 pub use error::CircuitTemplateError;
 
+use glob::glob;
 pub use rules::RuleArgument;
 use rules::Rules;
 
@@ -38,6 +40,8 @@ pub(self) use crate::admin::messages::{CreateCircuitBuilder, SplinterServiceBuil
 
 /// Default file location for circuit templates
 pub const DEFAULT_TEMPLATE_DIR: &str = "/usr/share/splinter/circuit-templates";
+/// Environment variable for file location for circuit templates
+pub const SPLINTER_CIRCUIT_TEMPLATE_PATH: &str = "SPLINTER_CIRCUIT_TEMPLATE_PATH";
 
 /// Manages circuit templates.
 ///
@@ -45,31 +49,32 @@ pub const DEFAULT_TEMPLATE_DIR: &str = "/usr/share/splinter/circuit-templates";
 /// list any availabe circuit templates found in the `path` of the `CircuitTemplateManager`.
 pub struct CircuitTemplateManager {
     /// Path of the directory containing the circuit template files.
-    path: String,
+    paths: Vec<String>,
 }
 
 impl Default for CircuitTemplateManager {
     /// Constructs a `CircuitTemplateManager` with the `DEFAULT_TEMPLATE_DIR`.
     fn default() -> Self {
         CircuitTemplateManager {
-            path: DEFAULT_TEMPLATE_DIR.to_string(),
+            paths: vec![DEFAULT_TEMPLATE_DIR.to_string()],
         }
     }
 }
 
 impl CircuitTemplateManager {
-    /// Constructs a `CircuitTemplateManager` with a custom `path` to the circuit templates.
-    pub fn new(path: &str) -> Result<CircuitTemplateManager, CircuitTemplateError> {
-        if !Path::new(&path).is_dir() {
-            Err(CircuitTemplateError::new(&format!(
-                "{} is not a valid directory",
-                path
-            )))
-        } else {
-            Ok(CircuitTemplateManager {
-                path: path.to_string(),
+    /// Constructs a `CircuitTemplateManager` with custom `paths` to the circuit templates.
+    pub fn new(paths: &[String]) -> CircuitTemplateManager {
+        let paths: Vec<String> = paths
+            .iter()
+            .filter_map(|path| {
+                if Path::new(&path).is_dir() {
+                    Some(path.to_string())
+                } else {
+                    None
+                }
             })
-        }
+            .collect();
+        CircuitTemplateManager { paths }
     }
 
     /// Loads the specified YAML circuit template file into a CircuitCreateTemplate.
@@ -78,7 +83,7 @@ impl CircuitTemplateManager {
     ///
     /// * `name` - file name indicating the circuit template to be loaded.
     pub fn load(&self, name: &str) -> Result<CircuitCreateTemplate, CircuitTemplateError> {
-        let path = format!("{}/{}.yaml", self.path, name);
+        let path = find_template(name, &self.paths)?;
         CircuitCreateTemplate::from_yaml_file(&path)
     }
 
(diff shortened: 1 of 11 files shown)


libsplinter/src/circuit/template/mod.rs
```diff
@@ -38,6 +38,8 @@ pub(self) use crate::admin::messages::{CreateCircuitBuilder, SplinterServiceBuil
 
 /// Default file location for circuit templates
 pub const DEFAULT_TEMPLATE_DIR: &str = "/usr/share/splinter/circuit-templates";
+/// Environment variable for file location for circuit templates
+pub const TEMPLATE_DIR_ENV: &str = "TEMPLATE_DIR";
```
vaporos, line 42: We should prefix this with SPLINTER_, and make it a path instead of a single directory. 

SPLINTER_CIRCUIT_TEMPLATE_PATH=dir1:dir2:dir3

libsplinter/src/circuit/template/yaml_parser/mod.rs
```diff
@@ -45,16 +45,36 @@ impl CircuitTemplate {
...
-        if !path.is_file() {
-            return Err(CircuitTemplateError::new(&format!(
-                "File does not exist or is inaccessible: {}",
-                file_path
-            )));
-        }
-        let file = File::open(path).map_err(|err| {
-            CircuitTemplateError::new_with_source("Error opening template file", err.into())
-        })?;
+        let custom_path = Path::new(file_path);
```
vaporos, line 48: Take a look at https://github.com/hyperledger/transact/blob/master/libtransact/src/contract/archive/mod.rs for an example of searching a path for a file.

vaporos (changes requested):

libsplinter/src/circuit/template/mod.rs
shannynalayna: Just to be clear here, you mean the `SPLINTER_CIRCUIT_TEMPLATE_PATH` would be something like `/usr/share/splinter` and then we would want to search there for the circuit template files? The second line formatting makes it look like you would be able to pass in multiple directories/paths like `SPLINTER_CIRCUIT_TEMPLATE_PATH=/usr/share/splinter:/tmp/lates:/foo/bar` and then all of those locations would be searched.

vaporos: Probably SPLINTER_CIRCUIT_TEMPLATE_PATH=/usr/share/splinter/circuit-templates:/tmp/lates:/foo/bar but otherwise yes, it should search the directories in the order given.

cli/packaging/ubuntu/postinst
```diff
@@ -12,19 +12,33 @@
...
 case "$1" in
     configure)
-
     # This value is present only during upgrades
     if [ -z "$2" ]; then
-        ln -s /usr/bin/splinter /usr/bin/splinter-cli
-
-        if [ ! -d /usr/share/splinter/circuit-templates ]; then
-          mkdir -p /usr/share/splinter/circuit-templates
+        if [ -z $SPLINTER_CIRCUIT_TEMPLATE_PATH ]
```
vaporos, line 30: This script should probably ignore SPLINTER_CIRCUIT_TEMPLATE_PATH and always install into the well-known directory; in any case, this code wouldn't work if SPLINTER_CIRCUIT_TEMPLATE_PATH had more than one directory specified.

libsplinter/src/circuit/template/mod.rs
```diff
@@ -38,38 +40,46 @@ pub(self) use crate::admin::messages::{CreateCircuitBuilder, SplinterServiceBuil
...
+            .iter()
+            .filter_map(|path| {
+                if Path::new(&path).is_dir() {
+                    Some(path.to_string())
+                } else {
+                    None
+                }
             })
+            .collect();
+        if let Ok(template_paths) = env::var(SPLINTER_CIRCUIT_TEMPLATE_PATH) {
```
vaporos, line 77: This should be pushed up a level outside CircuitTemplateManager; then paths passed in wouldn't need to be modified and the caller of new() is in full control of the behavior.

libsplinter/src/circuit/template/mod.rs
```diff
@@ -102,25 +112,30 @@ impl CircuitTemplateManager {
...
+            .iter()
+            .map(|path| {
+                let template_path = Path::new(path).join(&template_pattern);
+                let pattern_string = template_path.to_str().ok_or_else(|| {
+                    CircuitTemplateError::new(&String::from("Template path is not valid UTF-8"))
+                })?;
+                Ok(glob(pattern_string).map_err(|_| {
+                    CircuitTemplateError::new(&format!(
+                        "Cannot query path {:?} for pattern: {}",
+                        template_path, pattern_string
```
vaporos, line 127: You should be able to use template_path.display() here instead of converting it back to a string, avoiding the potential for the first CircuitTemplateError in this function.

libsplinter/src/circuit/template/mod.rs
```diff
@@ -364,6 +427,40 @@ rules:
...
+    ///    assert this is successfully loaded.
+    ///
+    /// Verifying the last 3 steps were done successfully verifies the `SPLINTER_CIRCUIT_TEMPLATE_PATH`
+    /// environment variable has been successfully set and used by the `CircuitTemplateManager` and
+    /// `CircuitCreateTemplate` modules to load a circuit template yaml file.
+    #[test]
+    fn test_template_from_env() {
+        let temp_dir = TempDir::new("test_builds_template_v1").unwrap();
+        let temp_dir = temp_dir.path().to_path_buf();
+        std::env::set_var("SPLINTER_CIRCUIT_TEMPLATE_PATH", &temp_dir);
```
vaporos, line 452: This test should take care to reset SPLINTER_CIRCUIT_TEMPLATE_PATH to its original value (maybe making this a hard function to write properly). Obviously this would conflict with any other test using SPLINTER_CIRCUIT_TEMPLATE_PATH.

cli/packaging/ubuntu/postinst
shannynalayna: That's fair, the script was initially meant to capture the environment variable assuming that the default directory wouldn't be used. Now, with the multiple directories, the default is always searched last so this commit is unnecessary - I'll remove it.

libsplinter/src/circuit/template/mod.rs
shannynalayna: I think it would make sense for part of this check to get moved into the `find_template` method, in the case that the function does not have any paths to search. Then, the `new` method is in full control of the caller (unless they don't provide any paths, in which case it'll still search the env var paths if set and the default)

libsplinter/src/circuit/template/mod.rs
```diff
@@ -364,6 +445,41 @@ rules:
...
+        let file_path = get_file_path(temp_dir);
+        write_yaml_file(&file_path, EXAMPLE_TEMPLATE_YAML);
+
+        let manager = CircuitTemplateManager::new(&[]);
+        let template = manager.load("example_template.yaml");
+        assert!(template.is_ok());
+        // Just pass the template name as the `SPLINTER_CIRCUIT_TEMPLATE_PATH` env var has been set
+        let template_v1 = CircuitCreateTemplate::from_yaml_file("example_template.yaml");
+        assert!(template_v1.is_ok());
+        std::env::remove_var("SPLINTER_CIRCUIT_TEMPLATE_PATH");
```
vaporos, line 480: It might have been set, so the value should be saved prior to setting it, then the value replaced here.

shannynalayna: Switched up this test to focus on testing the multiple directory functionality.

vaporos (dismissed):

rbuysse (approved):

vaporos (approved):

rbuysse (approved):

PR #953, opened 2020-09-30
Reorganize the structs used by the AdminServiceStore

agunde406:
This PR moves the structs into their own file such as`store::circuit::{Circuit, CircuitBuilder};` The structs are still made public from the `::store::*` with the use of `pub use`. With the move, it also adds getter methods that were previously unimplemented and updates the store implementations to use those instead of the fields directly. 

It also removes public fields from CircuitProposals and VoteRecord.

Finally removes the mutable method from CircuitProposal and replace it with a method to get a populated CircuitProposalBuilder that can be updated separately .


ltseeley (dismissed):

libsplinter/src/admin/store/circuit_proposal.rs
```diff
@@ -74,6 +69,17 @@ impl CircuitProposal {
     pub fn requester_node_id(&self) -> &str {
         &self.requester_node_id
     }
+
+    pub fn builder(&self) -> CircuitProposalBuilder {
```
peterschwarz, line 73: should this be `into_builder` and consume the struct?

agunde406: The builder was added originally because @vaporos  had mentioned that maybe CircuitProposal should be read only. Consuming the struct felt counter to that.

peterschwarz (approved):

ltseeley (approved):

PR #954, opened 2020-10-01
Admin and routing store updates

ltseeley:
Some updates in support of stabilizing the `admin-service-store` and `routing-table` features.

### DIFF
--- a/libsplinter/src/admin/store/mod.rs
+++ b/libsplinter/src/admin/store/mod.rs
@@ -12,36 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Defines an API for managing the API for writing and reading circuit state and pending circuit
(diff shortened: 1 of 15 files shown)


libsplinter/src/admin/store/mod.rs
```diff
@@ -336,10 +321,10 @@ impl CircuitPredicate {
     }
 }
 
-/// Defines methods for CRUD operations and fetching and listing circuits, proposals, nodes and
-/// services without defining a storage strategy
+/// Interface for performing CRUD operations on circuits, proposals, nodes, and services that does
```
peterschwarz, line 324: Not really sure that the "that does not define a storage strategy" part of the comment is necssary, as this is what traits impl by definition.

ltseeley: Removed that bit (also in the mod doc)

agunde406 (approved):

peterschwarz (approved):

PR #956, opened 2020-10-02
Initial Splinter CLI login subcommand

peterschwarz:
## Update

To test, using Github oauth

```sh
export OAUTH2_PROVIDER_ID=Github
export OAUTH2_CLIENT_ID="<your Github client id>"
export OAUTH2_CLIENT_SECRET="<your Github client secret>"
export OAUTH2_AUTH_URL="https://github.com/login/oauth/authorize"
export OAUTH2_TOKEN_URL="https://github.com/login/oauth/access_token"
export OAUTH2_SCOPES="user:email"
```

and 

```sh
$ cargo run --manifest-path cli/Cargo.toml --features experimental login
```

You should have a file at `$HOME/.splinter/auth_token` which contains the token values from the provider.

For Google,

```sh
export OAUTH2_PROVIDER_ID=Google
export OAUTH2_CLIENT_ID="<your Google client id>"
export OAUTH2_CLIENT_SECRET="<your Google client secret>"
export OAUTH2_AUTH_URL="https://accounts.google.com/o/oauth2/v2/auth"
export OAUTH2_TOKEN_URL="https://oauth2.googleapis.com/token"
export OAUTH2_SCOPES="openid,email,profile"
```

(This probably requires a clean build between changes to these env variables.)

## Note

The compile-time configuration will be replaced with a call to a Splinter REST API endpoint, such that the variables match.


cli/src/action/login/oauth/callback.rs
```diff
@@ -0,0 +1,107 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+
+use tiny_http::{Header, Method, Request, Response, Server};
+
+use super::error::OAuth2CallbackError;
+
+pub struct OAuth2Callback {
```
ltseeley, line 21: There should be some documentation for this module and/or this struct, maybe just what's in the commit message. It may not be immediately clear what this is for when other people are looking at it in the future.

cli/src/action/login/oauth/callback.rs
```diff
@@ -0,0 +1,107 @@
...
+            .map_err(|err| OAuth2CallbackError(err.to_string()))?;
+
+        Ok(Self { join_handle, port })
+    }
+
+    pub fn address(&self) -> String {
+        format!("localhost:{}", self.port)
+    }
+
+    pub fn receive_authorization_code(self) -> Result<String, OAuth2CallbackError> {
```
ltseeley, line 45: Can we rename this to `wait_for_authorization_code` or something similar to indicate that this is a blocking method?

cli/src/action/login/oauth/callback.rs
```diff
@@ -0,0 +1,107 @@
...
+        })?
+    }
+
+    fn run_callback_server(server: Server) -> Result<String, OAuth2CallbackError> {
+        loop {
+            let request = server
+                .recv()
+                .map_err(|err| OAuth2CallbackError(err.to_string()))?;
+
+            if request.method() == &Method::Get && request.url().starts_with("/auth_response") {
```
ltseeley, line 57: Should "/auth_response" be defined as a constant somewhere? I assume this same string is used when sending the token request.

cli/src/action/login/oauth/callback.rs
```diff
@@ -0,0 +1,107 @@
...
+        if let Some(query) = split.last() {
+            let params = query.split('&');
+            for param in params {
+                let mut pair = param.split('=');
+                if let Some(key) = pair.next() {
+                    let val = pair.next().unwrap_or("");
+                    result
+                        .entry(key)
+                        .and_modify(|vals| vals.push(val))
+                        .or_insert_with(|| vec![val]);
```
ltseeley, line 100: It looks like you just take the first value for the `code` key above. Do we need to store multiple values per key? Do we need to get any query values other than `code`?

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+impl OAuth2Provider for GithubProvider {
+    fn get_tokens(&self) -> Result<UserTokens, OAuth2ProviderError> {
+        let client_id = self.client_id.ok_or_else(|| {
+            OAuth2ProviderError("Github client id was not configured for this binary".into())
+        })?;
+        let client_secret = self.client_secret.ok_or_else(|| {
+            OAuth2ProviderError("Github client secret was not configured for this binary".into())
+        })?;
+
+        let callback = callback::OAuth2Callback::new().expect("Unable to start callback");
```
ltseeley, line 39: Is this `expect` intentional?

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+            .authorize_url(CsrfToken::new_random)
+            // Set the desired scopes.
+            .add_scope(Scope::new("user:email".to_string()))
+            // Set the PKCE code challenge.
+            .set_pkce_challenge(pkce_challenge)
+            .url();
+
+        info!("Opening {}...", auth_url);
+        webbrowser::open(&auth_url.to_string()).map_err(|err| {
+            debug!("Unable to open authorization URL: {}", err);
```
ltseeley, line 70: Is this debug necessary with the error being returned? Seems like it would result in a double log.

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+            OAuth2ProviderError(
+                "Unable to open Authorization URL in browser; \
+                ensure that a default browser is installed."
+                    .into(),
+            )
+        })?;
+
+        let auth_code = callback
+            .receive_authorization_code()
+            .expect("Unable receive authorization code");
```
ltseeley, line 80: s/Unable receive/Unable to receive/

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        let token_result = client
```
ltseeley, line 82: Before exchanging the code, we should be making sure the CSRF state returned with the auth code matches the CSRF token we set for the original request (this is mentioned in the examples for the oauth2 crate).

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+            .exchange_code(AuthorizationCode::new(auth_code))
+            .set_pkce_verifier(pkce_verifier)
+            .request(|req| {
+                Client::new()
+                    .execute(to_reqwest_req(req)?)
+                    .map(to_oauth2_res)
+                    .map_err(|_err| {
+                        OAuth2ProviderError("Unable to execute for OAuth2 Access Token".into())
+                    })
+            })
```
ltseeley, line 92: You can do `.request(oauth2::reqwest::http_client)` to simplify this

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+        client_id: &str,
+        client_secret: &str,
+        access_token: &AccessToken,
+    ) -> Result<String, OAuth2ProviderError> {
+        let body = serde_json::to_string(&GithubAccesToken {
+            access_token: access_token.secret(),
+        })
+        .map_err(|_| OAuth2ProviderError("Unable to write token request to JSON".into()))?;
+
+        let response = reqwest::blocking::Client::new()
```
ltseeley, line 126: `Client` is already imported in this file, no need to namespace it here.

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+#[derive(serde::Serialize)]
+struct GithubAccesToken<'a> {
+    access_token: &'a str,
+}
+
+#[derive(serde::Deserialize)]
+struct GithubTokenInfo {
```
ltseeley, line 171: Should probably be `GithubUserInfo`?

cli/src/action/login/oauth/github.rs
```diff
@@ -0,0 +1,238 @@
...
+            }
+            Ok(res) => {
+                debug!(
+                    "Received response {}: {}",
+                    res.status(),
+                    res.text()
+                        .unwrap_or_else(|_| "<empty or binary content>".into())
+                );
+                Err(OAuth2ProviderError(
+                    "Unable to validate OAuth1 access token.".into(),
```
ltseeley, line 152: Should this be "OAuth2"?

cli/src/action/login/oauth/google.rs
```diff
@@ -0,0 +1,237 @@
...
+use super::{OAuth2Provider, OAuth2ProviderError, ProviderType, UserTokens};
+
+pub struct GoogleProvider {
+    client_id: Option<&'static str>,
+    client_secret: Option<&'static str>,
+}
+
+impl OAuth2Provider for GoogleProvider {
+    fn get_tokens(&self) -> Result<UserTokens, OAuth2ProviderError> {
+        let client_id = self.client_id.ok_or_else(|| {
```
ltseeley, line 32: It seems like we should be able to refactor a lot of this code to be shared between the provider implementations. Would make it easier to add more providers in the future.

ltseeley (changes requested):

cli/src/action/login/oauth/callback.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub struct OAuth2Callback {
+    port: u16,
+    join_handle: std::thread::JoinHandle<Result<String, OAuth2CallbackError>>,
+}
+
+impl OAuth2Callback {
+    pub fn new() -> Result<Self, OAuth2CallbackError> {
+        let server =
+            Server::http("127.0.0.1:0").map_err(|err| OAuth2CallbackError(err.to_string()))?;
```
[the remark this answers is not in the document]
peterschwarz: Port `0` has the OS pick an available port.  This is way the callback address is returned.

cli/src/action/login/oauth/callback.rs
peterschwarz: Coupled with the above comment about the port, the callback should just return the entire callback url.

cli/src/action/login/oauth/github.rs
peterschwarz: No, it is not

cli/src/action/login/oauth/github.rs
peterschwarz: The debug flag is there to give the details of the source error.  If we are trying to move away from creating the poor-man's stack trace, by appending strings to the messages, then we still need a way to get the details out.  Presumable the error returned would be logged as error, where this one would be logged if debug details are turned on (i.e. someone is debugging)

cli/src/action/login/oauth/github.rs
peterschwarz: The default features are disabled, and I made this choice since we already had reqwest as a dependency.  Did not seem necessary to pull in multiple versions of reqwest (or use the alternative of a curl client).

cli/src/action/login/oauth/github.rs
peterschwarz: Technically, this is the response from the gitub token endpoint, so this is accurately named.  It will also be expanded on when refresh is implemented.

cli/src/action/login/oauth/provider.rs
```diff
@@ -0,0 +1,183 @@
...
+use oauth2::{
+    AuthUrl, AuthorizationCode, ClientId, ClientSecret, CsrfToken, PkceCodeChallenge, RedirectUrl,
+    Scope, TokenResponse, TokenUrl,
+};
+use reqwest::blocking::Client;
+
+use super::callback;
+use super::{error::OAuth2ProviderError, UserTokens};
+
+pub struct Provider {
```
ltseeley, line 25: Needs module and struct docs

ltseeley (approved):

peterschwarz:
Changed storage format from JSON to YAML

ltseeley (approved):

peterschwarz:
Discussions around this PR have concluded that an alternative approach is required.

PR #957, opened 2020-10-06
Remove serde serilization decorators from AdminServiceStore struct API

agunde406:
Serde should not be a part of the public API of the structs for the
AdminServiceStore.

This commit also add Yaml specific structs for CircuitProposal,
ProposedCircuit, and VoteRecord because the Vec<u8> in those
structs must be converted to a hex string for storage. This
was done automatically before with the serde decorators, and now
needs to be done explicitly.


ltseeley (approved):
LGTM

peterschwarz (approved):

PR #962, opened 2020-10-12
REST API OAuth config

ltseeley:
Adds basic OAuth provider configuration for the Splinter REST API

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -45,6 +45,7 @@ jsonwebtoken = { version = "6.0", optional = true }
 log = "0.3.0"
 mio = "0.6"
 mio-extras = "2"
+oauth2 = { version = "3.0", optional = true }
 openssl = "0.10"
 percent-encoding = { version = "2.0", optional = true }
 protobuf = "2"
@@ -121,7 +122,7 @@ biome-notifications = ["biome"]
 biome-user = ["biome"]
 circuit-template = ["glob"]
 events = ["actix-http", "futures", "hyper", "tokio", "awc"]
-oauth = ["auth"]
+oauth = ["auth", "oauth2"]
 postgres = ["diesel/postgres", "diesel_migrations"]
 registry = []
 registry-database = ["diesel"]
--- a/libsplinter/src/auth/oauth/error.rs
+++ b/libsplinter/src/auth/oauth/error.rs
@@ -1,0 +1,40 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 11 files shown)


libsplinter/src/auth/oauth/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    InvalidAuthUrl(String),
+    /// The specified token URL for the provider was invalid
+    InvalidTokenUrl(String),
+}
+
+impl fmt::Display for ProviderConfigurationError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::InvalidAuthUrl(msg) => {
+                write!(f, "provided authorization URL is invalid: {}", msg)
```
ltseeley, line 33: Currently just adding the source error to the message here. One alternative would be to map each of the variants of the source error (https://docs.rs/url/2.1.1/url/enum.ParseError.html) to a specific message that we have control over.

libsplinter/src/rest_api/mod.rs
```diff
@@ -564,6 +569,8 @@ pub struct RestApiBuilder {
     bind: Option<String>,
     #[cfg(feature = "rest-api-cors")]
     whitelist: Option<Vec<String>>,
+    #[cfg(all(feature = "rest-api-auth", feature = "oauth"))]
+    oauth_provider: Option<OAuthProvider>,
```
peterschwarz, line 573: This isn't configuration code, as it contains the oauth Client instance.  The forthcoming endpoint to provide the CLI with correct information will need the data that was provided to this OAuthProvider. Its current implementation cannot return the values, such as token url or client id/secret that the CLI will need.

peterschwarz (changes requested):

libsplinter/Cargo.toml
```diff
@@ -137,6 +138,7 @@ rest-api = [
     "percent-encoding",
 ]
 rest-api-actix = ["actix", "actix-http", "actix-web", "actix-web-actors"]
+rest-api-auth = ["auth"]
```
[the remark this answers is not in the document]
ltseeley: This is a bit different, though, since it isn't just guarding some endpoints; it guards some functionality of the REST API itself. I added this feature on the assumption that enabling the `auth` module is distinct from enabling auth for the REST API. This could be useful if we use the `auth` module for something other than the REST API. If this isn't something that seems necessary, though, I can remove this feature.

libsplinter/src/rest_api/mod.rs
ltseeley: Ah, right, forgot about that endpoint. Will update.

libsplinter/Cargo.toml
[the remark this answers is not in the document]
ltseeley: Removed this feature

libsplinter/src/rest_api/mod.rs
ltseeley: After conversations with @peterschwarz, we agreed to leave this as-is because we will likely not need to return the values.

peterschwarz (approved):

shannynalayna (approved):

PR #966, opened 2020-10-14
Make the field name for authorization type consistent in Circuit and CircuitProposal

agunde406:
The migrations file does not currently run successfully (This will be fixed in the next PR) so the migration file is updated directly.


agunde406:
authorization is reserved word in postgres, closing

PR #973, opened 2020-10-16
Test `CircuitTemplateManager` public methods

shannynalayna:
Adds tests for the public `CircuitTemplateManager` methods that were
previously not covered by unit tests to support the stabilization of the
`circuit-template` feature.

### DIFF
--- a/libsplinter/src/circuit/template/mod.rs
+++ b/libsplinter/src/circuit/template/mod.rs
@@ -528,6 +528,114 @@ rules:
         assert!(template.is_ok());
     }
 
+    /// Verifies a `CircuitTemplateManager` can be created using multiple paths, which are then
+    /// used to find the template files using `list_available_templates`.
+    ///
+    /// The test follows the procedure below:
+    /// 1. Sets up a temporary directory, to write circuit template files from the
+    ///    `EXAMPLE_TEMPLATE_YAML`.
+    /// 2. Create a `CircuitTemplateManager` using multiple temporary directories.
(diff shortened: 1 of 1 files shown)


agunde406 (approved):

libsplinter/src/circuit/template/mod.rs
```diff
@@ -540,4 +648,29 @@ rules:
         file.write_all(data)
             .expect("Error writing example template yaml.");
     }
+
+    fn verify_example_yaml_string(yaml: String) {
+        // Validate the main headers in the YAML string
+        assert!(yaml.contains("version: v1"));
```
peterschwarz, line 654: I would suggest also adding:

```
let _: serde_yaml::Value = serde_yaml::from_str(&yaml).expect("Invalid yaml was returned");
```
That way, you validate that it is actually parseable 

(likewise, you could then inspect the contents of that `Value`, but that seems like a lot of effort for something similar to your `contains` calls).

shannynalayna: I didn't initially include this because the `load_raw_yaml` is just returning the `serde_yaml::to_string` method so that seemed redundant, but now I see this being useful in the future if that method is ever changed so I included the line you suggested. I left the string inspection with the `contains` calls as-is.

peterschwarz (approved):

agunde406 (approved):

vaporos (approved):

PR #977, opened 2020-10-21
Add authentication/authorization to Splinter REST API

ltseeley:
Adds the Authorization middleware to the Actix REST API

This PR can be tested by checking out the test branch (https://github.com/ltseeley/splinter/tree/rest-api-auth-guard-test) and running `docker-compose up --build` from the root of the repository.

### DIFF
--- a/libsplinter/src/auth/oauth/mod.rs
+++ b/libsplinter/src/auth/oauth/mod.rs
@@ -93,6 +93,30 @@ impl OAuthClient {
         })
     }
 
+    /// Creates a new `OAuthClient` with GitHub's authorization and token URLs.
+    ///
+    /// # Arguments
+    ///
+    /// * `client_id` - The GitHub OAuth client ID
+    /// * `client_secret` - The GitHub OAuth client secret
+    /// * `redirect_url` - The endpoint that GitHub will redirect to after it has completed
+    ///   authorization
+    #[cfg(feature = "oauth-github")]
+    pub fn new_github(
+        client_id: String,
+        client_secret: String,
+        redirect_url: String,
+    ) -> Result<Self, OAuthClientConfigurationError> {
+        Self::new(
+            client_id,
+            client_secret,
+            "https://github.com/login/oauth/authorize".into(),
+            redirect_url,
+            "https://github.com/login/oauth/access_token".into(),
+            vec![],
+        )
+    }
+
     /// Generates the URL that the end user should be redirected to for authorization
     pub fn get_authorization_url(&self) -> Result<String, OAuthClientError> {
         let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
--- a/libsplinter/src/auth/rest_api/actix.rs
+++ b/libsplinter/src/auth/rest_api/actix.rs
@@ -1,0 +1,130 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Authorization middleware for the Actix REST API
+
+use actix_web::dev::*;
+use actix_web::{Error as ActixError, HttpResponse};
+use futures::{
+    future::{ok, FutureResult},
+    Future, IntoFuture, Poll,
+};
+
+use crate::rest_api::ErrorResponse;
+
+use super::{authorize, identity::IdentityProvider, AuthorizationResult};
+
+/// Wrapper for the authorization middleware
+#[derive(Clone)]
+pub struct Authorization {
+    identity_providers: Vec<Box<dyn IdentityProvider>>,
+}
+
+impl Authorization {
+    pub fn new(identity_providers: Vec<Box<dyn IdentityProvider>>) -> Self {
+        Self { identity_providers }
+    }
+}
+
+impl<S, B> Transform<S> for Authorization
+where
+    S: Service<Request = ServiceRequest, Response = ServiceResponse<B>, Error = ActixError>,
+    S::Future: 'static,
+    B: 'static,
+{
+    type Request = ServiceRequest;
+    type Response = ServiceResponse<B>;
+    type Error = S::Error;
+    type InitError = ();
+    type Transform = AuthorizationMiddleware<S>;
+    type Future = FutureResult<Self::Transform, Self::InitError>;
+
+    fn new_transform(&self, service: S) -> Self::Future {
+        ok(AuthorizationMiddleware {
+            identity_providers: self.identity_providers.clone(),
+            service,
+        })
+    }
+}
+
+/// Authorization middleware for the Actix REST API
+pub struct AuthorizationMiddleware<S> {
+    identity_providers: Vec<Box<dyn IdentityProvider>>,
+    service: S,
+}
+
+impl<S, B> Service for AuthorizationMiddleware<S>
+where
+    S: Service<Request = ServiceRequest, Response = ServiceResponse<B>, Error = ActixError>,
+    S::Future: 'static,
+    B: 'static,
+{
+    type Request = ServiceRequest;
+    type Response = ServiceResponse<B>;
+    type Error = S::Error;
+    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error>>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        let auth_header =
+            match req
+                .headers()
+                .get("Authorization")
+                .map(|auth| auth.to_str())
+                .transpose()
+            {
+                Ok(opt) => opt,
+                // Not including the error since it could leak secrets from the Authorization header
+                Err(_) => return Box::new(
+                    req.into_response(
+                        HttpResponse::BadRequest()
+                            .json(ErrorResponse::bad_request(
+                                "Authorization header must contain only visible ASCII characters",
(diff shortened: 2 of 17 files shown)


libsplinter/src/auth/identity/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+impl Clone for Box<dyn IdentityProvider> {
+    fn clone(&self) -> Box<dyn IdentityProvider> {
+        self.clone_box()
+    }
+}
+
+/// The authentication credentials that are passed to an `IdentityProvider`
+pub struct AuthCredentials {
+    /// The authentication type. Each identity provider may support only certain types.
+    auth_type: String,
```
peterschwarz, line 50: Should this be an enum?

libsplinter/src/auth/rest_api/actix.rs
```diff
@@ -0,0 +1,115 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Authorization middleware for the Actix REST API
+
+use actix_web::dev::*;
```
peterschwarz, line 17: This should probably be explicit uses.  We generally only use `*` when a prelude modules is available.

libsplinter/src/auth/rest_api/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        is_auth_endpoint = true;
+    }
+
+    if is_auth_endpoint {
+        AuthorizationResult::NoAuthorizationNecessary
+    } else {
+        match auth_header {
+            Some(auth_str) => {
+                // Parse the auth header
+                let mut auth_str_parts = auth_str.split_whitespace();
```
peterschwarz, line 54: What's the expected format?  How does type play into this?

libsplinter/src/auth/identity/mod.rs
ltseeley: That was my initial thought, but that would restrict the API to only support types implemented by the library itself; you couldn't add external implementations for custom types without modifying the library. Though, we could have an enum with known types, and a `Custom(String)` variant.

libsplinter/src/auth/rest_api/actix.rs
ltseeley: This is a prelude module: https://docs.rs/actix-web/1.0.5/actix_web/dev/index.html

libsplinter/src/auth/rest_api/mod.rs
ltseeley: This is the standard HTTP Authorization header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization. The format is "<type> <credentials>". I went ahead and added some extra details about the arguments to the function's docs, and the openapi.yaml will include information about this header when it gets updated.

libsplinter/src/auth/identity/mod.rs
peterschwarz: I think that would be a good compromise

libsplinter/src/auth/rest_api/actix.rs
peterschwarz: :/ Thanks

libsplinter/src/auth/rest_api/mod.rs
peterschwarz: Could you add a comment?

libsplinter/src/auth/identity/mod.rs
ltseeley: Replaced this struct with a couple enums that provide better structure and typing to the API.

libsplinter/src/auth/rest_api/mod.rs
ltseeley: Added some comments about the expected formats

libsplinter/src/auth/rest_api/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    if is_auth_endpoint {
+        return AuthorizationResult::NoAuthorizationNecessary;
+    }
+
+    // Parse the auth header, which should be in the format "<scheme> <value>"
+    let auth_str = match auth_header {
+        Some(auth_str) => auth_str,
+        None => return AuthorizationResult::Unauthorized,
+    };
+    let mut auth_str_parts = auth_str.splitn(2, ' ');
```
peterschwarz, line 62: You should implement `FromStr` for this

ltseeley: Implemented, also moved the `identity` module into the `auth::rest_api` module

peterschwarz (approved):

libsplinter/src/auth/rest_api/identity/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Parses an authorization string, which must be in the format "<scheme> <value>"
+impl FromStr for Authorization {
+    type Err = AuthorizationParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut authorization_parts = s.splitn(2, ' ');
+        let authorization_scheme = authorization_parts
+            .next()
+            .expect("cannot return empty iter");
```
vaporos, line 62: Suggest something more along these lines:

```
    let mut parts = s.splitn(2, ":");
    match (parts.next(), parts.next()) {
        (Some(part1), Some(part2)) => match part1 {
            "Bearer" => ...,
            other_schema => ...,
        }

        // avoid including the token in the error as it may contain secrets inappropriate for logging
        (Some(part1), None)  => Err(AuthorizationParseError::new("malformed token")),

        // splitn always returns at least one part
        _ => unreachable!(),
    }
```

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -93,6 +93,43 @@ impl OAuthClient {
...
+        client_id: String,
+        client_secret: String,
+        redirect_url: String,
+    ) -> Result<Self, OAuthClientConfigurationError> {
+        let client =
+            BasicClient::new(
+                ClientId::new(client_id),
+                Some(ClientSecret::new(client_secret)),
+                AuthUrl::new("https://github.com/login/oauth/authorize".into())
+                    .expect("Authorization URL should be valid"),
```
vaporos, line 115: Replace these expects with OAuthClientConfigurationError::InvalidXUrl() like in new(). Or maybe just have this function call new()?

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -93,6 +93,43 @@ impl OAuthClient {
...
+    /// Creates a new `OAuthClient` with GitHub's authorization and token URLs.
+    ///
+    /// # Arguments
+    ///
+    /// * `client_id` - The GitHub OAuth client ID
+    /// * `client_secret` - The GitHub OAuth client secret
+    /// * `redirect_url` - The endpoint that GitHub will redirect to after it has completed
+    ///   authorization
+    #[cfg(feature = "oauth-github")]
+    pub fn github(
```
vaporos, line 105: Maybe call this new_with_github_defaults() or new_github().

vaporos (changes requested):

libsplinter/src/auth/rest_api/actix.rs
```diff
@@ -0,0 +1,114 @@
...
+    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error>>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.service.poll_ready()
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        let auth_header = req.headers().get("Authorization").map(|auth| {
+            auth.to_str()
+                .expect("Authorization header has non-string value")
```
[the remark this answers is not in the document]
ltseeley: Good catch. Forgot to remove this expect. Updated.

libsplinter/src/auth/rest_api/identity/mod.rs
ltseeley: Updated

libsplinter/src/auth/oauth/mod.rs
ltseeley: More forgotten expects. Thanks again. Updated to just call `Self::new`

libsplinter/src/auth/oauth/mod.rs
ltseeley: Renamed to `new_github`

libsplinter/src/auth/rest_api/identity/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Parses an authorization string, which must be in the format "<scheme> <value>"
+impl FromStr for Authorization {
+    type Err = AuthorizationParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut parts = s.splitn(2, ':');
+        match (
+            parts.next().expect("cannot return empty iter"),
```
peterschwarz, line 61: this should be `ok_or_else(|| AuthorizationParseError...)`

libsplinter/src/auth/rest_api/identity/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Parses a bearer token string, which must be in the format "<type>:<value>"
+impl FromStr for BearerToken {
+    type Err = AuthorizationParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut parts = s.splitn(2, ':');
+        match (
+            parts.next().expect("cannot return empty iter"),
```
peterschwarz, line 92: this should be `ok_or_else(|| AuthorizationParseError...)`

libsplinter/src/auth/rest_api/identity/mod.rs
peterschwarz: I.e. it implies an empty string

ltseeley: Calling `splitn` on the empty string returns an iterator with a single item, the empty string. The iterator is never empty.

peterschwarz (dismissed):

peterschwarz (approved):

vaporos (approved):

peterschwarz (approved):

PR #981, opened 2020-10-22
Remaining AdminServiceStore derive annotations

isabeltomb:
Remove remaining serialization 'derive' annotations from AdminServiceStore structs at the trait level. Serialization should be
handled in reader/writer implementation.

Add serializable structs to the yaml implementation of AdminServiceStore and implement 'From' for each to allow for conversion between the trait level structs and the yaml structs.

### DIFF
--- a/libsplinter/src/admin/store/yaml/mod.rs
+++ b/libsplinter/src/admin/store/yaml/mod.rs
@@ -32,10 +32,11 @@ use self::error::YamlAdminStoreError;
 
 use super::{
     error::BuilderError, AdminServiceStore, AdminServiceStoreError, AuthorizationType, Circuit,
-    CircuitBuilder, CircuitNode, CircuitPredicate, CircuitProposal, CircuitProposalBuilder,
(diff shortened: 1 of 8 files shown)


libsplinter/src/admin/store/yaml/mod.rs
```diff
@@ -1130,7 +1131,14 @@ impl TryFrom<YamlCircuitState> for CircuitState {
 
     fn try_from(state: YamlCircuitState) -> Result<Self, Self::Error> {
         Ok(CircuitState {
-            nodes: state.nodes,
+            nodes: state
+                .nodes
+                .into_iter()
+                .map(|(id, node)| match CircuitNode::try_from(node) {
```
ltseeley, line 1137: This `match` looks like it can be replaced with a `.map()`

isabeltomb: Changed to use `.map()`

ltseeley (approved):

agunde406 (approved):

PR #988, opened 2020-10-29
Add key flag to Splinter CLI commands

isabeltomb:
'--key' flag added to all REST API calling Splinter CLI commands. The flag allows the user to specify the path to the file containing the private key.


vaporos (commented):
Do we need man page updates for this as well?

vaporos (approved):

peterschwarz (approved):

PR #989, opened 2020-10-29
Replace the use of sawtooth-sdk with transact-sdk-javascript

agunde406:
This commit updates the gameroom-app to use the transact sdk
instead of the sawtooth sdk.

This dependency is currently only available through github so
the docker image has to be changed to install as a non root
user.

This will also fix a build issue with the current gameroom-app
build.


docker/lint-client
```diff
@@ -15,11 +15,12 @@
...
 FROM node:lts-alpine as build-stage
 
-RUN apk update && apk add python g++ make && rm -rf /var/cache/apk/*
+RUN apk update && apk add python g++ git make && rm -rf /var/cache/apk/*
 
 WORKDIR /gameroom
 COPY package*.json ./
 
-RUN npm install
+# Need to set as none root user to properly install transact-sdk-javascript
```
ltseeley, line 23: s/none root/non-root/ (also in the other docker file)

agunde406: fixed

ltseeley (approved):

peterschwarz (approved):

PR #990, opened 2020-10-30
Update splinterd OpenAPI spec for auth updates

ltseeley:


splinterd/api/static/openapi.yaml
```diff
@@ -1568,6 +1568,74 @@ paths:
                 schema:
                   $ref: '#/components/schemas/ErrorBiome'
 
+  /oauth/login:
```
peterschwarz, line 1571: These aren’t exactly REST API endpoints. It should probably be more clear in the documentation that the expected client (at the moment) is the browser.

vaporos (approved):

peterschwarz (approved):

PR #991, opened 2020-10-30
Implement redirect from `oauth/callback`

shannynalayna:
Adds a `redirect_url` query parameter to the `oauth/login` endpoint,
used to direct the UI once the `oauth/callback` has successfully
exchanged the authorization code for the access token.

This also changes the `oauth/callback` behavior to successfully respond
even when the associated authorization request has expired, but without
an access token, rather than responding with an `Internal Server Error`.

To test: 
Splinter branch with oauth enabled in `splinterd` and the commit from this PR cherry-picked on top: https://github.com/shannynalayna/splinter/tree/shannynalayna-oauth-test, and then use this version of the UI: https://github.com/shannynalayna/splinter-ui/tree/shannynalayna-oauth-login. The UI test branch assumes the splinter and UI repos are at the same level (Splinter is used locally in the UI).

Once the test branches are checked out, run `docker-compose up --build` from within the Splinter-UI directory, go to `localhost:3030` and log-in or register, then navigate to `localhost:3030/oauth-login`, press the button and begin the authentication process. Once you are authorized through GitHub, you should be routed back to `localhost:3030` (which is the functionality implemented in this PR) with the access token information in the page's URL

### DIFF
--- a/libsplinter/src/auth/oauth/mod.rs
+++ b/libsplinter/src/auth/oauth/mod.rs
@@ -43,8 +43,9 @@ const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
 pub struct OAuthClient {
     /// The inner OAuth2 client
     client: BasicClient,
-    /// List of (CSRF token, PKCE verifier) pairs for pending authorization requests
-    pending_authorizations: Arc<Mutex<TtlMap<String, String>>>,
+    /// Pending authorization requests, including the CSRF token, PKCE verifier, and client's
+    /// redirect URL
+    pending_authorizations: Arc<Mutex<TtlMap<String, PendingAuthorization>>>,
     /// The scopes that will be requested for each user that's authenticated
     scopes: Vec<String>,
 }
@@ -118,7 +119,15 @@ impl OAuthClient {
     }
 
     /// Generates the URL that the end user should be redirected to for authorization
-    pub fn get_authorization_url(&self) -> Result<String, OAuthClientError> {
+    ///
+    /// # Arguments
+    ///
+    /// * `client_redirect_url` - The endpoint that Splinter will redirect to after it has
+    ///   completed authorization and the code exchange
+    pub fn get_authorization_url(
+        &self,
+        client_redirect_url: String,
+    ) -> Result<String, OAuthClientError> {
         let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
 
         let mut request = self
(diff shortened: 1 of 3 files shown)


libsplinter/src/auth/oauth/mod.rs
```diff
@@ -43,8 +44,9 @@ const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
 pub struct OAuthClient {
     /// The inner OAuth2 client
     client: BasicClient,
-    /// List of (CSRF token, PKCE verifier) pairs for pending authorization requests
-    pending_authorizations: Arc<Mutex<TtlMap<String, String>>>,
+    /// Information pertaining to pending authorization requests, including the CSRF token, PKCE
+    /// verifier, and client's redirect URL
+    pending_authorizations: Arc<Mutex<TtlMap<String, HashMap<String, String>>>>,
```
ltseeley, line 49: We should define a new struct for this information instead of using a HashMap.

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -150,15 +164,27 @@ impl OAuthClient {
         &self,
         auth_code: String,
         csrf_token: &str,
-    ) -> Result<Option<UserTokens>, OAuthClientError> {
-        let pkce_verifier = match self
+    ) -> Result<(Option<UserTokens>, String), OAuthClientError> {
```
ltseeley, line 167: Method docs should be updated to explain the return values

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -35,6 +35,14 @@ pub use error::{OAuthClientConfigurationError, OAuthClientError};
 /// The amount of time before a pending authorization expires and a new request must be made
 const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
 
+/// Information pertaining to pending authorization requests, including the PKCE verifier, and
+/// client's redirect URL
+#[derive(Clone)]
+struct PendingAuthorization {
+    pub(self) pkce_verifier: String,
```
peterschwarz, line 42: I don't think `pub(self)` - the struct isn't public

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -118,7 +127,12 @@ impl OAuthClient {
     }
 
     /// Generates the URL that the end user should be redirected to for authorization
-    pub fn get_authorization_url(&self) -> Result<String, OAuthClientError> {
+    ///
+    /// # Arguments
+    ///
+    /// * `redirect_url` - The endpoint that Splinter will redirect to after it has completed
+    ///   authorization and the code exchange
+    pub fn get_authorization_url(&self, redirect_url: String) -> Result<String, OAuthClientError> {
```
ltseeley, line 135: Can we rename this to `client_redirect_url` to clearly distinguish it from the OAuth redirect URL?

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -35,6 +35,14 @@ pub use error::{OAuthClientConfigurationError, OAuthClientError};
 /// The amount of time before a pending authorization expires and a new request must be made
 const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
 
+/// Information pertaining to pending authorization requests, including the PKCE verifier, and
+/// client's redirect URL
+#[derive(Clone)]
+struct PendingAuthorization {
```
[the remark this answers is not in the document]
shannynalayna: Updated

libsplinter/src/auth/oauth/mod.rs
shannynalayna: Updated

ltseeley (approved):

peterschwarz (approved):

PR #995, opened 2020-11-04
Add crate::error::InvalidStateError

vaporos:
An error returned when an operation cannot be completed because the
state of the struct is inconsistent.

### DIFF
--- a/libsplinter/src/error/invalid_state.rs
+++ b/libsplinter/src/error/invalid_state.rs
@@ -1,0 +1,71 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Module containing InvalidStateError implementation.
+
+use std::error;
+use std::fmt;
+
+/// An error returned when an operation cannot be completed because the state of the underlying
+/// struct is inconsistent.
+///
+/// This can be caused by a caller when a sequence of functions is called in a way that results in
+/// a state which is inconsistent.
+///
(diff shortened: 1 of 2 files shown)


peterschwarz (approved):

libsplinter/src/error/invalid_state.rs
```diff
@@ -0,0 +1,71 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Module containing InvalidStateError implementation.
+
+use std::error;
+use std::fmt;
+
+/// An error returned when an operation cannot be completed because the state of the struct is
```
ltseeley, line 20: I don't know if it's always the case that the thing that returns this error is a struct (could be a trait method, for instance).

libsplinter/src/error/invalid_state.rs
```diff
@@ -0,0 +1,71 @@
...
+/// inconsistent.
+///
+/// This can occur by a caller when a sequence of functions is called in a way that results in a
```
ltseeley, line 23: "This can occur by a caller" should be reworded

libsplinter/src/error/invalid_state.rs
```diff
@@ -0,0 +1,71 @@
...
+/// file state which is inconsistent.
```
ltseeley, line 24: I assume "file state" is just an example? Can you update the comment to make that clear?

libsplinter/src/error/invalid_state.rs
vaporos: Yeah, I wanted to say 'object' here, if it were another language. Traits are implemented on structs and it would be the data in the struct that would make it invalid. The only thing not backed by a struct would be a standalone function, but then its probably an InvalidArgumentError.

vaporos: I changed this to say "underlying struct" to attempt to make it more clear.

libsplinter/src/error/invalid_state.rs
vaporos: Changed to "This can be caused by a caller when"

libsplinter/src/error/invalid_state.rs
vaporos: Removed 'file'. I think this was left over from a previous iteration of the statement where I did have an example.

ltseeley (approved):

agunde406 (approved):

PR #1000, opened 2020-11-09
Update the `oauth/*` endpoints

shannynalayna:
Updates the `oauth/login` endpoint to add the `redirect_url` query
parameter. Also updates the `oauth/callback` endpoint to update the
response to the redirect and explain the contents of that
redirect URL.

### DIFF
--- a/splinterd/api/static/openapi.yaml
+++ b/splinterd/api/static/openapi.yaml
@@ -1577,6 +1577,18 @@ paths:
         authenticate the user.
       parameters:
         - $ref: "#/components/parameters/protocol_version"
+        - name: redirect_url
+          in: query
+          description: |
+            The URL the client will be redirected to once the OAuth
+            authentication process has been completed. If this parameter is not
+            provided, Splinter will attempt to retrieve the `referer` header
+            of the request. If this header value is also not available and the
+            `redirect_url` does not have a value, Splinter will respond with
+            400 Bad Request.
+          required: false
+          schema:
+            type: string
       responses:
         302:
           description: Found redirect
@@ -1586,7 +1598,9 @@ paths:
               schema:
                 type: string
         400:
-          description: Request was malformed
+          description: |
+            Request was malformed or no `redirect_url` was provided and the server
+            was unable to retrieve the request's `referer` header value.
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


splinterd/api/static/openapi.yaml
```diff
@@ -1577,6 +1577,14 @@ paths:
         authenticate the user.
       parameters:
         - $ref: "#/components/parameters/protocol_version"
+        - name: redirect_url
+          in: query
+          description: |-
```
ltseeley, line 1582: Is the `-` here intentional?

splinterd/api/static/openapi.yaml
```diff
@@ -1577,6 +1577,14 @@ paths:
...
+            The URL the client will be redirected to once the OAuth
+            authentication process has been completed.
```
ltseeley, line 1584: Should explain what happens when the parameter isn't provided.

splinterd/api/static/openapi.yaml
```diff
@@ -1603,28 +1613,26 @@ paths:
       tags:
         - OAuth
       description: |
-        Receives an authorization code from the OAuth provider and exchanges it
-        for an access token for the user.
+        Receives an authorization code from the OAuth provider to create an
+        access token for the user. Redirects to a URL provided by the client
+        when the OAuth authentication process is initiated.
```
ltseeley, line 1618: s/is/was/

splinterd/api/static/openapi.yaml
```diff
@@ -1603,28 +1613,26 @@ paths:
...
       parameters:
         - $ref: "#/components/parameters/protocol_version"
       responses:
-        200:
-          description: Obtained access token for user
-          content:
-            application/json:
+        302:
+          description: Found redirect
```
ltseeley, line 1623: This description should explain what this redirect means

ltseeley (changes requested):

splinterd/api/static/openapi.yaml
shannynalayna: `|-` seems to be used interchangeably with just `|` throughout the doc, I was just following an example that used `-`

splinterd/api/static/openapi.yaml
shannynalayna: I specified this in the 400 response description but can add a small explanation here as well. Was kind of awkward finding just the right place to put these descriptions...

ltseeley (approved):

peterschwarz (approved):

PR #1003, opened 2020-11-09
Add crate::error:ResourceTemporarilyUnavailableError

vaporos:
When a resource is not available, this error can be used to indicate
that the caller should (or may) retry the call again.

### DIFF
--- a/libsplinter/src/error/unavailable.rs
+++ b/libsplinter/src/error/unavailable.rs
@@ -1,0 +1,133 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


libsplinter/src/error/unavailable.rs
```diff
@@ -0,0 +1,133 @@
...
+            source,
+            retry_duration_hint: None,
+        }
+    }
+
+    /// Constructs a new `ResourceTemporarilyUnavailableError` from a specified source error with
+    /// a retry duration hint.
+    ///
+    /// The hint specified here can be used by the caller as the duration between retry attempts.
+    /// Callers may ignore this hint and provide their own algorithms, or may use this `Duraiton`
```
agunde406, line 56: Duraiton -> Duration

agunde406 (changes requested):

libsplinter/src/error/unavailable.rs
vaporos: fixed

peterschwarz (approved):

agunde406 (approved):

PR #1004, opened 2020-11-09
OAuth callback endpoint improvements

ltseeley:
A couple small improvements to the OAuth callback endpoint

### DIFF
--- a/libsplinter/src/auth/oauth/rest_api/resources/callback.rs
+++ b/libsplinter/src/auth/oauth/rest_api/resources/callback.rs
@@ -20,19 +20,14 @@ pub struct CallbackQuery {
     pub state: String,
 }
 
-#[derive(Debug, Serialize, Clone, PartialEq)]
-pub struct CallbackResponse<'a> {
-    pub access_token: &'a str,
-    pub expires_in: Option<u64>,
-    pub refresh_token: Option<&'a str>,
-}
-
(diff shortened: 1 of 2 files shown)


libsplinter/src/auth/oauth/rest_api/resources/callback.rs
```diff
@@ -12,27 +12,8 @@
...
-use crate::auth::oauth::UserTokens;
-
 #[derive(Deserialize)]
 pub struct CallbackQuery {
     pub code: String,
     pub state: String,
 }
-
-#[derive(Debug, Serialize, Clone, PartialEq)]
-pub struct CallbackResponse<'a> {
```
peterschwarz, line 24: Couldn't this struct provide a `to_query_params` function?

ltseeley: We could just as easily implement that function for the `UserTokens` struct, but I'm not sure how much value that provides since it only ever happens in this endpoint.

peterschwarz: Yes, but it the behaviour is actix-independent

ltseeley: Updated to be a method on `UserTokens`

peterschwarz (approved):

vaporos (approved):

PR #1006, opened 2020-11-10
Fix bugs in the DieselAdminServiceStore

agunde406:
Fixes two issues:
1. If a service has no arguments the service would not be returned when listing services
2. The circuits and proposals were not returned in a deterministic order. This will causes issues for the rest api paging functionality.

### DIFF
--- a/libsplinter/src/admin/store/diesel/operations/list_circuits.rs
+++ b/libsplinter/src/admin/store/diesel/operations/list_circuits.rs
@@ -66,8 +66,6 @@ where
             .flatten()
             .collect();
 
-        println!("{:?}", members);
-
         self.conn
             .transaction::<Box<dyn ExactSizeIterator<Item = Circuit>>, _, _>(|| {
                 // Collects circuits which match the circuit predicates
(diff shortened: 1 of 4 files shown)


peterschwarz (dismissed):

libsplinter/src/admin/store/diesel/operations/list_circuits.rs
```diff
@@ -87,19 +87,19 @@ where
...
-                    })?
-                    // Once the `CircuitModels` have been collected, organize into a HashMap
-                    .into_iter()
-                    .map(|model| (model.circuit_id.to_string(), model))
-                    .collect();
+                    })?;
 
                 // Store circuit IDs separately to make it easier to filter following queries
-                let circuit_ids: Vec<String> = circuits.keys().cloned().collect();
+                let circuit_ids: Vec<String> = circuits
```
peterschwarz, line 99: This can probably be `&str`.  Same below.

peterschwarz (approved):

ltseeley (approved):

PR #1007, opened 2020-11-10
Improve REST API auth config

ltseeley:


libsplinter/src/rest_api/errors.rs
```diff
@@ -49,6 +51,9 @@ impl fmt::Display for RestApiServerError {
             RestApiServerError::BindError(e) => write!(f, "Bind Error: {}", e),
             RestApiServerError::StartUpError(e) => write!(f, "Start-up Error: {}", e),
             RestApiServerError::StdError(e) => write!(f, "Std Error: {}", e),
+            RestApiServerError::InvalidConfig(field) => {
```
peterschwarz, line 54: This should probably use the `InvalidStateError` from `crate::error`

peterschwarz (changes requested):

libsplinter/src/rest_api/errors.rs
ltseeley: Updated

peterschwarz (approved):

shannynalayna (approved):

PR #1010, opened 2020-11-11
Add Cylinder JWT to Splinter CLI

isabeltomb:
- Add a function 'create_cylinder_jwt_auth' that uses Cylinder::jwt to build the authorization string that will be submitted the the REST API. The auth string includes a signed json web token. This function uses Cylinder's key loading module to retrieve the private key for signing tokens.

- Create a builder for splinter rest client which has a 'with_auth' function behind the splinter-cli-jwt feature to set the jwt_auth field of the splinter rest client when the feature is enabled.

- When the experimental feature splinter-cli-jwt is enabled use the 'create_cylinder_jwt_auth' function to create the jwt authorization string and call the splinter rest client builder 'with_auth' function.

- Modify the splinter rest client to submit an authorization string containing a jwt to the REST API when the splinter-cli-jwt feature is enabled.

- Because health::StatusAction doesn't use the Splinter rest client and instead contacts the REST API directly, the 'run' function is modified to create the jwt and submit it to the REST API in an authorization header.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -68,11 +68,14 @@ experimental = [
     "circuit-auth-type",
     "health",
     "sqlite",
+    "splinter-cli-jwt",
 ]
 
 circuit-auth-type = []
 circuit-template = ["splinter/circuit-template"]
 database-migrate-biome = ["splinter/biome"]
+
+splinter-cli-jwt = ["cylinder/jwt", "cylinder/key-load"]
 
 health = []
 
--- a/cli/src/action/api.rs
+++ b/cli/src/action/api.rs
@@ -67,8 +67,17 @@ pub struct SplinterRestClient {
 impl SplinterRestClient {
     /// Gets the Splinter node's status.
     pub fn get_node_status(&self) -> Result<NodeStatus, CliError> {
-        Client::new()
-            .get(&format!("{}/status", self.url))
+        // Allowing unused_mut because request must be mutable if experimental feature
+        // splinter-cli-jwt is enabled, if feature is removed unused_mut notation can be removed
+        #[allow(unused_mut)]
+        let mut request = Client::new().get(&format!("{}/status", self.url));
+
+        #[cfg(feature = "splinter-cli-jwt")]
+        {
+            request = request.header("Authorization", &self.auth);
+        }
+
+        request
             .send()
             .map_err(|err| CliError::ActionError(format!("Failed to fetch node ID: {}", err)))
             .and_then(|res| {
--- a/cli/src/action/api.rs
+++ b/cli/src/action/api.rs
@@ -19,17 +19,52 @@ use serde::Deserialize;
 
 use super::CliError;
 
-/// A wrapper around the Splinter REST API.
-pub struct SplinterRestClient<'a> {
-    pub url: &'a str,
+#[derive(Default)]
+pub struct SplinterRestClientBuilder {
+    pub url: Option<String>,
+    #[cfg(feature = "splinter-cli-jwt")]
+    pub auth: Option<String>,
 }
 
-impl<'a> SplinterRestClient<'a> {
-    /// Constructs a new client for a Splinter node at the given URL.
-    pub fn new(url: &'a str) -> Self {
-        Self { url }
+impl SplinterRestClientBuilder {
+    pub fn new() -> Self {
+        SplinterRestClientBuilder::default()
     }
 
+    pub fn with_url(mut self, url: String) -> Self {
+        self.url = Some(url);
+        self
+    }
+
+    #[cfg(feature = "splinter-cli-jwt")]
+    pub fn with_auth(mut self, auth: String) -> Self {
+        self.auth = Some(auth);
+        self
+    }
+
+    pub fn build(self) -> Result<SplinterRestClient, CliError> {
+        Ok(SplinterRestClient {
+            url: self.url.ok_or_else(|| {
+                CliError::ActionError("Failed to build client, url not provided".to_string())
+            })?,
+            #[cfg(feature = "splinter-cli-jwt")]
+            auth: self.auth.ok_or_else(|| {
+                CliError::ActionError(
+                    "Failed to build client, jwt authorization not provided".to_string(),
+                )
+            })?,
+        })
+    }
+}
+
+/// A wrapper around the Splinter REST API.
+pub struct SplinterRestClient {
+    pub url: String,
+    #[cfg(feature = "splinter-cli-jwt")]
+    pub auth: String,
+}
+
+impl SplinterRestClient {
     /// Gets the Splinter node's status.
     pub fn get_node_status(&self) -> Result<NodeStatus, CliError> {
         Client::new()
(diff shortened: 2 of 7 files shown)


cli/src/action/mod.rs
```diff
@@ -98,11 +99,21 @@ fn chown(path: &Path, uid: u32, gid: u32) -> Result<(), CliError> {
...
+///    implement std::fmt::Display
+fn read_private_key<T: AsRef<Path> + Copy>(
+    file_name: T,
+    print_file_name: &str,
+) -> Result<String, CliError> {
     let mut file = File::open(file_name).map_err(|err| {
         CliError::EnvironmentError(format!(
             "Unable to open key file '{}': {}",
-            file_name,
+            print_file_name,
```
vaporos, line 116: This can probably just be file_name.display()

cli/src/action/mod.rs
```diff
@@ -128,3 +139,48 @@ fn msg_from_io_error(err: IoError) -> String {
         _ => "Unknown I/O error".into(),
     }
 }
+
+// determine if a full file path or just a file name has been given for the private key
+fn determine_key_file_path(key: &str) -> Result<PathBuf, CliError> {
```
vaporos, line 144: We should take whatever the "best" code we have for this currently and make it a function in cylinder, as we probably have half a dozen different implementations currently.

cli/src/action/mod.rs
```diff
@@ -128,3 +139,48 @@ fn msg_from_io_error(err: IoError) -> String {
...
+                Err(CliError::ActionError(
+                    "failed to load signing key: could not be found".into(),
+                ))
+            }
+        }
+    }
+}
+
+// build a signed json web token using the private key
+fn create_token(key: &str) -> Result<String, CliError> {
```
vaporos, line 173: The argument name here, key, is confusing without reading the implementation of the function.

cli/src/action/circuit/mod.rs
```diff
@@ -166,7 +166,12 @@ impl Action for CircuitProposeAction {
                 .map(ToOwned::to_owned)
                 .or_else(|| std::env::var(SPLINTER_REST_API_URL_ENV).ok())
                 .unwrap_or_else(|| DEFAULT_SPLINTER_REST_API_URL.to_string());
-            let key = args.value_of("key").unwrap_or("./splinter.priv");
+
+            let key = args
+                .value_of("key")
+                .map(String::from)
+                .unwrap_or_else(whoami::username);
```
vaporos, line 173: We should move the use of whoami::username to the cylinder function for loading keys (see other comment about loading key via cylinder function).

vaporos (changes requested):

cli/src/action/circuit/mod.rs
isabeltomb: Fixed

cli/src/action/mod.rs
isabeltomb: Fixed

cli/src/action/mod.rs
isabeltomb: Fixed

cli/src/action/mod.rs
isabeltomb: Reverted this function to the original because it isn't needed in 'create_token' after moving key loading functionality to cylinder - fixed

cli/Cargo.toml
```diff
@@ -30,6 +30,7 @@ path = "src/main.rs"
 
 [dependencies]
 clap = "2"
+cylinder = { version = "0.2", features = ["jwt"] }
```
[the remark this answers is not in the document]
isabeltomb: fixed

cli/src/action/api.rs
```diff
@@ -65,8 +65,15 @@ pub struct SplinterRestClient {
 impl SplinterRestClient {
     /// Gets the Splinter node's status.
     pub fn get_node_status(&self) -> Result<NodeStatus, CliError> {
-        Client::new()
-            .get(&format!("{}/status", self.url))
+        #[allow(unused_mut)]
```
ltseeley, line 68: This should have a comment explaining why this lint is allowed; it should specifically mention the feature that, if removed, would allow us to remove this annotation. Same for the other occurrences in this PR.

cli/src/action/circuit/mod.rs
```diff
@@ -166,11 +169,37 @@ impl Action for CircuitProposeAction {
...
+                    .ok_or_else(|| CliError::EnvironmentError("Home directory not found".into()))?;
+                let default_path_string = {
+                    if let Some(path) = default_key_path.to_str() {
+                        path
+                    } else {
+                        return Err(CliError::ActionError(
+                            "Path is not valid unicode".to_string(),
+                        ));
+                    }
+                };
```
ltseeley, line 194: IMO, this should all be part of the `create_token` function. The only thing that needs to be passed to that function is the `key` argument.

cli/src/action/api.rs
```diff
@@ -65,8 +65,15 @@ pub struct SplinterRestClient {
...
+        let mut request = Client::new().get(&format!("{}/status", self.url));
+
+        #[cfg(feature = "splinter-cli-jwt")]
+        {
+            request = request.header("Authorization", format!("Bearer Cylinder:{}", &self.token));
```
ltseeley, line 73: I think it would be a better design to have the client take the entire authorization string (`Bearer Cylinder:<token>`) instead of just the token. The client shouldn't be restricted to only one type of authorization; in the future, we may decide it should support some other kind of authorization as well.

I'd suggest updating the `create_token` method to be something like `create_cylinder_jwt_auth` where it returns the whole authorization string, since the token is only really useful as part of the auth string.

cli/src/action/circuit/mod.rs
isabeltomb: Moved this to the function that creates the token

cli/src/action/api.rs
isabeltomb: Changed client to have a `jwt-auth` field instead of a token field which will hold the full authorization string. Changed create_token to create_cylinder_jwt_auth and modified it to build the whole authorization string

cli/src/action/api.rs
isabeltomb: Added comments

cli/src/action/api.rs
```diff
@@ -19,17 +19,52 @@ use serde::Deserialize;
...
+        SplinterRestClientBuilder::default()
+    }
+
+    pub fn with_url(mut self, url: String) -> Self {
+        self.url = Some(url);
+        self
     }
 
+    #[cfg(feature = "splinter-cli-jwt")]
+    pub fn with_jwt_auth(mut self, jwt_auth: String) -> Self {
```
ltseeley, line 40: This should just be `with_auth`, since the auth string isn't explicitly restricted to JWTs.

isabeltomb: Fixed

ltseeley (approved):

vaporos (approved):

peterschwarz (approved):

PR #1011, opened 2020-11-11
Retrieve user's identity before returning tokens

shannynalayna:
Adds a call using the `IdentityProvider` to retrieve the user's identity to be used for the UI's `display_name` value.

TESTING:

Set these environment variables or in the docker-compose file:
 
```
export OAUTH_PROVIDER="github"
export OAUTH_CLIENT_ID="<your client ID>"
export OAUTH_CLIENT_SECRET="<your client secret>"
export OAUTH_REDIRECT_URL="http://localhost:8080/oauth/callback"
``` 
Or, in the docker-compose: 
```
            --oauth-provider github \
            --oauth-client-secret <your client secret> \
            --oauth-client-id <your client ID> \
```

Run `docker-compose up --build` from the root splinter directory, and go to `http://localhost:8080/oauth/login?redirect_url=redirect`. 

Note, the `redirect_url` isn't valid so you'll hit a 404 when the callback returns, but you should be able to see the returned data in the URL it attempted to hit. So, with `redirect`, the final URL should look something like: `redirect?access_token=1234&display_name=<your GitHub username>` but the web page doesn't exist so you'll get a 404. This is the point that the oauth login sapling handles the data in the URL and would redirect to the base sapling after creating the Canopy user with the data in the URL.

### DIFF
--- a/libsplinter/src/auth/oauth/mod.rs
+++ b/libsplinter/src/auth/oauth/mod.rs
@@ -22,15 +22,13 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use oauth2::{
-    basic::{BasicClient, BasicTokenResponse},
-    reqwest::http_client,
-    AuthUrl, AuthorizationCode, ClientId, ClientSecret, CsrfToken, PkceCodeChallenge,
-    PkceCodeVerifier, RedirectUrl, Scope, TokenResponse, TokenUrl,
+    basic::BasicClient, reqwest::http_client, AuthUrl, AuthorizationCode, ClientId, ClientSecret,
+    CsrfToken, PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, Scope, TokenResponse, TokenUrl,
 };
 
 #[cfg(feature = "oauth-github")]
 use crate::auth::rest_api::identity::github::GithubUserIdentityProvider;
-use crate::auth::rest_api::identity::IdentityProvider;
+use crate::auth::rest_api::identity::{Authorization, BearerToken, IdentityProvider};
 use crate::collections::TtlMap;
 
 pub use error::{OAuthClientConfigurationError, OAuthClientError};
@@ -175,7 +173,7 @@ impl OAuthClient {
         &self,
         auth_code: String,
         csrf_token: &str,
-    ) -> Result<Option<(UserTokens, String)>, OAuthClientError> {
+    ) -> Result<Option<(UserInfo, String)>, OAuthClientError> {
         let pending_authorization = match self
             .pending_authorizations
             .lock()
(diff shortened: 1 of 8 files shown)


libsplinter/src/auth/oauth/mod.rs
```diff
@@ -194,6 +203,22 @@ impl OAuthClient {
             pending_authorization.client_redirect_url,
         )))
     }
+
+    /// Uses an access token to retrieve the user's identity from the OAuth provider.
+    ///
+    /// # Arguments
+    ///
+    /// * `access_token` - The access token required to authorize the request for the identity
+    pub fn get_identity(&self, access_token: String) -> Result<String, OAuthClientError> {
```
ltseeley, line 212: The identity should be added to the tokens struct (which should probably be renamed now) instead of being returned with a separate method.

ltseeley (changes requested):

libsplinter/src/rest_api/mod.rs
```diff
@@ -806,19 +806,18 @@ impl RestApiBuilder {
...
-                                    Box::new(OAuthUserStoreSaveTokensOperation::new(
+                                    Box::new(OAuthUserStoreSaveUserInfoOperation::new(
                                         oauth_provider,
-                                        oauth_identity_provider.clone(),
                                         user_store,
                                         oauth_user_store,
                                     ))
                                 }
-                                TokenSaveConfig::NoOp => Box::new(SaveTokensToNull),
+                                TokenSaveConfig::NoOp => Box::new(SaveUserInfoToNull),
```
ltseeley, line 815: `TokenSaveConfig` should be renamed as well

shannynalayna: Updated

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -102,11 +108,13 @@ impl OAuthClient {
     /// * `client_secret` - The GitHub OAuth client secret
     /// * `redirect_url` - The endpoint that GitHub will redirect to after it has completed
     ///   authorization
+    /// * `identity_provider` - The OAuth identity provider used to retrieve user's identity
     #[cfg(feature = "oauth-github")]
     pub fn new_github(
         client_id: String,
         client_secret: String,
         redirect_url: String,
+        identity_provider: Box<dyn IdentityProvider>,
```
peterschwarz, line 115: Given this is `new_github`, should this method just create the client with the github identity provider?

ltseeley: Yeah, this is a good point; this method shouldn't take an identity provider, it should create the `GitHubUserIdentityProvider` internally.

shannynalayna: That makes sense looking at the `GitHubUserIdentityProvider`

libsplinter/src/auth/oauth/mod.rs
```diff
@@ -102,6 +110,7 @@ impl OAuthClient {
     /// * `client_secret` - The GitHub OAuth client secret
     /// * `redirect_url` - The endpoint that GitHub will redirect to after it has completed
     ///   authorization
+    /// * `identity_provider` - The OAuth identity provider used to retrieve user's identity
```
ltseeley, line 111: This is no longer an argument

shannynalayna: Updated

ltseeley (dismissed):

peterschwarz (dismissed):

ltseeley (approved):

peterschwarz (approved):

PR #1027, opened 2020-11-17
Add missing fields to CircuitProposal

agunde406:

### DIFF
--- a/libsplinter/src/admin/store/diesel/mod.rs
+++ b/libsplinter/src/admin/store/diesel/mod.rs
@@ -876,6 +876,8 @@ pub mod tests {
                             .build().expect("Unable to build node"),
                         ]
                     )
+                    .with_application_metadata(b"test")
+                    .with_comments("This is a test")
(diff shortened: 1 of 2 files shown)


ltseeley (approved):

libsplinter/src/admin/store/diesel/mod.rs
```diff
@@ -876,6 +876,8 @@ pub mod tests {
                             .build().expect("Unable to build node"),
                         ]
                     )
+                    .with_application_metadata(b"test")
```
peterschwarz, line 879: Is there a test that validates that these fields were stored?

agunde406: yes, test_add_get_proposals creates this proposal, stores it and validates that on get this same proposal is returned

agunde406: Those fields are just optional so were not included in the original test by mistake

peterschwarz (approved):

PR #1032, opened 2020-11-18
Integrate AdminServiceStore and RoutingTable into Splinterd

agunde406:
This PR replaces the use of SplinterState and OpenProposals with AdminServiceStore and the RoutingTable. Before splinter state could only be persisted using yaml files. After this PR there is experimental support for SQLite (requires stabilizing sqlite in `splinter database migrate`) and PostgreSQL (requires stablizing `admin-service-store-postgres`) backends for storing splinter state, as well as a backwards compatible YAML backend. 

Important API changes:
1. `--storage` is now hidden from the splinterd CLI but if provided will still use the existing YAML state files for circuits and proposals
2. `--database` is now always available as an option in splinterd CLI and will be used for both AdminServiceStore and Biome (if enabled) unless --storage is provided.
3. SplinterState, CircuitDirectory, OpenProposals, and splinter::storage have been removed. They are no longer used and splinter::storage is not the pattern we wish to support going forwards. As such splinter::storage::sets has been moved to splinter::sets
4. The default state file location will be changed to /var/lib/splinter/splinter_state.db (a SQLite database) but this will require 
running `splinter database migrate` before starting splinter

The AdminService is now the sole owner of circuit and proposal state. To share this information with other splinter components the AdminService will use a RoutingTableWriter to update the circuit routing information (and rebuild the routing table on startup). Other components, such as the Circuit message dispatch Handlers, will now use a RoutingTableReader to access routing information. The RoutingTable information is currently only stored in memory.

With no experimental features the YamlAdminServiceStore is the only available backend (add --storage yaml to the command line or specify in the splinter config)


peterschwarz (approved):
Tested

splinterd/src/daemon.rs
```diff
@@ -133,28 +132,65 @@ impl SplinterDaemon {
...
+                    }
+                    _ => {
+                        return Err(StartError::StorageError(format!(
+                            "storage type is not supported: {}",
+                            storage
+                        )))
+                    }
+                }
+            } else {
+                //
```
ltseeley, line 182: Looks like a comment that never got filled out

vaporos (approved):

ltseeley (approved):

splinterd/src/daemon.rs
agunde406: will fix in another PR

PR #1034, opened 2020-11-18
OAuth stabilization updates

ltseeley:


libsplinter/src/auth/oauth/rest_api/actix/callback.rs
```diff
@@ -45,20 +45,17 @@ pub fn make_callback_route(
                             Ok(Some((user_info, redirect_url))) => {
                                 if let Err(err) = save_info_op.save_user_info(&user_info) {
                                     error!("Unable to store user info: {}", err);
-                                    HttpResponse::InternalServerError()
```
peterschwarz, line 48: I'm not sure that we should let this silently fail.  Other parts of the system may not work either, in ways that may be unpredictable.

ltseeley: You're right, removed this commit

peterschwarz (approved):

shannynalayna (approved):

PR #1038, opened 2020-11-19
Guard AdminService code with admin-service feature

agunde406:
The features is added to stable.

Also includes guarding code that is only used by the
admin service.

Also adds admin-service feature as a dependency to the
cli, splinterd and circuit-template.


libsplinter/src/base62.rs
```diff
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(feature = "admin-service")]
+use rand::{distributions::Alphanumeric, Rng};
```
ltseeley, line 16: Why is this dependency moved? Convention would have it in its previous spot.

agunde406: just a mistake,  ill move it back

agunde406: fixed

ltseeley (approved):

peterschwarz (approved):

PR #1039, opened 2020-11-19
Use top-level InvalidStateError for biome oauth

ltseeley:
Updates the Biome OAuth user store's `OAuthUserBuilder` to use the
top-level `InvalidStateError` instead of its own.

### DIFF
--- a/libsplinter/src/biome/oauth/store/mod.rs
+++ b/libsplinter/src/biome/oauth/store/mod.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Defines a representation of an OAuth user and provides an API to manage them.
+//! Defines a representation of OAuth users and provides an API to manage them.
 //!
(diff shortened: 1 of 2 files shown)


libsplinter/src/biome/oauth/store/mod.rs
```diff
@@ -21,7 +21,8 @@ pub(in crate::biome) mod diesel;
 mod error;
 pub(in crate::biome) mod memory;
 
-pub use error::InvalidStateError;
+pub use crate::error::InvalidStateError;
```
vaporos, line 24: I don't think we should pub use crate::error:: things to local modules.

ltseeley: Oops, good catch; forgot to remove the `pub`. Updated.

vaporos (approved):

vaporos (approved):

shannynalayna (approved):

peterschwarz (approved):

PR #1043, opened 2020-11-20
Add `/oauth/logout` endpoint

shannynalayna:
Adds an `/oauth/logout` endpoint that removes a user's access and refresh tokens from Biome-backed `OAuthUserStore` and adds a no-op implementation. 

Using docker, could probably just run `docker-compose up --build` and try to hit the endpoints as mentioned below.

Local testing: 
Setting up postgres backend:

```
```

Set up the tables:
```
$ cargo run --manifest-path cli/Cargo.toml -- database migrate -C postgres://postgres:mypassword@localhost:5432/splinter
```

Run Splinter: 
```
$ mkdir -p /tmp/splinter # a splinter home dir for local use
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml -- cert generate --skip
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features auth,biome-oauth -- \
    --database postgres://postgres:mypassword@localhost:5432/splinter \
    --enable-biome \
    --tls-insecure \
    --OAUTH_PROVIDER github \
    --OAUTH_CLIENT_ID <your client id> \
    --OAUTH_CLIENT_SECRET <your client secret> \
    --OAUTH_REDIRECT_URL http://localhost:8080/oauth/callback \
    -vv
```

In a browser, go to `localhost:8080/oauth/login?redirect_url=redirect`. (Note, you'll get a 404 because this URL does not exist, but you should see an access token and display name in the final URL that 404'd.

Check the database tables to make sure the `OAuthUser` has been stored: 
```
$ docker run -it --rm --link postgres:postgres postgres psql -h postgres -U postgres
Password for user postgres:
postgres-# \c splinter
You are now connected to database "splinter" as user "postgres".
splinter=# select * from oauth_user;
 id |               user_id                | provider_user_ref |               access_token               | refresh_token | provider_id
----+--------------------------------------+-------------------+------------------------------------------+---------------+-------------
  1 | 88074e7f-18f1-47f7-ab67-65c6968127ba | your_username     | aaaaaaaaaaaaaaaaaaaaaaa00000000000000000 |               |           1
(1 row)
```

Then, after taking the value from the `access_token` column, make the following curl request: 

```
$ curl --header "Authorization: Bearer OAuth2:$ACCESS_TOKEN" http://localhost:8080/oauth/logout
{"message": "User successfully logged out"}
```

You can go back to the postgres container and run `select * from oauth_user;` to verify the tokens have been removed from the oauth user's entry.

### DIFF
--- a/libsplinter/src/auth/oauth/rest_api/actix/logout.rs
+++ b/libsplinter/src/auth/oauth/rest_api/actix/logout.rs
@@ -1,0 +1,54 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The `GET /oauth/logout` endpoint for removing a user's tokens.
+
+use actix_web::HttpResponse;
+use futures::future::IntoFuture;
+
+use crate::auth::oauth::rest_api::OAuthUserInfoStore;
+use crate::protocol;
+use crate::rest_api::{ErrorResponse, Method, ProtocolVersionRangeGuard, Resource};
+
+use crate::biome::rest_api::auth::OAuthUserIdentityRef;
+
+pub fn make_logout_route(user_info_store: Box<dyn OAuthUserInfoStore>) -> Resource {
+    Resource::build("/oauth/logout")
+        .add_request_guard(ProtocolVersionRangeGuard::new(
+            protocol::OAUTH_LOGOUT_MIN,
+            protocol::OAUTH_PROTOCOL_VERSION,
+        ))
+        .add_method(Method::Get, move |req, _| {
+            Box::new(match req.extensions().get::<OAuthUserIdentityRef>() {
+                Some(OAuthUserIdentityRef(identity)) => {
+                    match user_info_store.remove_user_tokens(&identity) {
+                        Ok(()) => HttpResponse::Ok()
+                            .json(json!({
+                                "message": "User successfully logged out"
+                            }))
+                            .into_future(),
+                        Err(err) => {
+                            error!("Unable to remove user tokens: {}", err);
+                            HttpResponse::InternalServerError()
+                                .json(ErrorResponse::internal_error())
+                                .into_future()
+                        }
+                    }
+                }
+                None => HttpResponse::Unauthorized()
+                    .json(ErrorResponse::unauthorized())
+                    .into_future(),
+            })
+        })
+}
--- a/libsplinter/src/auth/oauth/rest_api/mod.rs
+++ b/libsplinter/src/auth/oauth/rest_api/mod.rs
@@ -72,6 +72,7 @@ impl OAuthUserInfoStore for OAuthUserInfoStoreNoOp {
 ///
 /// * `GET /oauth/login` - Get the URL for requesting authorization from the provider
 /// * `GET /oauth/callback` - Receive the authorization code from the provider
+/// * `GET /oauth/logout` - Remove the user's access and refresh tokens
 ///
 /// These endpoints are only available if the following REST API backend feature is enabled:
 ///
@@ -96,6 +97,7 @@ impl OAuthResourceProvider {
 ///
 /// * `GET /oauth/login` - Get the URL for requesting authorization from the provider
 /// * `GET /oauth/callback` - Receive the authorization code from the provider
+/// * `GET /oauth/logout` - Remove the user's access and refresh tokens
 ///
 /// These endpoints are only available if the following REST API backend feature is enabled:
 ///
@@ -115,6 +117,7 @@ impl RestResourceProvider for OAuthResourceProvider {
                     self.client.clone(),
                     self.user_info_store.clone(),
                 ),
+                actix::logout::make_logout_route(self.user_info_store.clone()),
             ]);
         }
 
--- a/libsplinter/src/auth/oauth/rest_api/mod.rs
+++ b/libsplinter/src/auth/oauth/rest_api/mod.rs
@@ -23,40 +23,40 @@ use crate::rest_api::{Resource, RestResourceProvider};
 
 use super::{OAuthClient, UserInfo};
 
-/// Saves a set of user info.
-pub trait SaveUserInfoOperation: Sync + Send {
+/// Operations that handle an OAuth user.
+pub trait OAuthUserInfoStore: Sync + Send {
     /// Executes a save operation on the given user info.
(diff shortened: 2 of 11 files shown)


libsplinter/src/auth/oauth/rest_api/mod.rs
```diff
@@ -23,13 +23,13 @@ use crate::rest_api::{Resource, RestResourceProvider};
 
 use super::{OAuthClient, UserInfo};
 
-/// Saves a set of user info.
-pub trait SaveUserInfoOperation: Sync + Send {
+/// Operations that handle an auth user.
+pub trait AuthUserOperation: Sync + Send {
```
ltseeley, line 27: Don't think I like this name; this trait isn't an operation, it's a storage API. It's also OAuth-specific. Maybe `OAuthUserStore`?

libsplinter/src/auth/oauth/rest_api/mod.rs
```diff
@@ -28,6 +28,12 @@ pub trait AuthUserOperation: Sync + Send {
     /// Executes a save operation on the given user info.
     fn save_user_info(&self, user_info: &UserInfo) -> Result<(), InternalError>;
 
+    /// Executes a `RemoveUserAccessOperation` on the user with the given access token.
+    ///
+    /// Returns an optional user identity, signifying the user was found in underlying storage
+    /// and successfully updated.
+    fn remove_user_access(&self, access_token: &str) -> Result<Option<String>, InternalError>;
```
ltseeley, line 35: Why does this return the user's identity? Is it used by anything? Seems like a bool would be more appropriate. I'd also rename this to `remove_user_access_token` to be more clear.

libsplinter/src/auth/oauth/rest_api/actix/logout.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::auth::oauth::rest_api::AuthUserOperation;
+use crate::protocol;
+use crate::rest_api::{
+    get_authorization_token, ErrorResponse, Method, ProtocolVersionRangeGuard, Resource,
+};
+
+pub fn make_logout_route(remove_access_op: Box<dyn AuthUserOperation>) -> Resource {
+    Resource::build("/oauth/logout")
+        .add_request_guard(ProtocolVersionRangeGuard::new(
+            protocol::OAUTH_LOGIN_MIN,
```
ltseeley, line 29: This should have it's own protocol version.

libsplinter/src/auth/oauth/rest_api/actix/logout.rs
```diff
@@ -0,0 +1,70 @@
...
+            protocol::OAUTH_PROTOCOL_VERSION,
+        ))
+        .add_method(Method::Get, move |req, _| {
+            // Extract access token from request's `Authorization` header
+            Box::new(match get_authorization_token(&req) {
```
ltseeley, line 34: Don't think we should be getting the header and parsing it here. IMO, we should update the auth middleware to insert the user's identity into the request's extensions and remove the token by OAuth identity. This would require updating the `auth::rest_api::authorize` function to check authorization for this endpoint.

libsplinter/src/auth/oauth/rest_api/mod.rs
```diff
@@ -75,6 +75,7 @@ impl AuthUserOperation for AuthUserOperationNoOp {
 ///
 /// * `GET /oauth/login` - Get the URL for requesting authorization from the provider
 /// * `GET /oauth/callback` - Receive the authorization code from the provider
+/// * `GET /oauth/logout` - Remove the user's access
```
ltseeley, line 78: s/access/access token/, also later in this file

libsplinter/src/auth/oauth/rest_api/mod.rs
peterschwarz: `OAuthUserInfoStore`, as `OauthUserStore` is already a biome trait.  This store also acts on UserInfo structs, so it is probably better to be more specific.

libsplinter/src/auth/oauth/rest_api/mod.rs
peterschwarz: Does it actually need to return anything? I.e. could it be `Result<(), InternalError>`?

shannynalayna: The issue I was running into, is the function to fetch the `OAuthUser` returns `Ok(None)` if it doesn't find the user. I assumed we would want to return a 401 in this case. So, since the `identity` field will always be there, if it is returned as `None`, that means the user doesn't exist and the `/oauth/logout` endpoint can return a 401.

shannynalayna: Updated how the access token is handled, as suggested in another comment, so this method now returns just `Result<(), InternalError>`.

libsplinter/src/auth/oauth/rest_api/actix/logout.rs
shannynalayna: Updated the middleware, as suggested. The endpoint now extracts the user's identity from the request extensions and passes that to the `remove_user_tokens` method, rather than the `access_token`

peterschwarz (dismissed):
Tested

peterschwarz (approved):

isabeltomb (dismissed):
tested - working properly

libsplinter/src/auth/rest_api/mod.rs
```diff
@@ -56,6 +56,11 @@ fn authorize(
     if endpoint.starts_with("/oauth") {
         is_auth_endpoint = true;
     }
+    #[cfg(feature = "oauth")]
+    if endpoint == "/oauth/logout" {
+        is_auth_endpoint = false;
+    }
+
```
ltseeley, line 63: Can you change this to just set the bool to `true` if the endpoint is `/oauth/login` or `/oauth/callback`, like the biome endpoints?

libsplinter/src/auth/oauth/rest_api/actix/logout.rs
```diff
@@ -0,0 +1,54 @@
...
+use actix_web::HttpResponse;
+use futures::future::IntoFuture;
+
+use crate::auth::oauth::rest_api::OAuthUserInfoStore;
+use crate::protocol;
+use crate::rest_api::{ErrorResponse, Method, ProtocolVersionRangeGuard, Resource};
+
+use crate::biome::rest_api::auth::OAuthUserIdentityRef;
+
+pub fn make_logout_route(user_info_store: Box<dyn OAuthUserInfoStore>) -> Resource {
```
ltseeley, line 26: The commit message says it updates the `authorize` to do auth on this endpoint, but that change is in the previous commit; it should probably be moved to this one.

ltseeley (changes requested):

libsplinter/src/auth/oauth/rest_api/actix/logout.rs
shannynalayna: Updated to add that change (updated based on other comment) to this commit

ltseeley (approved):

isabeltomb (approved):

peterschwarz (approved):
Tested.

PR #1044, opened 2020-11-23
Add experimental support for sqlite to `splinter database migrate`

agunde406:
If sqlite is enabled, the default database connection string
is now /var/lib/splinter/splinter_state.db. This commit
also adds support for changing this location to
SPLINTER_HOME/data/splinter_state.db and
STATE_DIR/splinter_state.db.

If the connection string is :memory: the sqlite migrations
will be run against an in memory sqlite database. This can
be used as "dry-run" test for sqlite migrations.

If sqlite is not enabled, the old default is returned:
postgres://admin:admin@localhost:5432/splinterd


cli/src/action/database.rs
```diff
@@ -12,37 +12,170 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use clap::ArgMatches;
+#[cfg(feature = "sqlite")]
```
peterschwarz, line 15: Could all this sqlite code move under its own module?  it would make this module much easier to read

cli/src/action/database.rs
```diff
@@ -12,37 +12,170 @@
...
+                })?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+fn get_default_database() -> Result<String, CliError> {
+    #[cfg(feature = "sqlite")]
```
peterschwarz, line 101: Instead of doing this as blocks in the function, just create two functions with the guards here - it will be easier to follow what happens.

peterschwarz (changes requested):

peterschwarz (dismissed):

cli/man/splinter-database-migrate.1.md
```diff
@@ -19,12 +19,11 @@ SYNOPSIS
 DESCRIPTION
 ===========
 
-Biome is the Splinter module that manages users, credentials, and private keys,
-using a PostgreSQL database to store this information.
+Splinter state can be stored in PostgreSQL database or a SQLite database.
```
vaporos, line 22: Probably meant "a PostgreSQL" here.

peterschwarz (approved):

vaporos (approved):

PR #1068, opened 2020-12-07
Add display name to circuit and proposed circuit

agunde406:
This PR adds an optional display name to circuit and proposed circuit. It also updates the version of the REST APIto version 2. If version 2 is requested the display name and service node id is returned (instead of allowed_nodes). If version 1 is requested the results are the same as before.


ltseeley (approved):
LGTM

vaporos (approved):

peterschwarz (approved):

PR #1070, opened 2020-12-07
Fix errors when splinterd is run with no default features

agunde406:

### DIFF
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -181,12 +181,21 @@ impl SplinterDaemon {
                     }
                 }
             } else {
-                // If storage is not provided, db_url is required
-                let db_url = self.db_url.clone().ok_or_else(|| {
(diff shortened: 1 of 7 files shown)


ltseeley (approved):

splinterd/src/daemon.rs
```diff
@@ -169,6 +169,7 @@ impl SplinterDaemon {
                                 })?,
                         )
                     }
+                    #[cfg(feature = "database")]
```
peterschwarz, line 172: Why is this one guarded by “database”?

agunde406: Because the only supported in "memory" implementation for admin service store is a sqlite database so it requires create_store_factory. Although looking at it closer the store factory should probably always been pulled in. ill update that.

ltseeley (approved):

peterschwarz (approved):

PR #1073, opened 2020-12-08
Modify OpenID OAuth identity provider

isabeltomb:
Modify the OpenID identity provider to return the 'sub' value from the json object returned by the userinfo_endpoint. This value, the subject identifier, is the unique identifier for End-User accounts. Google and Azure AD both use subject identifiers and all other OpenID providers should as well.

### DIFF
--- a/libsplinter/src/rest_api/auth/identity/openid.rs
+++ b/libsplinter/src/rest_api/auth/identity/openid.rs
@@ -56,12 +56,12 @@ impl IdentityProvider for OpenIdUserIdentityProvider {
             }
         }
 
-        let email = response
+        let user_identity = response
             .json::<UserResponse>()
             .map_err(|_| InternalError::with_message("Received unexpected response body".into()))?
-            .email;
+            .sub;
 
-        Ok(Some(email))
+        Ok(Some(user_identity))
     }
 
     fn clone_box(&self) -> Box<dyn IdentityProvider> {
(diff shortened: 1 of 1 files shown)


libsplinter/src/auth/rest_api/identity/openid.rs
```diff
@@ -56,12 +56,13 @@ impl IdentityProvider for OpenIdUserIdentityProvider {
...
 
-        let email = response
+        let user_info = response
             .json::<UserResponse>()
-            .map_err(|_| InternalError::with_message("Received unexpected response body".into()))?
-            .email;
+            .map_err(|_| InternalError::with_message("Received unexpected response body".into()))?;
 
-        Ok(Some(email))
+        let user_identity = user_info.clone().email.or(user_info.name);
```
peterschwarz, line 63: This clone doesn't seem necessary.  You should be able to destructure `user_info` and just use the `email` and `name` directly.

isabeltomb: Fixed, the `clone` was needed when `.or_else()` was used here but isn't with `.or()`

peterschwarz (approved):

ltseeley (approved):

ltseeley (approved):

peterschwarz (approved):

PR #1074, opened 2020-12-08
Inflight OAuth request store

peterschwarz:
This PR refactors the in-flight state saved between the initiation of an OAuth2 request and the receipt of the callback. It places the original memory implementation behind a trait, and updates the builders to take a trait implementation. 

Additionally, 

- Adding a new parameter create a clippy error with the OAuthClient constructor. As this is a private method, it has been decomposed into two functions

### DIFF
--- a/libsplinter/src/oauth/mod.rs
+++ b/libsplinter/src/oauth/mod.rs
@@ -18,8 +18,8 @@ mod builder;
 mod error;
 #[cfg(feature = "rest-api")]
 pub mod rest_api;
+pub mod store;
 
-use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use oauth2::{
@@ -27,7 +27,6 @@ use oauth2::{
     CsrfToken, PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, Scope, TokenResponse, TokenUrl,
 };
 
-use crate::collections::TtlMap;
 use crate::error::{InternalError, InvalidArgumentError};
 use crate::rest_api::auth::identity::{Authorization, BearerToken, IdentityProvider};
 
@@ -38,9 +37,6 @@ pub use builder::OAuthClientBuilder;
 pub use builder::OpenIdOAuthClientBuilder;
 pub use error::OAuthClientBuildError;
 
-/// The amount of time before a pending authorization expires and a new request must be made
-const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
-
 /// An OAuth2 client for Splinter
 ///
 /// This client currently supports OAuth2 authorization code grants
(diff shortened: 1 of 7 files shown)


libsplinter/src/auth/oauth/builder/mod.rs
```diff
@@ -144,4 +154,14 @@ impl OAuthClientBuilder {
         self.identity_provider = Some(identity_provider);
         self
     }
+
+    /// Sets the in-flight request store to use to store values between requests to and from the
```
agunde406, line 154: I think "to use" should be removed from this comment?

agunde406: same for other instance of with_inflight_request_store

peterschwarz: Reworded in all places.

agunde406 (approved):

agunde406 (dismissed):

libsplinter/src/oauth/mod.rs
```diff
@@ -49,13 +45,14 @@ const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
...
     client: BasicClient,
-    /// Pending authorization requests, including the CSRF token, PKCE verifier, and client's
-    /// redirect URL
-    pending_authorizations: Arc<Mutex<TtlMap<String, PendingAuthorization>>>,
     /// The scopes that will be requested for each user that's authenticated
     scopes: Vec<String>,
     /// OAuth2 identity provider used to retrieve users' identities
     identity_provider: Box<dyn IdentityProvider>,
+
+    /// Store for pending authorization requests , including the CSRF token, PKCE verifier, and client's
```
ltseeley, line 53: There's an extra space in this line

libsplinter/src/oauth/store/memory.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::time::Duration;
+
+use crate::oauth::{InflightOAuthRequestStore, PendingAuthorization};
+use crate::collections::TtlMap;
+use crate::error::InternalError;
+
+/// The amount of time before a pending authorization expires and a new request must be made
+const PENDING_AUTHORIZATION_EXPIRATION_SECS: u64 = 3600; // 1 hour
+
+#[derive(Clone)]
```
ltseeley, line 27: Needs a doc string to provide some information about this implementation, especially around the timed expiration

libsplinter/src/oauth/store/mod.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2018-2020 Cargill Incorporated
```
ltseeley, line 1: We're planning on having more than one store that belongs to OAuth, right? Probably want this module to have a more specific name.

libsplinter/src/oauth/mod.rs
```diff
@@ -74,30 +74,11 @@ impl OAuthClient {
     ///
     /// Returns an error if any of the auth, redirect, or token URLs are invalid
     fn new(
-        client_id: String,
```
ltseeley, line 80: Doc string for this method needs to be updated

libsplinter/src/oauth/store/mod.rs
peterschwarz: We can rearrange them when we create the next one

peterschwarz: Or, there should be one OAuth store

ltseeley (dismissed):

ltseeley (approved):

agunde406 (approved):

PR #1076, opened 2020-12-09
`AdminServiceEventStore` trait and memory implementation

shannynalayna:
Adds the `AdminServiceEventStore` trait and in-memory implementation behind the experimental `admin-service-event-store` trait. To test the in-memory implementation, run `cargo test --features experimental event_store`.

### DIFF
--- a/libsplinter/src/admin/rest_api/actix/ws_register_type.rs
+++ b/libsplinter/src/admin/rest_api/actix/ws_register_type.rs
@@ -15,6 +15,8 @@
 use actix_web::{web, HttpResponse};
 use futures::IntoFuture;
 use std::collections::HashMap;
+#[cfg(feature = "admin-service-event-store")]
+use std::convert::TryInto;
 use std::time;
 
 use crate::admin::messages::AdminServiceEvent;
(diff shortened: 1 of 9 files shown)


libsplinter/src/admin/service/event/store/memory.rs
```diff
@@ -0,0 +1,319 @@
...
+use std::sync::{Arc, Mutex};
+use std::time::SystemTime;
+
+use crate::admin::service::event::store::{AdminServiceEventStore, AdminServiceEventStoreError};
+use crate::admin::service::mailbox::EventEntry;
+use crate::admin::service::messages::AdminServiceEvent;
+use crate::error::InternalError;
+
+#[derive(Default, Clone)]
+pub struct MemoryAdminServiceEventStore {
```
agunde406, line 32: Can you add documentation comments for MemoryAdminServiceEventStore

libsplinter/src/admin/service/mailbox.rs
```diff
@@ -26,8 +26,8 @@ use super::messages::AdminServiceEvent;
 /// A simple entry for AdminServiceEvent values, marked with a timestamp
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub struct EventEntry {
```
peterschwarz, line 28: This is an internal implementation detail of the memory store. It should actually be moved there.

peterschwarz (changes requested):
This pr should include the usage of the store. As you've added the features, you can either use the old way on stable, or the new way in experimental.

libsplinter/src/admin/service/event/store/memory.rs
shannynalayna: Updated!

libsplinter/src/admin/service/mailbox.rs
shannynalayna: Moved this struct into the `memory` module

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
```diff
@@ -61,34 +63,68 @@ pub fn make_application_handler_registration_route<A: AdminCommands + Clone + 's
...
+                        return Box::new(
+                            HttpResponse::InternalServerError().finish().into_future(),
+                        );
+                    }
                 }
             };
 
             let request = Request::from((request, payload));
-            match new_websocket_event_sender(request, Box::new(initial_events.skip(skip))) {
+            match new_websocket_event_sender(request, Box::new(initial_events)) {
```
peterschwarz, line 127: Why was this skip removed?

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
```diff
@@ -134,28 +189,47 @@ impl AdminServiceEventSubscriber for WsAdminServiceEventSubscriber {
...
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_id: Option<i64>,
 }
 
 impl From<(time::SystemTime, AdminServiceEvent)> for JsonAdminEvent {
     fn from(raw_evt: (time::SystemTime, AdminServiceEvent)) -> Self {
         Self {
-            timestamp: raw_evt.0,
+            timestamp: Some(raw_evt.0),
```
peterschwarz, line 205: These fields should actually both be set to the event id value, for backwards compatibility.

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
shannynalayna: The skip wasn't removed, I just moved it up to use `skip` where `initial_events` is being returned.

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
shannynalayna: This conversion is used by the `Mailbox`, so the timestamp is being used rather than the event ID

peterschwarz: But this is the JsonAdminEvent, which is what is sent out of the websocket to external subscribers

shannynalayna: The event ID isn't being produced when the `Mailbox` is being used, though.

shannynalayna: Perhaps this implementation should also be behind feature flags so this is obvious? The conversion happening below (`impl From<(i64, AdminServiceEvent)> for JsonAdminEvent`) is what is used when the event store is being used.

shannynalayna: Ok, I've updated this now to set the value of the `timestamp` to the current time (Since the field is typed as `SystemTime`), unless there is a different way you would suggest working around that type to set the `timestamp` to the event ID value.

agunde406 (approved):

peterschwarz (approved):

PR #1077, opened 2020-12-10
Implement Diesel-backed InflightOAuthRequestStore

peterschwarz:


libsplinter/src/oauth/store/mod.rs
```diff
@@ -16,4 +16,81 @@
 
 mod memory;
 
+use crate::error::InternalError;
```
ltseeley: The module doc comment should be updated

libsplinter/src/migrations/diesel/sqlite/migrations/2020-12-08-121200-oauth-create-inflight-requests/up.sql
```diff
@@ -0,0 +1,21 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+-- -----------------------------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS oauth_inflight_request (
+    id                  TEXT            PRIMARY KEY,
+    pkce_verifier       TEXT            NOT NULL,
+    client_redirect_url TEXT            NOT NULL,
+    last_updated        TIMESTAMP       DEFAULT CURRENT_TIMESTAMP NOT NULL
```
ltseeley, line 20: Would `created` be more accurate, since these entries never really get updated?

libsplinter/src/oauth/store/diesel/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn clone_box(&self) -> Box<dyn InflightOAuthRequestStore> {
+        Box::new(Self {
+            connection_pool: self.connection_pool.clone(),
+        })
+    }
+}
+
+impl From<models::OAuthInflightRequest> for PendingAuthorization {
```
ltseeley, line 128: In other stores we've put all `From` implementations in the store's top-level error module to keep all error boilerplate in one place.

libsplinter/src/rest_api/mod.rs
```diff
@@ -859,22 +859,28 @@ impl RestApiBuilder {
                             ));
                         }
 
+                        #[cfg(feature = "biome-oauth")]
```
ltseeley, line 862: This change seems like it should be in its own commit.

libsplinter/src/oauth/store/diesel/mod.rs
peterschwarz: This isn't an error conversion, it converts the model struct to the one the oauth client uses.

libsplinter/src/rest_api/mod.rs
peterschwarz: No, this change was necessary because the ownership changed (see the above `into_iter` change).  The feature guard is due to the fact that it was previously only used inside a later block that was guarded by that feature.

libsplinter/src/oauth/store/diesel/mod.rs
ltseeley: Oops, meant to put this on the error conversions below

libsplinter/src/migrations/diesel/sqlite/migrations/2020-12-08-121200-oauth-create-inflight-requests/up.sql
peterschwarz: renamed

libsplinter/src/oauth/store/mod.rs
peterschwarz: updated

ltseeley (approved):

agunde406 (approved):

PR #1087, opened 2020-12-17
Implement Add subcommand for node registry CLI

isabeltomb:
Add the `add` subcommand including its options and flags to the splinter registry subcommand in splinter cli. 

Add the functionality for adding a node to the registry using the `splinter registry add` command. This includes the addition of a `registry::api` module to handle REST API requests. This is behind an experimental feature `registry-add-node`.

Implement std::fmt::Display for the registry Node struct to allow it to be printed after successfully adding a node to the registry.

## **Testing:**

use keygen to create a private key and add it to the allow_keys file:

from the `splinter/cli` directory run:
```
$ cargo run keygen
```

Start splinterd locally

In a separate terminal window, from the `splinter/cli` directory, try the following `splinter registry add` commands:

`$ cargo run --features experimental registry add node1 --key-file <public key file> --key <private key file> --endpoint tcps://splinterd-node-1:8001 --metadata key1:value1`
(the private key should be the same as the one added to `allow_keys`)

This should successfully add a node named "node1" to the local registry

`$ cargo run --features experimental registry add node1 --key-file <public key file> --key <private key file> --endpoint tcps://splinterd-node-1:8002`

This should fail because a node with name "node1" already exists in the registry

`$ cargo run --features experimental registry add node2 --key-file <public key file> --key <private key file> --endpoint tcps://splinterd-node-1:8001`

This should fail because a node with endpoint "tcps://splinterd-node-1:8001" already exists in the registry

`$ cargo run --features experimental registry add node2 --key-file <public key file> --key <private key file> --endpoint tcps://splinterd-node-1:8002 --metadata key1:`

This should fail because the metadata key is missing a value

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -74,12 +74,15 @@ experimental = [
     "health",
     "https-certs",
     "permissions",
+    "registry",
 ]
 
 authorization-handler-maintenance = []
 authorization-handler-rbac = []
 circuit-auth-type = []
 circuit-template = ["splinter/circuit-template"]
+
+registry = []
 
 health = []
 
--- a/cli/man/splinter-registry-add.1.md
+++ b/cli/man/splinter-registry-add.1.md
@@ -1,0 +1,133 @@
+% SPLINTER-REGISTRY-ADD(1) Cargill, Incorporated | Splinter Commands
+<!--
+  Copyright 2018-2021 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**splinter-registry-add** — Add a node to the local registry
+
+SYNOPSIS
+========
+
+**splinter registry add** \[**FLAGS**\] \[**OPTIONS**\] IDENTITY
+
+DESCRIPTION
+===========
+
+Add a new node to the local node registry. The node may be entirely new to the
+registry, or it may be copied from the remote registries with the
+`--from-remote` flag. If the `--from-remote` flag is used the `--display-name`,
+`--endpoint`, `--key` and `--metadata` options may not be used to alter the
+node being copied from the remote registry. When run, the command will
+display the resulting changes as confirmation.
+
+FLAGS
+=====
+`--dry-run`
+: Shows the expected changes without submitting the node.
+
+`--from-remote`
+: Copies an existing node definition from the remote registries. 
+
+`-h`, `--help`
+: Prints help information
+
+`-q`, `--quiet`
+: Decreases verbosity (the opposite of -v). When specified, only errors or
+  warnings will be output.
+
+`-V`, `--version`
+: Prints version information
+
+`-v`
+: Increases verbosity (the opposite of -q). Specify multiple times for more
+  output.
+
+OPTIONS
+=======
+
+`--display-name` DISPLAY_NAME
+: Sets a human-readable name for the new node. If not provided, a default value
+based on the node's ID will be used.
+
+`--endpoint ENDPOINT`
+: Adds a network endpoint for the new node. At least one endpoint must be
+provided, and all endpoints must be non-empty and unique in the registry (two
+nodes cannot share the same endpoint). Repeat this option to specify multiple
+endpoints.
+
+`--key-file KEY`
+: Add the public key to the new node. At least one key must be provided, and all
+keys must be non-empty. Repeat this option to specify multiple keys.
+
+`-k`, `--key KEY`
+: Name or path of private key to be used for REST API authorization.
+
+`--metadata METADATA_STRING`
+: Adds the metadata to the new node, using the format
+`METADATA_KEY:METADATA_VALUE`. If an entry for the given `METADATA_KEY` already
+exists, it will be replaced. Repeat this option to specify multiple metadata
+entries.
+
+`-U`, `--url URL`
+: Specifies the URL for the `splinterd` REST API. The URL is required unless
+  `$SPLINTER_REST_API_URL` is set.
+
+ARGUMENTS
+=========
+
+`IDENTITY`
+Identity of the new node. Must be unique in the local registry.
+
+EXAMPLES
+========
+
+The simplest use of this command is to create a new node with an identity, a
+single endpoint, and a single key:
+
+```
+splinter registry add example-node-1 \
+  --endpoint tcps://splinterd-node-1:8044 \
+  --key /path/to/public/key/file \
+  --url http://splinterd-rest-api:8085
+```
+
+Multiple endpoints, keys, and metadata entries can be provided by specifying the
+arguments multiple times:
+
+```
+splinter registry add example-node-2 \
+  --endpoint tcps://splinterd-node-2:8044 \
+  --endpoint tcp://splinterd-node-2:8045 \
+  --key /path/to/public/key/file1 \
+  --key /path/to/public/key/file2 \
+  --metadata key1:value1 \
+  --metadata key2:value2 \
+  --url http://splinterd-rest-api:8085
+```
+
+A node that exists in one or more remote registries can be copied to the local
+registry with just the node's identity and the `--from-remote` flag, the
+`--display-name`, `--endpoint`, `--key` and `--metadata` options may not be used
+with this flag:
+
+```
+splinter registry add example-node-3 \
+  --from-remote \
+  --url http://splinterd-rest-api:8085
+```
+
+ENVIRONMENT VARIABLES
+=====================
+
+**SPLINTER_REST_API_URL**
+: URL for the `splinterd` REST API. (See `-U`, `--url`.)
+
+SEE ALSO
+========
+
+| Splinter documentation: https://www.splinter.dev/docs/0.5/
--- a/cli/src/action/registry/api.rs
+++ b/cli/src/action/registry/api.rs
@@ -1,0 +1,121 @@
+// Copyright 2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt;
+
+use reqwest::blocking::Client;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+use crate::action::api::{ServerError, SplinterRestClient};
+use crate::error::CliError;
+
+impl SplinterRestClient {
+    /// Adds a new node to the registry.
+    pub fn add_node(&self, node: &RegistryNode) -> Result<(), CliError> {
+        let request = Client::new()
+            .post(&format!("{}/registry/nodes", self.url))
+            .json(&node)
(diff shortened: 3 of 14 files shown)


cli/src/main.rs
```diff
@@ -580,53 +580,118 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
+                ),
+        );
+
+    #[cfg(feature = "registry-add-node")]
+    let registry_command = registry_command.subcommand(
+        SubCommand::with_name("add")
+            .about("Add a node to the local registry")
+            .arg(
+                Arg::with_name("display_name")
+                    .long("display")
```
ltseeley, line 634: This should be `display-name`

cli/src/main.rs
```diff
@@ -580,53 +580,118 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
+                    .multiple(true)
+                    .required_unless("from_remote")
+                    .help("Path of public key file to include with node"),
+            )
+            .arg(
+                Arg::with_name("metadata")
+                    .long("metadata")
+                    .takes_value(true)
+                    .multiple(true)
+                    .help("Metadata to include with node (<key>=<value>)"),
```
ltseeley, line 674: Should be `<key>:<value>`

cli/src/main.rs
```diff
@@ -580,53 +580,118 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
+            )
+            .arg(
+                Arg::with_name("private_key_file")
+                    .value_name("private-key-file")
+                    .short("k")
+                    .long("key")
+                    .takes_value(true)
+                    .help("Name or path of private key"),
```
ltseeley, line 682: Might want to make this help text a bit more explicit so it's clear that this doesn't get attached to the node in any way.

cli/Cargo.toml
```diff
@@ -74,13 +74,15 @@ experimental = [
...
     "health",
+    "registry-add-node",
     "splinter-cli-jwt",
 ]
 
 circuit-auth-type = []
 circuit-template = ["splinter/circuit-template"]
 
 splinter-cli-jwt = ["cylinder/jwt", "cylinder/key-load"]
+registry-add-node=[]
```
ltseeley, line 85: Maybe we can just do `registry` and put all the experimental registry commands under this feature instead of having a separate feature for each; that way we can stabilize them all at once. Also, make sure this is formatted consistently (spaces around the `=`).

cli/src/action/registry/api.rs
```diff
@@ -0,0 +1,153 @@
...
+use serde::Deserialize;
+use splinter::registry::Node;
+use std::collections::HashMap;
+
+use crate::action::api::{ServerError, SplinterRestClient};
+use crate::error::CliError;
+
+impl SplinterRestClient {
+    /// Adds a new node to the registry.
+    pub fn add_node(&self, node: &Node) -> Result<(), CliError> {
```
ltseeley, line 25: We should avoid using Splinter stores' internal structs, which may differ in format from what the CLI sends or the REST API accepts. In fact, this shouldn't work because the `Node` struct shouldn't have the `Serialize` derivation that it does (that's just a result of it being implemented before we started enforcing these guidelines). We should define a struct for this client to represent a node for submission; could probably just re-purpose `NodeResponse`.

cli/src/action/registry/api.rs
```diff
@@ -0,0 +1,153 @@
...
+                }
+            })
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct NodeResponse {
+    pub identity: String,
+    pub endpoints: Vec<String>,
+    pub display_name: Option<String>,
```
ltseeley, line 150: I think the display name is always set in the response.

cli/src/action/registry/mod.rs
```diff
@@ -177,3 +182,161 @@ impl Action for RegistryGenerateAction {
...
+
+        let identity = args
+            .value_of("identity")
+            .ok_or_else(|| CliError::ActionError("Identity must be specified".into()))?
+            .to_string();
+
+        let display_name = args
+            .value_of("display_name")
+            .unwrap_or(&identity)
+            .to_string();
```
ltseeley, line 208: I don't expect this change in this PR, but at some point we should probably update the registry REST API to accept a new node without the display name field set; the REST API would then generate a default value for this field internally. Can you create a doc for this feature add this as a change before we stabilize?

cli/src/action/registry/mod.rs
```diff
@@ -177,3 +182,161 @@ impl Action for RegistryGenerateAction {
...
+            let private_key = args.value_of("private_key_file");
+            builder = builder.with_auth(create_cylinder_jwt_auth(private_key)?);
+        }
+
+        let client = builder.build()?;
+
+        let metadata: Vec<(String, String)> = if let Some(metadata) = args.values_of("metadata") {
+            metadata
+                .map(|kv| {
+                    let mut kv_iter = kv.splitn(2, '=');
```
ltseeley, line 224: We've had some discussions recently about how to handle parsing like this and we settled on a slightly different pattern that's a bit more clear and explicit: https://github.com/Cargill/splinter/blob/master/libsplinter/src/rest_api/auth/identity/mod.rs#L113

cli/src/main.rs
```diff
@@ -580,53 +580,118 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
+            .arg(
+                Arg::with_name("endpoint")
+                    .long("endpoint")
+                    .takes_value(true)
+                    .multiple(true)
+                    .required_unless("from_remote")
+                    .help("Network endpoint for the new node"),
+            )
+            .arg(
+                Arg::with_name("from_remote")
```
ltseeley, line 652: I actually think that all the options for modifying the node (display name, endpoints, keys, metadata) should not be allowed when using `--from-remote`. If you take a look at the design we have for the update command, the ways in which you can modify these things is pretty complicated; the add command doesn't support removing or replacing some of these values. So IMO we're better off disabling any modifications when copying the node from remote to local.

cli/src/action/registry/mod.rs
```diff
@@ -177,3 +182,161 @@ impl Action for RegistryGenerateAction {
...
+            for (key, value) in node_metadata {
+                node_builder = node_builder.with_metadata(key, value);
+            }
+
+            let node = node_builder
+                .build()
+                .map_err(|err| CliError::ActionError(format!("Unable to build node: {}", err)))?;
+
+            if !args.is_present("dry_run") {
+                client.update_node(&node)?;
```
ltseeley, line 297: I think we discussed how this should use `add_node` but can't because of the way the registry and its REST API are currently designed; the changes we discussed to correct this should be put in the feature doc for the registry CLI.

ltseeley (changes requested):

cli/src/main.rs
isabeltomb: fixed

cli/src/main.rs
isabeltomb: changed to `<key>:<value>`

cli/src/main.rs
isabeltomb: Updated

cli/Cargo.toml
isabeltomb: changed to just `registry`

cli/src/action/registry/api.rs
isabeltomb: Changed to use the `RegistryNode` struct

cli/src/action/registry/api.rs
isabeltomb: fixed

cli/src/action/registry/mod.rs
isabeltomb: Created a doc with this comment, will add it to the feature stabilization list

cli/src/action/registry/mod.rs
isabeltomb: Fixed

cli/src/main.rs
isabeltomb: if `--from-remote` is used now the node will be copied from the remote registry without changing any information about it

cli/src/action/registry/mod.rs
isabeltomb: Added the commit with these changes

cli/src/main.rs
ltseeley: To make this clear and explicit to the user, you should add `.conflicts_with_all(&["display_name", "endpoint", "identity", "key_files", "metadata"])`. This will actively prevent users from trying to change the node when using `--from-remote`. The man page should describe this behavior.

cli/src/action/registry/api.rs
ltseeley: `RegistryNode` is good, but it should be defined in the API module. As it is currently, there's a circular dependency: the API module depends on the action module for `RegistryNode`, and the action module depends on the API. Conceptually, the `RegistryNode` struct is a part of the API (it's accepted and returned by its methods).

cli/src/action/registry/mod.rs
```diff
@@ -172,3 +181,149 @@ impl Action for RegistryGenerateAction {
...
+            Ok(())
+        }
+    }
+}
+
+#[cfg(feature = "registry")]
+fn parse_metadata(metadata: &str) -> Result<(String, String), CliError> {
+    let mut parts = metadata.splitn(2, ':');
+    match (parts.next(), parts.next()) {
+        (Some(key), Some(value)) => match key {
```
ltseeley, line 288: I belive `value` could be an empty string as well if the metadata arg is `key:`.

isabeltomb: added a match to check for that case

cli/src/main.rs
isabeltomb: Fixed, also added a man page for this subcommand that includes a note about this in the description

cli/src/action/registry/api.rs
isabeltomb: Moved to the API module

cli/src/main.rs
```diff
@@ -659,53 +659,125 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
+                    .long("from-remote")
+                    .conflicts_with_all(&[
+                        "display_name",
+                        "endpoint",
+                        "identity",
```
ltseeley, line 736: `identity` is still required with `--from-remote` since it determines which node to copy. I just realized that I said to include this in my comment above, so that's 100% my mistake.

isabeltomb: Fixed

ltseeley (approved):

cli/man/splinter-registry-add.1.md
```diff
@@ -0,0 +1,128 @@
...
+`--key-file KEY`
+: Add the public key to the new node. At least one key must be provided, and all
+keys must be non-empty. Repeat this option to specify multiple keys.
+
+`-k`, `--key KEY`
+: Name or path of private key to be used for REST API authorization.
+
+`--metadata METADATA_STRING`
+: Adds the metadata to the new node, using the format
+`METADATA_KEY:METADATA_VALUE`. If an entry for the given `METADATA_KEY` already exists, it will be replaced. Repeat this option to specify multiple metadata
```
agunde406, line 71: This line is too long, should be wrapped

agunde406 (changes requested):

cli/man/splinter-registry-add.1.md
isabeltomb: Fixed

agunde406 (approved):
tested

ltseeley (approved):

PR #1088, opened 2020-12-18
Remove raw SQL from the DieselRegistry

agunde406:
Using raw SQL queries is error prone and makes the database
vulnerable to SQL injection attack. These issues are fixed
by instead using diesel queries, which can be dynamically build
just as easily as a string.

This PR also add sqlite unit tests for the DieselRegistry

### DIFF
--- a/libsplinter/src/registry/diesel/operations/count_nodes.rs
+++ b/libsplinter/src/registry/diesel/operations/count_nodes.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 //! Provides the "count nodes" operation for the `DieselRegistry`.
+use std::convert::TryFrom;
 
 use diesel::{dsl::count_star, prelude::*};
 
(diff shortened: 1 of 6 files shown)


peterschwarz (changes requested):

libsplinter/src/registry/diesel/operations/count_nodes.rs
```diff
@@ -47,23 +44,20 @@ where
...
-                )
-            })
+            let mut query = splinter_nodes::table
+                .into_boxed()
+                .select(splinter_nodes::all_columns);
+            query = apply_predicate_filters(query, predicates);
+            query
+                .select(count_star())
+                .first::<i64>(self.conn)
+                .map(|count| count as u32)
```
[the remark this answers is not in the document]
agunde406: Okay, Ill also update the old code above that does the same thing.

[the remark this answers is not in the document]
agunde406: I added a separate commit for the change as this was old code and present in another location.

libsplinter/src/registry/diesel/operations/count_nodes.rs
```diff
@@ -32,32 +33,41 @@ where
...
                 .map_err(|err| {
                     RegistryError::general_error_with_source(
                         "Failed to count all nodes",
                         Box::new(err),
                     )
-                })
+                })?;
+
+            Ok(u32::try_from(count).map_err(|err| {
+                RegistryError::general_error_with_source("Failed to count all nodes", Box::new(err))
```
peterschwarz, line 48: As this is a pretty unlikely error, I think the message should be fairly specific. Something like "The number of nodes is larger than the max u32" (since count can't be negative), and we can omit the source, since I think the error type is uninteresting.

agunde406: updated

agunde406: updated

peterschwarz (approved):

ltseeley (approved):

PR #1091, opened 2020-12-21
AdminServiceEventStore database models & migrations

shannynalayna:
Adds the models, schema, and PostgreSQL and SQLite migration data to support the `AdminServiceEventStore`. 

Testing:
To run the SQLite migration test, run

```
$ cargo test --features experimental test_event_store_migrations
```

Otherwise, to test the PostgreSQL migrations, create a local postgres instance:

```
```

Then, run the migrations:

```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- database migrate -C postgres://postgres:mypassword@localhost:5432/splinter
```

These should complete successfully. To verify these migrations have been applied, docker exec into the postgres container created above:

```
telander@branon splinter % docker run -it --rm --link postgres:postgres postgres psql -h postgres -U postgres
Password for user postgres: 
psql (13.0 (Debian 13.0-1.pgdg100+1))
Type "help" for help.

postgres=# \c splinter
You are now connected to database "splinter" as user "postgres".
splinter=# select * from admin_service_event;
 id | event_type | data 
----+------------+------
(0 rows)
```

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -96,6 +96,7 @@ experimental = [
     "stable",
     # The following features are experimental:
     "admin-service-event-store",
+    "admin-service-event-store-diesel",
     "auth",
     "biome-notifications",
     "biome-oauth",
@@ -118,6 +119,7 @@ benchmark = []
 
 admin-service = []
 admin-service-event-store = ["admin-service"]
+admin-service-event-store-diesel = ["diesel", "admin-service-event-store"]
 auth = []
 biome-credentials = ["bcrypt"]
 biome-key-management = []
--- a/libsplinter/src/admin/service/event/store/diesel/mod.rs
+++ b/libsplinter/src/admin/service/event/store/diesel/mod.rs
@@ -1,0 +1,19 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Database backend support for the `AdminServiceEventStore`, powered by
+//! [`Diesel`](https://crates.io/crates/diesel).
+
+mod models;
+mod schema;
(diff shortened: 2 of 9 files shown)


libsplinter/Cargo.toml
```diff
@@ -110,6 +110,7 @@ experimental = [
     "registry-database",
     "service-arg-validation",
     "service-network",
+    "sqlite-admin-service-event-store",
```
peterschwarz, line 113: In other uses of guarded implementations, we've used the main feature, and then added the subtype.  So this should probably be: `admin-service-event-store-sqlite`.

Though, in the sqlite case, we haven't been guarded that on its own, as it can be unit tested, unlike the postgres-backed store.

peterschwarz (changes requested):
The commit summary line should for the migrations be "Add AdminServiceEventStore sqlite migrations`

libsplinter/Cargo.toml
shannynalayna: I changed the other feature added to `admin-service-event-store-diesel` and put the sqlite test (originally behind `sqlite-admin-service-event-store`) is now guarded by: `#[cfg(all(test, feature = "admin-service-event-store-diesel", feature = "sqlite"))]`

peterschwarz (dismissed):

libsplinter/src/admin/service/event/store/diesel/models.rs
```diff
@@ -0,0 +1,337 @@
...
+    fn from((event_id, create_circuit): (i64, &CreateCircuit)) -> Self {
+        AdminEventProposedCircuitModel {
+            event_id,
+            circuit_id: create_circuit.circuit_id.to_string(),
+            authorization_type: String::from(&create_circuit.authorization_type),
+            persistence: String::from(&create_circuit.persistence),
+            durability: String::from(&create_circuit.durability),
+            routes: String::from(&create_circuit.routes),
+            circuit_management_type: create_circuit.circuit_management_type.to_string(),
+            application_metadata: Some(create_circuit.application_metadata.to_vec()),
```
agunde406, line 102: This should check if the application metadata is empty set None if it is.

libsplinter/src/admin/service/event/store/diesel/mod.rs
```diff
@@ -17,3 +17,35 @@
...
+
+        run_sqlite_migrations(&*pool.get().expect("Failed to get connection for migrations"))
+            .expect("Failed to run migrations");
+
+        pool
+    }
+
+    #[test]
+    /// Test that the sqlite migrations can be run successfully
+    fn test_event_store_migrations() {
```
agunde406, line 48: Technically we don't need to add this test for every store because all migrations always get run.

agunde406 (changes requested):

agunde406 (approved):

peterschwarz (approved):

PR #1094, opened 2020-12-22
Generate REST API certs for https

peterschwarz:
This PR generates the certs for the splinter REST API in https mode.

```
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml --features https-certs -- cert generate
```
will generate two new files, `rest_api.crt` and `rest_api.key` in the appropriate directories under the specified `SPLINTER_HOME`

### DIFF
--- a/cli/src/action/certs.rs
+++ b/cli/src/action/certs.rs
@@ -51,14 +51,22 @@ const SERVER_KEY: &str = "server.key";
 const CA_CERT: &str = "generated_ca.pem";
 const CA_KEY: &str = "generated_ca.key";
 
+#[cfg(feature = "https-certs")]
+const REST_API_CERT: &str = "rest_api.crt";
+#[cfg(feature = "https-certs")]
+const REST_API_KEY: &str = "rest_api.key";
+
 impl Action for CertGenAction {
     fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
         let args = arg_matches.ok_or(CliError::RequiresArgs)?;
 
-        let common_name = args
-            .value_of("common_name")
-            .unwrap_or("localhost")
-            .to_string();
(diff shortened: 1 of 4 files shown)


cli/src/action/certs.rs
```diff
@@ -231,21 +236,38 @@ fn handle_skip(
...
         && server_cert_path.exists()
         && server_key_path.exists()
         && ca_cert_path.exists()
         && ca_key_path.exists()
-    {
+        && rest_api_cert_path.exists()
+        && rest_api_key_path.exists();
+
+    // if all exists, log existence and return
+    if all_exist {
```
vaporos, line 270: If we hit this block, it won't print out that the REST API certs already exist. Some with the block below - which is inconsistent with how the CLI is supposed to work.

I'm not sure why we have this all_exist check at all though, since each individual check prints it out.

cli/src/action/certs.rs
```diff
@@ -424,6 +466,45 @@ fn handle_skip(
             return Err(CliError::ActionError("CA does not exist".into()));
         }
     }
+
+    #[cfg(feature = "https-certs")]
+    if rest_api_cert_path.exists() && rest_api_key_path.exists() {
+        info!(
+            "Server certificate exists, skipping: {}",
```
vaporos, line 473: This looks like a mismatch - was "REST API certificate exists" intended?

vaporos (changes requested):

cli/src/action/certs.rs
peterschwarz: Fixed.

cli/src/action/certs.rs
peterschwarz: Added a commit to remove this redundant check.

vaporos (approved):

agunde406 (approved):
tested

PR #1096, opened 2020-12-22
Add re-authentication to OAuth identity provider

ltseeley:
Updates the `OAuthUserIdentityProvider` to re-authenticate users'
sessions when a configurable amount of time has elapsed since the last
time the session was authenticated with the OAuth server.

### DIFF
--- a/libsplinter/src/biome/oauth/store/diesel/mod.rs
+++ b/libsplinter/src/biome/oauth/store/diesel/mod.rs
@@ -233,13 +233,19 @@ pub mod tests {
         oauth_user_session_store
             .add_session(session)
             .expect("Unable to add session");
+        let session = oauth_user_session_store
+            .get_session(splinter_access_token)
+            .expect("Unable to get session")
+            .expect("Session not found");
+        let originally_authenticated = session.last_authenticated();
+
+        // Wait for a few seconds so we can check that the timestamp changes (graunularity of the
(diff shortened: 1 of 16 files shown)


libsplinter/src/biome/oauth/store/diesel/operations/update_session.rs
```diff
@@ -57,22 +61,71 @@ where
...
                     } else {
-                        // All checks have completed, update the entry
-                        update(
-                            oauth_user_sessions::table.filter(
-                                oauth_user_sessions::splinter_access_token
-                                    .eq(session.splinter_access_token()),
+                        // All checks have completed; remove the old entry and add the updated one.
+                        // We can't just update the existing entry because Diesel doesn't provide a
+                        // way to set the "last_authenticated" column to the default value
+                        // (database-assigned timestamp).
```
vaporos, line 67: Should use a trigger instead.

vaporos (changes requested):

libsplinter/src/biome/oauth/store/diesel/operations/update_session.rs
ltseeley: Replaced the code change with migrations for adding a trigger.

vaporos (approved):

peterschwarz (approved):

PR #1097, opened 2020-12-22
Stabilize auth feature

peterschwarz:
This change stabilizes the "auth" feature by removing it.  It is an internal infrastructural feature, and therefore is only required by other features.


peterschwarz:
Closing in favor of #1138

PR #1100, opened 2020-12-23
Add event store's `AdminServiceEvent` object, update store to use it

shannynalayna:
Makes a quick update to the `AdminServiceEventStore` to use a newly defined `AdminServiceEvent` that accurately represents the `CircuitProposal` as it is defined across the admin service. The definition of the `CircuitProposal` from the admin::service::messages::AdminServiceEvent did not match this definition, as the `SplinterService` had an `allowed_nodes` field now represented as just a `node_id` field in the admin service `ProposedService` struct (equivalent to the message's `SplinterService`)

### DIFF
--- a/libsplinter/src/admin/service/event/mod.rs
+++ b/libsplinter/src/admin/service/event/mod.rs
@@ -12,5 +12,77 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Module containing the `AdminServiceEventStore` trait and related implementations.
+//! Defines an `AdminServiceEvent`, stored and returned by the `AdminServiceEventStore`.
+//!
+//! The public interface includes the struct [`AdminServiceEvent`].
+//!
+//! [`AdminServiceEvent`]: struct.AdminServiceEvent.html
+
 pub mod store;
+
+use std::convert::TryFrom;
+
+use crate::admin::service::messages;
+use crate::admin::store as admin_store;
+use crate::error::InvalidStateError;
+
+#[derive(Debug, PartialEq, Eq)]
+/// Representation of an `AdminServiceEvent` defined by the admin messages
+pub struct AdminServiceEvent {
+    pub event_id: i64,
+    pub event_type: EventType,
+    pub proposal: admin_store::CircuitProposal,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+/// Native representation of the `AdminServiceEvent` enum variants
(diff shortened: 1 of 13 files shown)


libsplinter/src/admin/store/proposed_circuit.rs
```diff
@@ -491,3 +493,39 @@ impl ProposedCircuitBuilder {
...
+        let mut circuit_builder = ProposedCircuitBuilder::new()
+            .with_circuit_id(&create_circuit.circuit_id)
+            .with_roster(&services)
+            .with_members(&nodes)
+            .with_authorization_type(&AuthorizationType::from(&create_circuit.authorization_type))
+            .with_persistence(&PersistenceType::from(&create_circuit.persistence))
+            .with_durability(&DurabilityType::from(&create_circuit.durability))
+            .with_routes(&RouteType::from(&create_circuit.routes))
+            .with_circuit_management_type(&create_circuit.circuit_management_type)
+            .with_application_metadata(&create_circuit.application_metadata);
```
agunde406, line 520: I think this should check to see if its not  empty before setting.

libsplinter/src/admin/service/event/mod.rs
```diff
@@ -12,5 +12,71 @@
...
+    pub event_id: i64,
+    pub event_type: EventType,
+    pub proposal: admin_store::CircuitProposal,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+/// Native representation of the `AdminServiceEvent` enum variants
+pub enum EventType {
+    ProposalSubmitted,
+    ProposalVote(Vec<u8>),
```
agunde406, line 41: Thoughts on changing this to something like to make it clear what the Vec<u8> is?
 ```
pub enum EventType {
    ProposalSubmitted,
    ProposalVote{
        requester: Vec<u8>,
    },
...

shannynalayna: I think that makes sense. I know there was a comment with the original `AdminServiceEvent` that the use of the included data isn't totally obvious. 👍

libsplinter/src/admin/store/proposed_circuit.rs
shannynalayna: Updated to check if the metadata is empty before adding it

agunde406 (approved):

peterschwarz (approved):

PR #1101, opened 2020-12-28
Add refresh token support for OAuth

ltseeley:


libsplinter/src/oauth/mod.rs
```diff
@@ -177,6 +178,20 @@ impl OAuthClient {
...
     }
 
+    /// Exchanges the given refresh token for an access token.
+    pub fn exchange_refresh_token(&self, refresh_token: String) -> Result<String, InternalError> {
+        self.client
+            .exchange_refresh_token(&RefreshToken::new(refresh_token))
+            .request(http_client)
+            .map(|response| response.access_token().secret().into())
+            .map_err(|err| {
+                InternalError::with_message(format!(
```
vaporos, line 196: This should probably be from_source_with_message(Box::new(err), "Failed to make refresh token exchange request".to_string)

libsplinter/src/rest_api/auth/identity/oauth.rs
```diff
@@ -110,10 +110,55 @@ impl IdentityProvider for OAuthUserIdentityProvider {
...
+                                    // handled here.
+                                    match self.oauth_client.get_subject(&access_token)? {
+                                        Some(_) => Ok(Some(user_id)),
+                                        None => Err(InternalError::with_message(
+                                            "failed to authenticate user with new access token"
+                                                .into(),
+                                        )),
+                                    }
+                                }
+                                Err(_) => {
```
vaporos, line 143: It doesn't matter what error this is? Should we be logging here?

libsplinter/src/rest_api/mod.rs
```diff
@@ -933,6 +933,28 @@ impl RestApiBuilder {
...
+                                oauth_openid_url,
+                                inflight_request_store,
+                            } => OpenIdOAuthClientBuilder::new()
+                                .with_discovery_url(oauth_openid_url)
+                                .with_client_id(client_id)
+                                .with_client_secret(client_secret)
+                                .with_redirect_url(redirect_url)
+                                .with_inflight_request_store(inflight_request_store)
+                                .build()
+                                .map_err(|err| {
```
vaporos, line 950: This is incorrect, as OAuthClientBuildError has multiple variants, not all of which should be mapped to InvalidStateError. (This might suggests OpenIdOAuthClientBuilder.build()'s signature is problematic, though RestApiServerError does have an InternalError variant, so the non-InvalidStateError cases could be mapped to that.) If OAuthClientBuildError::InternalStateError is the only one handled here, it may be better to extract the InternalStateError and insert it into the RestApiServerError without wrapping it, since the additional level of wrapping is probably not useful.

vaporos (changes requested):

libsplinter/src/oauth/mod.rs
ltseeley: Unfortunately the errors returned by the `oauth2` crate don't implement `std::error::Error` so they can't be used as a source with `InternalError`.

libsplinter/src/rest_api/auth/identity/oauth.rs
ltseeley: Changed this to log the error, but only with `debug` since this may not be a true error; it could just be that the token has been revoked. It's not easy to tell from the error provided by the oauth crate what the true cause of the failure is.

libsplinter/src/rest_api/mod.rs
ltseeley: Added a new commit to replace the `map_err`s from each branch with a `From<OAuthClientBuildError>` implementation for `RestApiServerError`.

libsplinter/src/oauth/builder/openid.rs
```diff
@@ -40,12 +44,26 @@ impl OpenIdOAuthClientBuilder {
         }
     }
 
+    /// Constructs a new [`OpenIdOAuthClientBuilder`] that's pre-configured with the scope for
+    /// getting refresh tokens.
+    pub fn new_azure() -> Self {
+        Self {
+            openid_discovery_url: None,
+            inner: OAuthClientBuilder::default().with_scopes(vec![AZURE_SCOPE.into()]),
```
peterschwarz, line 52: Does this lose the default scopes for azure?

ltseeley: No, `with_scopes` appends to the existing Vec

peterschwarz (approved):

vaporos (approved):

PR #1105, opened 2020-12-30
RoleBasedAuthorizationStore trait and supporting structs

peterschwarz:
This PR adds the `RoleBasedAuthorizationStore` and its supporting structs

### DIFF
--- a/libsplinter/src/rest_api/auth/roles/store/mod.rs
+++ b/libsplinter/src/rest_api/auth/roles/store/mod.rs
@@ -1,0 +1,379 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


libsplinter/src/rest_api/auth/roles/store/mod.rs
```diff
@@ -0,0 +1,379 @@
...
+        Ok(Role {
+            id: self.id,
+            display_name,
+            permissions: self.permissions,
+        })
+    }
+}
+
+/// An identity that may be assigned roles.
+pub enum Identity {
```
ltseeley, line 184: This is fine for now, but we may want to consider either removing this or renaming it so it doesn't conflict with the `Identity` struct introduced in https://github.com/Cargill/splinter/pull/1104

ltseeley (approved):

libsplinter/src/rest_api/auth/roles/store/mod.rs
```diff
@@ -0,0 +1,379 @@
...
+            permissions: self.permissions,
+        })
+    }
+}
+
+/// Updates an existing role.
+///
+/// This builder only allows the updatable fields to be modified.
+pub struct RoleUpdateBuilder {
+    id: String,
```
vaporos, line 133: Should these all be Option<> (its a builder).

peterschwarz: In the update builder, only certain fields can be modified. Id is not one of them.

vaporos (approved):

ltseeley (approved):

PR #1107, opened 2020-12-31
Add permissions for REST API authorization

ltseeley:
This PR adds permission strings to all Splinter REST API endpoints and updates the REST API's authorization guard to start using these permission strings for authorization checks.

This PR should not change any functionality for the end user, but it will enable restricting endpoints based on permissions in a future PR.

#### Testing

1. Start splinterd with Biome enabled

```bash
$ mkdir /tmp/splinter
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml -- database migrate
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml -- cert generate
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features experimental -- --enable-biome --tls-insecure -vv
```

2. In a new terminal window, register and login a user to verify the "allow all" endpoints work properly:

```bash
$ curl -X POST -H 'Content-Type: application/json' -d '{"username": "user", "hashed_password": "123"}' http://localhost:8080/biome/register
$ curl -X POST -H 'Content-Type: application/json' -d '{"username": "user", "hashed_password": "123"}' http://localhost:8080/biome/login
```

3. Verify that other endpoints still require an authorization header by attempting to access Biome keys without authorization:

```bash
$ curl http://localhost:8080/biome/keys
```

4. Use the token from your login in step 2 to verify you can get Biome keys with correct authorization:

```bash
$ curl -H 'Authorization: Bearer Biome:<token>' http://localhost:8080/biome/keys
```


libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -24,6 +24,11 @@ use crate::admin::service::AdminService;
 use crate::admin::store::AdminServiceStore;
 use crate::rest_api::{Resource, RestResourceProvider};
 
+#[cfg(all(feature = "authorization", feature = "rest-api-actix"))]
```
vaporos, line 27: We should consider putting these in an rest_api::permissions:: module so we can protect them with a single cfg statement -- primarily because actix 3 stuff is going to start to hit the repo soon and this will get messy cfg-wise if we do it per-permission.

libsplinter/src/biome/rest_api/actix/key_management.rs
```diff
@@ -35,31 +37,62 @@ pub fn make_key_management_route(
...
-            Method::Patch,
-            handle_patch(rest_config, key_store, secret_manager),
-        )
+        ));
+    #[cfg(feature = "authorization")]
+    {
+        resource
+            .add_method(
+                Method::Post,
+                NO_CHECK_PERMISSION.into(),
```
vaporos, line 50: I wonder if we should define

enum Permission {
    Allow(String),
    AllowAll,
}

That would make this look like:

.add_method(Method::Post, Permission::AllowAll, ...).

other places we would use

const CIRCUIT_READ_PERMISSION: Permission = Permission::Allow("circuit.read");

This would also type the permission arguments in functions.

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
```diff
@@ -31,12 +33,134 @@ use crate::rest_api::{
...
-        ))
-        .add_method(Method::Get, move |request, payload| {
+        ),
+    );
+
+    #[cfg(feature = "authorization")]
+    {
+        resource.add_method(
+            Method::Get,
+            CIRCUIT_READ_PERMISSION,
```
ltseeley, line 47: I'm actually curious if people think this should be a write permission

libsplinter/src/biome/rest_api/actix/key_management.rs
ltseeley: Updated permissions to be an enum similar to the one you suggested.

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
peterschwarz: It is just a read only stream

libsplinter/src/rest_api/auth/actix.rs
```diff
@@ -18,7 +18,7 @@ use actix_web::dev::*;
 use actix_web::{
     http::{
         header::{self, HeaderValue},
-        Method,
+        Method as ActixMethod,
```
peterschwarz, line 21: This is more accurately `HttpMethod`

vaporos (dismissed):

libsplinter/src/rest_api/auth/actix.rs
```diff
@@ -184,6 +184,16 @@ where
                     .into_future(),
                 )
             }
+            AuthorizationResult::UnknownEndpoint => {
```
peterschwarz, line 187: Does this mean it it is possible to add an endpoint that would be unknown to authorization?

libsplinter/src/rest_api/auth/actix.rs
ltseeley: The other `Method` enum imported in this file also represents HTTP methods. The key distinction is that one is provided by the Actix crate, the other is our own.

libsplinter/src/rest_api/auth/actix.rs
ltseeley: No; all endpoints require a permission field, so you can't add an endpoint that authorization wouldn't know about. If authorization doesn't know about an endpoint, the endpoint is unknown to the whole REST API.

peterschwarz: Does that mean this is unreachable?

ltseeley: No, this is what enables us to detect 404's. The middleware is run before Actix matches the request to a resource definition, so we have to distinguish between a known endpoint where the client is authorized and an unknown endpoint (which we can't determine how to authorize).

peterschwarz (approved):

vaporos (dismissed):

vaporos (approved):

peterschwarz (approved):

PR #1108, opened 2021-01-04
Clippy fixes for 1.49.0 Rust

agunde406:


shannynalayna (approved):

peterschwarz (approved):
Not sure I agree with the range change but I don’t necessarily want to start removing lints

vaporos (approved):

PR #1109, opened 2021-01-05
Break up crate::rest_api module and create actix_web_1 submodule

vaporos:
Moves the contents of the rest_api module into multiple modules and
re-exports within the rest_api module. This will make it easier to work
with as we start moving toward

The majority of the implementation is placed under a new actix_web_1
submodule. This is in anticipation of supporting Actix Web 3 under a
similar actix_web_3 module.

Although the rest_api module re-exports everything necessary, the code
within libsplinter is updated to use rest_api::actix_web_1 module
specifically. This makes it more explicit what code is tied to the
current implementation and will assist in deprecating portions of the
current rest_api module's API later.

This commit does not intentionally change any behavior.


peterschwarz (approved):
Should add some module documentation in a future PR

agunde406 (approved):

PR #1112, opened 2021-01-05
RoleBasedAuthorizationStore Diesel implementation

peterschwarz:
This PR implements the Diesel-backed `RoleBasedAuthorizationStore`. 

It only provides the SQLite implementation and tests.  ~PostgreSQL implementation will be included in a future PR.~ It also now includes the PostgreSQL implementation, but there are currently no tests for this.

### DIFF
--- a/libsplinter/src/rest_api/auth/roles/store/diesel/models.rs
+++ b/libsplinter/src/rest_api/auth/roles/store/diesel/models.rs
@@ -1,0 +1,106 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io::Write;
+
+use diesel::{
+    backend::Backend,
+    deserialize::{self, FromSql},
+    expression::{helper_types::AsExprOf, AsExpression},
+    serialize::{self, Output, ToSql},
+    sql_types::SmallInt,
+};
+
+use super::schema::{assignments, identities, role_permissions, roles};
+
+#[derive(Debug, PartialEq, Associations, Identifiable, Insertable, Queryable)]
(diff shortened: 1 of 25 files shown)


agunde406 (dismissed):

libsplinter/src/rest_api/auth/roles/store/diesel/models.rs
```diff
@@ -0,0 +1,106 @@
...
+    serialize::{self, Output, ToSql},
+    sql_types::SmallInt,
+};
+
+use super::schema::{assignments, identities, role_permissions, roles};
+
+#[derive(Debug, PartialEq, Associations, Identifiable, Insertable, Queryable)]
+#[table_name = "roles"]
+#[primary_key(id)]
+pub struct RoleModel {
```
vaporos, line 30: The pub / pub(super) here should probably be swapped around, and other things in this module could be pub(super).

agunde406 (dismissed):

libsplinter/src/rest_api/auth/roles/store/mod.rs
```diff
@@ -14,13 +14,19 @@
 
 //! This module defines the store trait for roles and their assignments to identities.
 
-#[cfg(feature = "sqlite")]
+#[cfg(any(
+    feature = "role-based-authorization-store-postgres",
+    feature = "sqlite"
+))]
```
vaporos, line 20: These should just be #[cfg(feature = "diesel")].

libsplinter/src/store/postgres.rs
```diff
@@ -95,7 +95,21 @@ impl StoreFactory for PgStoreFactory {
...
+        &self,
+    ) -> Box<dyn crate::rest_api::auth::roles::store::RoleBasedAuthorizationStore> {
+        Box::new(
+            crate::rest_api::auth::roles::store::DieselRoleBasedAuthorizationStore::new(
+                self.pool.clone(),
+            ),
+        )
+    }
+
+    #[cfg(all(
```
vaporos, line 109: Suggest removing this unimplemented!() implementation and using generic postgres feature instead of "role-based-authorization-store-postgres".

peterschwarz: This is part of a pattern we've been using as the postgres stores are not directly tested.  That way, the feature can be stabilized on SQLite before postgres stuff is tested in a running system.  The feature will be removed and replaced when stabilized

libsplinter/src/rest_api/auth/roles/store/mod.rs
peterschwarz: Fixed

vaporos (approved):

agunde406 (approved):

PR #1113, opened 2021-01-06
Add diesel implementation for `AdminServiceEventStore`

shannynalayna:
Adds the diesel implementation of the `AdminServiceEventStore` trait. Also includes a few unit tests for the operations in `libsplinter/src/admin/service/event/store/diesel/mod.rs`.

### DIFF
--- a/libsplinter/src/admin/service/event/store/diesel/mod.rs
+++ b/libsplinter/src/admin/service/event/store/diesel/mod.rs
@@ -32,6 +32,11 @@ use crate::admin::service::event::{
 };
 use crate::admin::service::messages;
 
+use operations::add_event::AdminServiceEventStoreAddEventOperation as _;
+use operations::list_events_by_management_type_since::AdminServiceEventStoreListEventsByManagementTypeSinceOperation as _;
+use operations::list_events_since::AdminServiceEventStoreListEventsSinceOperation as _;
+use operations::AdminServiceEventStoreOperations;
+
 /// A database-backed AdminServiceEventStore, powered by [`Diesel`](https://crates.io/crates/diesel).
 pub struct DieselAdminServiceEventStore<C: diesel::Connection + 'static> {
     connection_pool: Pool<ConnectionManager<C>>,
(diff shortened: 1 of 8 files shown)


libsplinter/src/admin/service/event/store/diesel/mod.rs
```diff
@@ -14,6 +14,12 @@
 
 //! Database backend support for the `AdminServiceEventStore`, powered by
 //! [`Diesel`](https://crates.io/crates/diesel).
+//!
```
agunde406, line 17: Several items listed in the commit message are actually in the last commit.

agunde406 (changes requested):

libsplinter/src/admin/service/event/store/diesel/mod.rs
shannynalayna: Updated to add the traits / impls mentioned in the commit message

agunde406 (approved):

libsplinter/src/admin/service/event/store/error.rs
```diff
@@ -45,3 +48,63 @@ impl fmt::Display for AdminServiceEventStoreError {
         }
     }
 }
+
+impl From<AdminServiceStoreError> for AdminServiceEventStoreError {
```
peterschwarz, line 52: What's the relationship between these errors?

shannynalayna: The `AdminServiceStore` has `TryFrom` implementations for some of the `CircuitProposal` data. Specifically, the enums that are represented as strings in the database have `TryFrom` implemented, so rather than re-implement here, those TryFroms that are defined in the AdminServiceStore are being used.

libsplinter/src/admin/service/event/store/diesel/operations/list_events_by_management_type_since.rs
```diff
@@ -0,0 +1,334 @@
...
+impl<'a, C> AdminServiceEventStoreListEventsByManagementTypeSinceOperation
+    for AdminServiceEventStoreOperations<'a, C>
+where
+    C: diesel::Connection,
+    C::Backend: HasSqlType<diesel::sql_types::BigInt>,
+    String: diesel::deserialize::FromSql<diesel::sql_types::Text, C::Backend>,
+    i64: diesel::deserialize::FromSql<diesel::sql_types::BigInt, C::Backend>,
+    Vec<u8>: diesel::deserialize::FromSql<diesel::sql_types::Binary, C::Backend>,
+{
+    fn list_events_by_management_type_since(
```
peterschwarz, line 67: Is this `list_by` operation only different from the other one due to a difference in the initial filter?

shannynalayna: Yeah, similar to how the `AdminServiceStore` lists things based on the circuit predicates, but the rest of the operation is relatively the same. It takes the event IDs from that filtered query, and uses those to filter the later queries.

peterschwarz: can these be consolidated into a single operation?

shannynalayna: Probably, one that only the event IDs are passed to. Assuming you mean within the separate `list_since` and `list_by_management_type_since` operations, and not altering the trait's operations (which could also probably be consolidated by just making the management type optional).

peterschwarz: Right, I'm only saying at the operation level.  The trait should remain as it is.

shannynalayna: Updated to include a `list_events` operation that just takes a list of event IDs and holds all of the code that translates the database objects to the native structs, used by both the `list_*` operations.

peterschwarz (approved):

agunde406 (approved):

PR #1118, opened 2021-01-12
Add `AdminServiceEventStore` to `StoreFactory`

shannynalayna:
Adds the `AdminServiceEventStore` to the store factory, as well as integrate the feature into splinterd to allow the daemon to use the event store.

### DIFF
--- a/libsplinter/src/store/postgres.rs
+++ b/libsplinter/src/store/postgres.rs
@@ -120,4 +120,25 @@ impl StoreFactory for PgStoreFactory {
     fn get_biome_user_profile_store(&self) -> Box<dyn crate::biome::UserProfileStore> {
         Box::new(crate::biome::DieselUserProfileStore::new(self.pool.clone()))
     }
+
+    #[cfg(feature = "admin-service-event-store-diesel")]
+    fn get_admin_service_event_store(
+        &self,
+    ) -> Box<dyn crate::admin::service::event::store::AdminServiceEventStore> {
+        Box::new(
+            crate::admin::service::event::store::diesel::DieselAdminServiceEventStore::new(
+                self.pool.clone(),
+            ),
+        )
+    }
+
+    #[cfg(all(
+        feature = "admin-service-event-store",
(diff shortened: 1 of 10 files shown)


splinterd/Cargo.toml
```diff
@@ -96,6 +96,10 @@ admin-service-event-store = [
   "splinter/admin-service-event-store",
   "splinter/admin-service"
 ]
+admin-service-event-store-diesel = [
+  "admin-service-event-store",
+  "splinter/admin-service-event-store-diesel"
```
peterschwarz, line 101: I don’t think this needs to be a separate feature in splinterd, the diesel feature should just be added to the “admin-service-event-store” feature (since splinterd requires a DB now)

libsplinter/src/store/postgres.rs
```diff
@@ -120,4 +120,15 @@ impl StoreFactory for PgStoreFactory {
     fn get_biome_user_profile_store(&self) -> Box<dyn crate::biome::UserProfileStore> {
         Box::new(crate::biome::DieselUserProfileStore::new(self.pool.clone()))
     }
+
+    #[cfg(feature = "admin-service-event-store")]
```
peterschwarz, line 124: This needs to have a method for when the “admin-service-event-store-diesel” is enabled an one for when it isn’t, same with the SQLite version.

peterschwarz (changes requested):

splinterd/Cargo.toml
shannynalayna: I changed the `admin-service-event-store` feature here to pull in `splinter/admin-service-event-store-diesel`

libsplinter/src/store/postgres.rs
shannynalayna: Updated with `unimplemented!()` in these cases

peterschwarz (approved):

agunde406 (dismissed):

vaporos (approved):

agunde406 (approved):

PR #1121, opened 2021-01-12
Add profile provider and biome/profile REST API endpoints

isabeltomb:
Create a ProfileProvider trait and implement it for github and openid. The trait has a method 'get_profile' which takes an access token as an argument and queries the user info endpoint to retrieve the profile information given by the oauth provider.

Add ProfileProvider to the OAuthClient struct. Add a 'with_profile_provider' method to the OAuthClientBuilder as well as the github and openid OAuthClientBuilders.

Add a 'profile' field to the UserInfo struct. Modify the 'exchange_authorization_code' method to retrieve the profile information
using the profile provider and set the profile field of the UserInfo struct when the biome-profile feature is enabled.

Modify the 'make_callback_route' method to take a UserProfileStore as an argument and save the profile information stored in the profile field of the UserInfo struct if the biome-profile feature is enabled. If the given profile already exists in the profile store it will be updated otherwise it will be added as a new profile.

Add a user_profile_store field to AuthConfig::OAuth and OAuthResourceProvider these fields are only available when the 'biome-profile' feature is enabled.

Impl From oauth::Profile for biome::profile::store::Profile to allow for conversion between the two structs.

Add the biome/profile/{id} and biome/profiles endpoints for retrieving and listing biome user profiles.

## **Testing:**
1. Start splinterd with an oauth provider
```
$ cargo run --manifest-path cli/Cargo.toml -- database migrate -C postgres://postgres:mypassword@localhost:5432/splinter
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml -- cert generate
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features experimental -- \
  --database postgres://postgres:mypassword@localhost:5432/splinter \
  --oauth-provider <provider> \
  --oauth-client-id <client-id> \
  --oauth-client-secret <client-secret> \
  --rest-api-endpoint http://localhost:8080
  --tls-insecure
  -vv
```

2. In a browser go to `localhost:8080/oauth/login?redirect_url=redirect`

3. Go back to the terminal window where splinterd is running and check that it shows `INFO [splinter::oauth::rest_api::actix::callback] User profile saved`
4. Copy the access token from splinterd, it should be in a log message: `GET /oauth/redirect?access_token=<access-token>` and Make the following curl calls

`curl -i -H'Authorization:Bearer <access-token>' http://localhost:8080/biome/profile`
should return the profile information for the profile you used to log in

copy the user_id from the _Authenticated user_ debug message from splinterd 
`curl -i -H'Authorization:Bearer <access-token>' http://localhost:8080/biome/profiles/<user_id>`
should get a `HTTP/1.1 401 Unauthorized` indicating the endpoint exists

`curl -i -H'Authorization:Bearer <access-token>' http://localhost:8080/biome/profiles`
should get a `HTTP/1.1 401 Unauthorized` indicating the endpoint exists

### DIFF
--- a/libsplinter/src/biome/rest_api/actix/profile.rs
+++ b/libsplinter/src/biome/rest_api/actix/profile.rs
@@ -1,0 +1,72 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use super::authorize::get_authorized_user;
+use crate::actix_web::HttpResponse;
+#[cfg(feature = "biome-profile")]
+use crate::biome::profile::store::UserProfileStore;
+use crate::futures::IntoFuture;
+use crate::protocol;
+#[cfg(feature = "authorization")]
+use crate::rest_api::auth::Permission;
+use crate::rest_api::{
+    actix_web_1::{HandlerFunction, Method, ProtocolVersionRangeGuard, Resource},
+    ErrorResponse,
+};
+
+pub fn make_profile_route(profile_store: Arc<dyn UserProfileStore>) -> Resource {
+    let resource =
+        Resource::build("/biome/profile").add_request_guard(ProtocolVersionRangeGuard::new(
+            protocol::BIOME_FETCH_PROFILE_PROTOCOL_MIN,
+            protocol::BIOME_PROTOCOL_VERSION,
+        ));
+    #[cfg(feature = "authorization")]
+    {
+        resource.add_method(
+            Method::Get,
+            Permission::AllowAuthenticated,
+            handle_get(profile_store),
+        )
+    }
+    #[cfg(not(feature = "authorization"))]
+    {
+        resource.add_method(Method::Get, handle_get(profile_store))
+    }
+}
+
+/// Defines a REST endpoint for retrieving the profile of the authenticated user
+fn handle_get(profile_store: Arc<dyn UserProfileStore>) -> HandlerFunction {
+    Box::new(move |request, _| {
+        let profile_store = profile_store.clone();
+
+        let user = match get_authorized_user(&request) {
+            Ok(user) => user,
+            Err(response) => return response,
+        };
+
+        match profile_store.get_profile(&user) {
+            Ok(profile) => Box::new(HttpResponse::Ok().json(profile).into_future()),
+            Err(err) => {
+                debug!("Failed to fetch profile {}", err);
+                Box::new(
+                    HttpResponse::InternalServerError()
+                        .json(ErrorResponse::internal_error())
+                        .into_future(),
+                )
+            }
+        }
+    })
+}
--- a/libsplinter/src/oauth/profile/github.rs
+++ b/libsplinter/src/oauth/profile/github.rs
@@ -1,0 +1,83 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A profile provider that looks up GitHub profile information
+
+use reqwest::{blocking::Client, StatusCode};
+use serde::Deserialize;
+
+use crate::error::InternalError;
+use crate::oauth::Profile;
+
+use super::ProfileProvider;
+
+/// Retrieves a GitHub profile information from the GitHub servers
+#[derive(Clone)]
+pub struct GithubProfileProvider;
+
+impl ProfileProvider for GithubProfileProvider {
+    fn get_profile(&self, access_token: &str) -> Result<Option<Profile>, InternalError> {
+        let response = Client::builder()
+            .build()
+            .map_err(|err| InternalError::from_source(err.into()))?
+            .get("https://api.github.com/user")
+            .header("Authorization", format!("Bearer {}", access_token))
+            .header("User-Agent", "splinter")
+            .send()
+            .map_err(|err| InternalError::from_source(err.into()))?;
+
+        if !response.status().is_success() {
+            match response.status() {
+                StatusCode::UNAUTHORIZED => return Ok(None),
+                status_code => {
+                    return Err(InternalError::with_message(format!(
+                        "Received unexpected response code: {}",
+                        status_code
+                    )))
+                }
+            }
+        }
+
+        let user_profile = response
+            .json::<GithubProfileResponse>()
+            .map_err(|_| InternalError::with_message("Received unexpected response body".into()))?;
+
+        Ok(Some(Profile::from(user_profile)))
+    }
+
+    fn clone_box(&self) -> Box<dyn ProfileProvider> {
+        Box::new(self.clone())
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct GithubProfileResponse {
+    pub login: String,
+    pub name: Option<String>,
+    pub email: Option<String>,
+    pub avatar_url: Option<String>,
+}
+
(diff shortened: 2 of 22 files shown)


vaporos (dismissed):

libsplinter/src/oauth/profile/github.rs
```diff
@@ -0,0 +1,84 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A profile provider that looks up GitHub profile information
+
+use reqwest::{blocking::Client, StatusCode};
+use serde::Deserialize;
+
+use crate::error::InternalError;
+
```
ltseeley, line 21: Don't need a new line here; `use crate::` imports should be grouped together. This applies to the other files in this commit as well.

libsplinter/src/oauth/rest_api/actix/callback.rs
```diff
@@ -83,15 +90,57 @@ pub fn make_callback_route(
...
-                                                Ok(_) => HttpResponse::Found()
+                                                Ok(_) => {
+                                                    // Save the user profile data if the biome-profile feature is enabled
+                                                    #[cfg(feature = "biome-profile")]
+                                                    {
+                                                        let profile = Profile::from(&user_info.profile().clone());
+                                                        match user_profile_store.get_profile(profile.user_id()) {
+                                                            Ok(profile) => { match user_profile_store.update_profile(profile) {
+                                                                    Ok(_) =>  {
+                                                                        info!("User profile updated");
```
ltseeley, line 101: This should probably be `debug` instead of `info`, since it generally won't be important information for admins to see every time a user's profile is updated. Same for the "User profile saved" message below.

libsplinter/src/oauth/rest_api/actix/callback.rs
```diff
@@ -83,15 +90,57 @@ pub fn make_callback_route(
                                     {
                                         Ok(session) => {
                                             match oauth_user_session_store.add_session(session) {
-                                                Ok(_) => HttpResponse::Found()
+                                                Ok(_) => {
```
ltseeley, line 93: We probably shouldn't nest this in the `add_session` match. `cargo fmt` can't handle this because the nesting is too deep for it to properly format these lines.

libsplinter/src/biome/rest_api/actix/profile.rs
```diff
@@ -0,0 +1,121 @@
...
+};
+
+#[cfg(feature = "biome-profile")]
+use crate::biome::profile::store::{UserProfileStore, UserProfileStoreError};
+
+#[cfg(feature = "authorization")]
+use crate::biome::rest_api::BIOME_USER_READ_PERMISSION;
+
+/// Defines a REST endpoint to list profiles from the database
+pub fn make_profile_list_route(profile_store: Arc<dyn UserProfileStore>) -> Resource {
```
ltseeley, line 31: We've recently started breaking up our REST API endpoints into different modules based on the endpoint, so these two `make_profile_*` methods should be in separate files (see the admin or registry REST APIs for reference).

libsplinter/src/protocol/mod.rs
```diff
@@ -81,3 +81,6 @@ pub(crate) const BIOME_VERIFY_PROTOCOL_MIN: u32 = 1;
 
 #[cfg(all(feature = "biome-key-management", feature = "rest-api",))]
 pub(crate) const BIOME_KEYS_PROTOCOL_MIN: u32 = 1;
+
+#[cfg(all(feature = "biome-profile", feature = "rest-api",))]
+pub(crate) const BIOME_PROFILE_PROTOCOL_MIN: u32 = 1;
```
ltseeley, line 86: Each endpoint should have its own protocol number, since they may be versioned differently.

libsplinter/src/biome/rest_api/actix/profile.rs
```diff
@@ -0,0 +1,121 @@
...
+                        .into_future()
+                }
+            })
+        })
+    }
+}
+
+pub fn make_profile_routes(profile_store: Arc<dyn UserProfileStore>) -> Resource {
+    let resource =
+        Resource::build("/biome/profile/{id}").add_request_guard(ProtocolVersionRangeGuard::new(
```
ltseeley, line 71: Looks like this path is incorrect (shouldn't have `/{id}`). Two other thoughts on these endpoints:
* I think they should be `/biome/profiles` and `/biome/profiles/{id}`; "profiles" plural because you're listing all profiles or getting one of the profiles in the system (this is a common naming convention for REST APIs).
* We probably need a `/biome/profile` endpoint that uses the authenticated user's identity to get their own profile information. I assume this is how UI will be getting information. Most users won't have the `BIOME_USER_READ_PERMISSION`, but should still be able to access their own profile info. This is how the biome key management endpoints work, for reference.

ltseeley (changes requested):

libsplinter/src/oauth/profile/github.rs
isabeltomb: Fixed

libsplinter/src/oauth/rest_api/actix/callback.rs
isabeltomb: Fixed

libsplinter/src/oauth/rest_api/actix/callback.rs
isabeltomb: Fixed

libsplinter/src/protocol/mod.rs
isabeltomb: Fixed

libsplinter/src/biome/rest_api/actix/profile.rs
isabeltomb: Changed the existing endpoints to `/biome/profiles` and `/biome/profiles/{id}` and added `biome/profile` for retrieving the authenticated user's profile

peterschwarz (dismissed):

libsplinter/src/oauth/rest_api/actix/callback.rs
```diff
@@ -98,6 +106,27 @@ pub fn make_callback_route(
                                             HttpResponse::InternalServerError()
                                                 .json(ErrorResponse::internal_error())
                                         }
```
ltseeley, line 108: The branches here are all being discarded because they aren't assigned to a variable or returned. The `Ok(_)` branch can just be `=> {}` and the `Err` branches should `return` their `HttpResponse`s.

libsplinter/src/oauth/rest_api/actix/callback.rs
```diff
@@ -98,6 +106,27 @@ pub fn make_callback_route(
...
+                                                debug!("User profile saved");
+                                                HttpResponse::Found()
+                                                    .header(LOCATION, redirect_url)
+                                                    .finish()
+                                            }
+                                            Err(err) => {
+                                                error!("Failed to save profile: {}", err);
+                                                HttpResponse::InternalServerError()
+                                                    .json(ErrorResponse::internal_error())
+                                            }
```
ltseeley, line 128: The `HttpResponse::Found` response should be outside of this match and the `HttpResponse::InternalServerError` should be explicitly `return`ed. If the `biome-profile` feature isn't enabled, this code won't compile because there won't be anything returning an `HttpResponse`.

libsplinter/src/oauth/rest_api/actix/callback.rs
```diff
@@ -214,6 +264,46 @@ fn new_splinter_access_token() -> String {
...
+            .with_email(profile.email.clone())
+            .with_picture(profile.picture.clone())
+            .build()
+            .map_err(|err| InternalError::from_source(Box::new(err)))?;
+
+        match user_profile_store.get_profile(&user.user_id()) {
+            Ok(_) => user_profile_store
+                .update_profile(profile)
+                .map_err(|err| InternalError::from_source(Box::new(err))),
+            Err(UserProfileStoreError::InvalidArgument(_)) => user_profile_store
```
ltseeley, line 295: This can be done in a follow-up PR, but `get_profile` should return `Result<Option<Profile>, UserProfileStoreError>` and `Ok(None)` when the profile doesn't exist. This is generally how we do it because it's more intuitive and idiomatic.

libsplinter/src/biome/rest_api/actix/profile.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let user = match get_authorized_user(&request) {
+            Ok(user) => user,
+            Err(response) => return response,
+        };
+
+        match profile_store.get_profile(&user) {
+            Ok(profile) => Box::new(HttpResponse::Ok().json(profile).into_future()),
+            Err(err) => {
+                debug!("Failed to fetch keys {}", err);
```
ltseeley, line 63: s/keys/profile/

isabeltomb: Fixed

libsplinter/src/oauth/rest_api/actix/callback.rs
isabeltomb: Fixed

libsplinter/src/oauth/rest_api/actix/callback.rs
isabeltomb: Fixed

ltseeley (approved):

peterschwarz (approved):

PR #1125, opened 2021-01-14
Role-based Authorization Handler

peterschwarz:
This PR implements the Role-based Authorization handler.  It includes an update to the `RoleBaseAuthorizationStore` to provide a function to return the roles assigned to a given identity.

### DIFF
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -81,6 +81,7 @@ experimental = [
     "stable",
     # The following features are experimental:
     "authorization",
+    "authorization-handler-rbac",
     "admin-service-event-store",
     "biome-oauth",
     "health",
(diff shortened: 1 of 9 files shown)


ltseeley (approved):

ltseeley (dismissed):

splinterd/Cargo.toml
```diff
@@ -81,6 +81,7 @@ experimental = [
     # The following features are experimental:
     "auth",
     "authorization",
+    "authorization-handler-rbac",
```
ltseeley, line 84: Can we shorten this to `auth-handler-rbac` for brevity? The maintenance and allow_keys ones would especially benefit from this (they'll match). Can you also guard the handler itself in libsplinter?

peterschwarz: I think the name is correct, but I have added the guard to libsplinter as well.

vaporos (approved):

ltseeley (approved):

vaporos (approved):

PR #1127, opened 2021-01-14
Add `circuit_status` to circuits

shannynalayna:
Adds a `circuit_status` flag to circuits, from the proto message up.

Also renames the `destroy` request in the protos to `disband`, including the affected structs, mainly the `ProposalType` in a `CircuitProposal` which represented the `destroy` option, now renamed to `disband`.

### DIFF
--- a/libsplinter/src/admin/service/event/store/diesel/models.rs
+++ b/libsplinter/src/admin/service/event/store/diesel/models.rs
@@ -15,6 +15,15 @@
 //! Database representations used to implement a diesel backend for the `AdminServiceEventStore`.
 
 use std::convert::TryFrom;
+use std::io::Write;
+
+use diesel::{
+    backend::Backend,
+    deserialize::{self, FromSql},
+    expression::{helper_types::AsExprOf, AsExpression},
+    serialize::{self, Output, ToSql},
+    sql_types::SmallInt,
+};
 
 use crate::admin::service::event::{
     store::{
@@ -92,6 +101,7 @@ pub struct AdminEventProposedCircuitModel {
     pub comments: Option<String>,
     pub display_name: Option<String>,
     pub circuit_version: i32,
+    pub circuit_status: CircuitStatusModel,
 }
 
 impl From<(i64, &CreateCircuit)> for AdminEventProposedCircuitModel {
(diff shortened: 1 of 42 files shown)


libsplinter/src/admin/service/messages/mod.rs
```diff
@@ -85,6 +86,17 @@ impl CreateCircuit {
             }
         };
 
+        let circuit_status = match proto.get_circuit_status() {
+            admin::Circuit_CircuitStatus::ACTIVE => CircuitStatus::Active,
+            admin::Circuit_CircuitStatus::DISBANDED => CircuitStatus::Disbanded,
+            admin::Circuit_CircuitStatus::ABANDONED => CircuitStatus::Abandoned,
+            admin::Circuit_CircuitStatus::UNSET_CIRCUIT_STATUS => {
```
agunde406, line 93: I think this needs to default to Active for 0.4 circuit compatibility. Since those circuits will not have this set.

agunde406 (changes requested):
circuit_status should also be added to the YAML implementation

libsplinter/src/admin/service/event/store/diesel/models.rs
```diff
@@ -92,6 +92,7 @@ pub struct AdminEventProposedCircuitModel {
     pub comments: Option<String>,
     pub display_name: Option<String>,
     pub circuit_version: i32,
+    pub circuit_status: String,
```
peterschwarz, line 95: This should be an enum (see https://github.com/Cargill/splinter/blob/master/libsplinter/src/rest_api/auth/rbac/store/diesel/models.rs#L46-L89 for an example)

libsplinter/src/admin/service/messages/mod.rs
shannynalayna: I changed this to default to active if not set

libsplinter/src/admin/service/event/store/diesel/models.rs
shannynalayna: Updated to be represented as an enum in the admin store, and event store

peterschwarz (approved):

peterschwarz (approved):

agunde406 (dismissed):

agunde406 (approved):

peterschwarz (approved):

PR #1134, opened 2021-01-18
Add actix 1 paging extraction method

peterschwarz:
This change introduces a function that will extract paging information from a query and convert it into a PagingQuery struct that contains the limit and offset. Any errors are returned in a single BadPagingRequest error that can be converted into an HttpResponse future containing a BadRequest.

This change also includes updating existing routes that make use of these functions.

### DIFF
--- a/libsplinter/src/rest_api/actix_web_1/paging.rs
+++ b/libsplinter/src/rest_api/actix_web_1/paging.rs
@@ -1,0 +1,83 @@
+// Copyright (c) 2019 Target Brands, Inc.
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


libsplinter/src/rest_api/actix_web_1/paging.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::actix_web::{web, Error, HttpResponse};
+use crate::futures::{future::IntoFuture, Future};
+use crate::rest_api::{
+    paging::{PagingQuery, DEFAULT_LIMIT, DEFAULT_OFFSET},
+    ErrorResponse,
+};
+
+/// Return the PagingQuery from a Query object of key-value pairs.
+pub fn get_paging_query(
```
ltseeley, line 26: Instead of implementing the parsing ourselves, maybe we can just do something like this: https://docs.rs/actix-web/1.0.9/actix_web/web/struct.Query.html#example

peterschwarz: We could do that as a future refactor or implementation - particularly in the actix3 work.

peterschwarz: Mainly, the reason to keep it as a map is due to the fact that most other areas were using it that way.

peterschwarz:
Closing this to be handled in future actix 3 work.

PR #1135, opened 2021-01-18
Update methods using the cylinder key-load feature

isabeltomb:
Update the `create_cylinder_jwt_auth` method to use the updated cylinder key load API to retrieve the private key used in jwt signing.

Update the default location that `keygen` saves keys to be ~/.cylinder/keys

Modify gameroom daemon to use the new `load_key` method from the cylinder key::load module.

Modify the scabbard `create_cylinder_jwt_auth` method to use the new methods in the cylinder key::load module.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -30,7 +30,7 @@ path = "src/main.rs"
 
 [dependencies]
 clap = "2"
-cylinder = "0.2"
+cylinder = "0.2.2"
 diesel = { version = "1.0", features = ["postgres"], optional = true }
 dirs = "2.0"
 flexi_logger = "0.14"
--- a/cli/src/action/mod.rs
+++ b/cli/src/action/mod.rs
@@ -25,6 +25,8 @@ pub mod keygen;
 pub mod registry;
 
 use std::collections::HashMap;
+#[cfg(feature = "splinter-cli-jwt")]
+use std::env;
 use std::ffi::CString;
 use std::fs::File;
 use std::io::{Error as IoError, ErrorKind, Read};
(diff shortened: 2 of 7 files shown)


cli/Cargo.toml
```diff
@@ -30,7 +30,7 @@ path = "src/main.rs"
 
 [dependencies]
 clap = "2"
-cylinder = "0.2"
+cylinder = "^0.2.2"
```
vaporos, line 33: Specifying the ^ here is unnecessary as that is implied in "0.2.2".

isabeltomb: Fixed

ltseeley (approved):

cli/src/action/mod.rs
```diff
@@ -134,26 +137,21 @@ fn msg_from_io_error(err: IoError) -> String {
...
         } else {
-            return Err(CliError::ActionError(
-                "Path is not valid unicode".to_string(),
-            ));
+            load_key(key_name, &current_user_search_path())
+                .map_err(|err| CliError::ActionError(format!("Unable to retrieve key: {}", err)))?
+                .ok_or_else(|| CliError::ActionError("Key not found".to_string()))?
         }
+    } else {
+        load_key(&current_user_key_name(), &current_user_search_path())
```
vaporos, line 150: This needs to add ~/.splinter/keys to the search path to be consistent with keygen

vaporos (changes requested):

cli/src/action/mod.rs
isabeltomb: Fixed

ltseeley (approved):

vaporos (approved):

PR #1137, opened 2021-01-19
Remove feature guards from Identity

ltseeley:
Removes the feature guards for the various `Identity` variants so
they're always available.

### DIFF
--- a/libsplinter/src/rest_api/auth/identity/mod.rs
+++ b/libsplinter/src/rest_api/auth/identity/mod.rs
@@ -21,9 +21,6 @@ pub mod cylinder;
 #[cfg(feature = "oauth")]
 pub mod oauth;
 
-#[cfg(feature = "cylinder-jwt")]
-use ::cylinder::PublicKey;
-
 use crate::error::InternalError;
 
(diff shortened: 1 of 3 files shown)


peterschwarz (approved):

libsplinter/src/rest_api/auth/identity/mod.rs
```diff
@@ -33,10 +33,8 @@ use super::AuthorizationHeader;
 pub enum Identity {
     /// A custom identity
     Custom(String),
-    #[cfg(feature = "cylinder-jwt")]
     /// A Cylinder public key
     Key(PublicKey),
```
vaporos, line 37: This will need to be a String to prevent the cylinder dependency.

vaporos (changes requested):

libsplinter/src/rest_api/auth/identity/mod.rs
ltseeley: Updated

peterschwarz (approved):

vaporos (approved):

PR #1146, opened 2021-01-22
Implement `CircuitDisbandRequest` in admin service

shannynalayna:
Adds a few commits necessary to implement the `CircuitDisbandRequest` in the admin service. 

This feature will be fully testable once the CLI commands are implemented (wip), but otherwise run `cargo test --features experimental disband` to specify the tests added for this implementation.

### DIFF
--- a/libsplinter/src/admin/service/error.rs
+++ b/libsplinter/src/admin/service/error.rs
@@ -17,7 +17,9 @@ use std::fmt;
 
 use crate::admin::store::error::AdminServiceStoreError;
 use crate::consensus::error::ProposalManagerError;
-use crate::orchestrator::{InitializeServiceError, ShutdownServiceError};
+use crate::orchestrator::InitializeServiceError;
+#[cfg(feature = "circuit-disband")]
+use crate::orchestrator::ShutdownServiceError;
 use crate::service::error::{ServiceError, ServiceSendError};
 
 use protobuf::error;
@@ -161,6 +163,7 @@ pub enum AdminSharedError {
         context: String,
         source: Option<InitializeServiceError>,
     },
+    #[cfg(feature = "circuit-disband")]
     ServiceShutdownFailed {
         context: String,
         source: Option<ShutdownServiceError>,
@@ -190,6 +193,7 @@ impl Error for AdminSharedError {
                     None
                 }
             }
+            #[cfg(feature = "circuit-disband")]
             AdminSharedError::ServiceShutdownFailed { source, .. } => {
                 if let Some(ref err) = source {
                     Some(err)
@@ -224,6 +228,7 @@ impl fmt::Display for AdminSharedError {
                     f.write_str(&context)
                 }
             }
+            #[cfg(feature = "circuit-disband")]
             AdminSharedError::ServiceShutdownFailed { context, source } => {
                 if let Some(ref err) = source {
                     write!(f, "{}: {}", context, err)
--- a/libsplinter/src/admin/service/error.rs
+++ b/libsplinter/src/admin/service/error.rs
@@ -17,7 +17,7 @@ use std::fmt;
 
 use crate::admin::store::error::AdminServiceStoreError;
 use crate::consensus::error::ProposalManagerError;
-use crate::orchestrator::InitializeServiceError;
+use crate::orchestrator::{InitializeServiceError, ShutdownServiceError};
 use crate::service::error::{ServiceError, ServiceSendError};
 
 use protobuf::error;
@@ -161,6 +161,10 @@ pub enum AdminSharedError {
         context: String,
         source: Option<InitializeServiceError>,
     },
+    ServiceShutdownFailed {
+        context: String,
+        source: Option<ShutdownServiceError>,
+    },
     ServiceSendError(ServiceSendError),
     UnknownAction(String),
     ValidationFailed(String),
@@ -180,6 +184,13 @@ impl Error for AdminSharedError {
             AdminSharedError::InvalidMessageFormat(err) => Some(err),
             AdminSharedError::NoPendingChanges => None,
             AdminSharedError::ServiceInitializationFailed { source, .. } => {
+                if let Some(ref err) = source {
+                    Some(err)
+                } else {
+                    None
+                }
+            }
+            AdminSharedError::ServiceShutdownFailed { source, .. } => {
                 if let Some(ref err) = source {
                     Some(err)
                 } else {
@@ -207,6 +218,13 @@ impl fmt::Display for AdminSharedError {
                 write!(f, "tried to commit without pending changes")
             }
             AdminSharedError::ServiceInitializationFailed { context, source } => {
+                if let Some(ref err) = source {
+                    write!(f, "{}: {}", context, err)
+                } else {
+                    f.write_str(&context)
+                }
+            }
+            AdminSharedError::ServiceShutdownFailed { context, source } => {
                 if let Some(ref err) = source {
                     write!(f, "{}: {}", context, err)
                 } else {
--- a/libsplinter/src/admin/service/messages/mod.rs
+++ b/libsplinter/src/admin/service/messages/mod.rs
@@ -612,6 +612,7 @@ pub enum AdminServiceEvent {
     ProposalAccepted((CircuitProposal, Vec<u8>)),
     ProposalRejected((CircuitProposal, Vec<u8>)),
     CircuitReady(CircuitProposal),
+    CircuitDisbanded(CircuitProposal),
 }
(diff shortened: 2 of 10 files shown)


libsplinter/src/admin/service/mod.rs
```diff
@@ -1089,6 +1090,214 @@ mod tests {
         mesh.shutdown_signaler().shutdown();
     }
 
+    /// Test that a circuit disband sends the appropriate messages.
```
agunde406, line 1093: Can you expand this description so it explains the test more.

libsplinter/src/admin/service/shared.rs
```diff
@@ -390,89 +391,143 @@ impl AdminServiceShared {
...
+                                        "Unable to remove circuit from routing table: {}",
+                                        circuit_id
+                                    ))
+                                })?;
+                            // Removing the circuit's peer refs
+                            for member in circuit_proposal.get_circuit_proposal().get_members() {
+                                self.remove_peer_ref(member.get_node_id());
+                            }
+                            // Removing the corresponding `active` circuit from the admin store
+                            self.admin_store.remove_circuit(circuit_id).map_err(|_| {
```
agunde406, line 498: We should use update_circuit here instead of remove/upgrade to keep the operation atomic. This will require you get the circuit from the proposal here instead of relying on the admin service store to do the conversion.

libsplinter/src/admin/service/shared.rs
```diff
@@ -506,6 +561,18 @@ impl AdminServiceShared {
...
+                            CircuitManagementPayload_Action::CIRCUIT_DISBAND_REQUEST => {
+                                // notify registered application authorization handlers of the
+                                // committed disband circuit proposal
+                                let event = messages::AdminServiceEvent::ProposalSubmitted(
+                                    messages::CircuitProposal::from_proto(circuit_proposal.clone())
+                                        .map_err(AdminSharedError::InvalidMessageFormat)?,
+                                );
+                                self.send_event(&mgmt_type, event);
+
+                                info!("committed changes to disband circuit {}", circuit_id);
```
agunde406, line 573: This should say `info!("committed changes for new circuit proposal to disband circuit {}", circuit_id);` and the create log above should say to `proposal to create circuit {}`

libsplinter/src/admin/service/shared.rs
```diff
@@ -913,6 +1062,69 @@ impl AdminServiceShared {
...
+        message_sender: String,
+    ) -> Result<(), ServiceError> {
+        let mut missing_protocol_ids = vec![];
+        let mut pending_peers = vec![];
+        let mut pending_members = vec![];
+        let mut added_peers: Vec<String> = vec![];
+        for node in members {
+            if self.node_id() != node.get_node_id() {
+                debug!("Referencing node {:?}", node);
+                let peer_ref = self
```
agunde406, line 1078: This should not be adding peer_refs, required peer_refs will have been added when the circuit was created. This is adding an extra set of peer_refs that will not be removed when the disband request is approved. Only create_payloads should add new peer_refs

libsplinter/src/admin/service/shared.rs
```diff
@@ -1500,12 +1795,6 @@ impl AdminServiceShared {
                     return Err(AdminSharedError::ValidationFailed(
                         "Proposed circuit cannot have a display name on protocol 1".to_string(),
                     ));
-                } else if circuit.get_circuit_status()
```
agunde406, line 1503: why was this deleted?

libsplinter/src/admin/service/shared.rs
```diff
@@ -1824,6 +2113,81 @@ impl AdminServiceShared {
...
+                AdminSharedError::ValidationFailed(format!(
+                    "{} is not permitted to vote for node {}",
+                    to_hex(signer_public_key),
+                    requester_node_id
+                ))
+            })?;
+
+        if self.has_proposal(circuit.get_circuit_id())? {
+            return Err(AdminSharedError::ValidationFailed(format!(
+                "Ignoring duplicate disband proposal of circuit {}",
```
agunde406, line 2153: This should be updated to say "Ignoring duplicate proposal for circuit" (and in the create validation) since it might be a different proposal type

agunde406 (changes requested):

libsplinter/src/admin/service/mod.rs
agunde406: same for other tests added

shannynalayna: Yep!

libsplinter/src/admin/service/shared.rs
shannynalayna: This is checking the protocol version, rather than the `circuit_version` (This was an error I was initially running into while testing the disband request). I added a check to verify the circuit version in the `validate_disband_request` rather than the admin service protocol, as it is doing here.

libsplinter/src/admin/service/shared.rs
```diff
@@ -487,7 +552,11 @@ impl AdminServiceShared {
                                 );
                                 self.send_event(&mgmt_type, event);
 
-                                info!("committed changes for new circuit proposal {}", circuit_id);
+                                info!(
+                                    "committed changes for new ciircuit proposal to create circuit \
```
agunde406, line 556: typo ciircuit -> circuit

libsplinter/src/admin/service/shared.rs
```diff
@@ -705,6 +789,66 @@ impl AdminServiceShared {
...
+                let mut protocol = ADMIN_SERVICE_PROTOCOL_VERSION;
+                for member in circuit_proposal.get_circuit_proposal().get_members() {
+                    verifiers.push(admin_service_id(member.get_node_id()));
+                    // Figure out what protocol version should be used for this proposal
+                    if let Some(protocol_version) = self
+                        .service_protocols
+                        .get(&admin_service_id(member.get_node_id()))
+                    {
+                        if protocol_version < &protocol {
+                            protocol = *protocol_version
```
agunde406, line 829: This protocol isnt used. If the protocol version is 1, it should reject this proposal right away since Disband is not supported.

agunde406 (changes requested):
This should be behind an experimental feature.

libsplinter/src/admin/service/shared.rs

libsplinter/src/admin/service/shared.rs
shannynalayna: Fixed to send this protocol to the `validate_disband_circuit` method, and added a check to this method for the protocol (returning an error if it's 1) and added a test `test_validate_disband_circuit_invalid_protocol` to validate this.

agunde406 (approved):

libsplinter/src/admin/service/messages/mod.rs
```diff
@@ -210,6 +210,75 @@ impl CreateCircuit {
 
         Ok(create_request)
     }
+
+    pub fn into_circuit_proto(self) -> Result<admin::Circuit, MarshallingError> {
```
peterschwarz, line 214: This seems a bit odd to include here, in a struct that transforms to several different types. Perhaps in a future PR, this should be moved to a `TryInto` implementation, or some other strategy.

libsplinter/src/admin/service/error.rs
```diff
@@ -161,6 +161,10 @@ pub enum AdminSharedError {
         context: String,
         source: Option<InitializeServiceError>,
     },
+    ServiceShutdownFailed {
```
peterschwarz, line 167: This enum should be reduced to have the types defined in the `crate::error` module.  Another note to make for a future PR

libsplinter/src/admin/store/circuit.rs
```diff
@@ -92,6 +95,56 @@ impl Circuit {
     }
 }
 
+impl TryFrom<&admin::Circuit> for Circuit {
+    type Error = InvalidStateError;
+
+    fn try_from(proto: &admin::Circuit) -> Result<Self, Self::Error> {
```
peterschwarz, line 101: This should be in a separate commit.

shannynalayna: Addressed

libsplinter/src/admin/service/messages/mod.rs
shannynalayna: I added these comments (besides the separate commit one; I addressed in the PR) to the `circuit-disband` feature doc so these will get revisited in the stabilization process

agunde406 (approved):

peterschwarz (approved):

PR #1152, opened 2021-01-26
Update `circuit_status` handling

shannynalayna:
Makes a few small updates to the handling of the `circuit_status` to ensure it is represented correctly in the database and in proto messages. Updates this added column to default to 1, as well as changing the postgres-specific migration file to have the type `SMALLINT` rather than `INTEGER`. Also updates the `into_proto` methods to exclude the `circuit_status` field from being set when converting into the proto message. 

This also includes a 1 line bug fix where the `circuit_version` field was not being set in the postgres implementation of the admin store's `upgrade` operation.

This can be tested by running a splinter node (could be on stable, experimental is not required to test) and submitting a circuit proposal, then attempting to view that proposal via the CLI (`splinter circuit proposals`) or REST API (`/admin/proposals`). Before, you would see an error that `circuit_status` was unset from the database (set to 0). This PR both changes the default of this column to be 1 (Active) and also updates the type the `circuit_status` is converted into/from in the database so you should successfully be able to view the circuit proposals. (Before via the CLI, an error would occur that the ProposalStore returned some type of 500 error.)

### DIFF
--- a/cli/src/action/circuit/builder.rs
+++ b/cli/src/action/circuit/builder.rs
@@ -15,8 +15,8 @@
 #[cfg(feature = "circuit-auth-type")]
 use splinter::admin::messages::AuthorizationType;
 use splinter::admin::messages::{
-    BuilderError, CreateCircuit, CreateCircuitBuilder, SplinterNode, SplinterNodeBuilder,
-    SplinterServiceBuilder,
+    BuilderError, CircuitStatus, CreateCircuit, CreateCircuitBuilder, SplinterNode,
+    SplinterNodeBuilder, SplinterServiceBuilder,
 };
 
 use crate::error::CliError;
@@ -36,6 +36,7 @@ pub struct CreateCircuitMessageBuilder {
     comments: Option<String>,
     display_name: Option<String>,
     circuit_version: Option<i32>,
+    circuit_status: Option<CircuitStatus>,
 }
 
 impl CreateCircuitMessageBuilder {
@@ -51,6 +52,7 @@ impl CreateCircuitMessageBuilder {
             comments: None,
             display_name: None,
             circuit_version: None,
+            circuit_status: None,
         }
     }
 
@@ -244,6 +246,10 @@ impl CreateCircuitMessageBuilder {
         self.circuit_version = Some(circuit_version);
     }
 
+    pub fn set_circuit_status(&mut self, circuit_status: CircuitStatus) {
+        self.circuit_status = Some(circuit_status);
+    }
+
     pub fn build(mut self) -> Result<CreateCircuit, CliError> {
         let circuit_builder = self.create_circuit_builder();
 
@@ -313,6 +319,10 @@ impl CreateCircuitMessageBuilder {
 
         if let Some(circuit_version) = self.circuit_version {
             create_circuit_builder = create_circuit_builder.with_circuit_version(circuit_version);
+        }
+
+        if let Some(circuit_status) = self.circuit_status {
+            create_circuit_builder = create_circuit_builder.with_circuit_status(&circuit_status);
         }
 
         #[cfg(feature = "circuit-auth-type")]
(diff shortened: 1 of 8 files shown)


agunde406 (approved):
This fixes the issues i was seeing

libsplinter/src/admin/store/diesel/models.rs
```diff
@@ -24,7 +24,7 @@ use diesel::{
     deserialize::{self, FromSql},
     expression::{helper_types::AsExprOf, AsExpression},
     serialize::{self, Output, ToSql},
-    sql_types::SmallInt,
+    sql_types::Integer,
```
peterschwarz, line 27: For which backend is this the case?

cli/src/action/circuit/builder.rs
```diff
@@ -315,6 +321,10 @@ impl CreateCircuitMessageBuilder {
             create_circuit_builder = create_circuit_builder.with_circuit_version(circuit_version);
         }
 
+        if let Some(circuit_status) = self.circuit_status {
```
peterschwarz, line 324: Isn’t this for creating proposals?  Why does the status need to be part of the proposal?

libsplinter/src/admin/store/diesel/operations/upgrade.rs
```diff
@@ -67,6 +67,7 @@ impl<'a> AdminServiceStoreUpgradeProposalToCircuitOperation
                 .with_durability(proposed_circuit.durability())
                 .with_routes(proposed_circuit.routes())
                 .with_circuit_management_type(proposed_circuit.circuit_management_type())
+                .with_circuit_version(proposed_circuit.circuit_version())
```
peterschwarz, line 70: Shouldn’t the status be a outcome of the proposal’s application?

libsplinter/src/admin/store/diesel/models.rs
shannynalayna: Postgres was producing errors where the `SmallInt` type was unable to be converted, and would always return a `0` despite the column being set to an actual value (ie, 1). This resulted in an error from the `FromSql` implementation that would make it unable to view proposals at all.

libsplinter/src/admin/store/diesel/operations/upgrade.rs
shannynalayna: The `status` is being set based on the `CircuitProposal` produced by the admin service, so if it is a `CircuitDisbandRequest`, the status of the `CircuitProposal` is `Disbanded`. Then, member nodes are able to vote on that proposal to actually disband the circuit, in which case the active `Circuit` in the admin store is updated to show the `disbanded` status.

shannynalayna: This fix (adding the `circuit_version` here) was a bug fix where the circuit version was not being accurately carried over when upgrading a proposal to a circuit. As the default of this column is 1, all circuits upgraded would default to 1 (despite possibly being a version 2 circuit) so this line fixes that, specifically.

cli/src/action/circuit/builder.rs
shannynalayna: I explained the reasoning for the status being in a proposal in a response to another comment, if you have further questions, let me know.

shannynalayna: It's checking here first if this value is set to allow for 0.4 compatibility, when creating a proto version of this struct.

libsplinter/src/admin/store/diesel/models.rs
shannynalayna: Changing the postgres migration file from `INTEGER` to `SMALLINT` has solved the error described in this commit, so I added that change to the initial commit that also sets the default in the migration files. And removed the commit that changed the representation from an i16 -> i32

peterschwarz (dismissed):

agunde406 (changes requested):
When testing with the CLI, I am unable to accept a proposal for creating a 0.4 compat proposal. 

```
splinterd-node-bubba    | [2021-01-27 17:58:58.636] T["consensus-admin::bubba-node-000"] ERROR [splinter::consensus::two_phase] failed to accept proposal 32643762376631306631643130336230613631636461303665336265313738396530653831333663346262663935353033613239326233623731353864343162 due to error: failed to commit proposal: error using store Proposed circuit WnkEr-fBWQF must be ACTIVE
```

agunde406 (approved):
tested

peterschwarz (approved):

PR #1154, opened 2021-01-26
Add biome feature to guard `splinter::biome` mod

ltseeley:
Adds the `biome` feature to libsplinter to guard the `splinter::biome`
module. All biome sub-features and oauth depend on this feature. This
simplifies the feature guard for the biome module.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -76,6 +76,7 @@ default = []
 
 stable = [
     "admin-service",
+    "biome",
     "biome-credentials",
     "biome-key-management",
     "circuit-template",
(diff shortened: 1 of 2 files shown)


libsplinter/Cargo.toml
```diff
@@ -128,17 +129,18 @@ authorization-handler-allow-keys = ["authorization"]
 authorization-handler-maintenance = ["authorization"]
 authorization = ["rest-api"]
 authorization-handler-rbac = ["authorization"]
-biome-credentials = ["bcrypt"]
-biome-key-management = []
-biome-notifications = []
+biome = []
```
peterschwarz, line 132: May be useful to comment on this feature being internal - it doesn't provide any functionality on its own

vaporos: But... if that changes, do we need to remember to update the comment here? This is conceptually similar to admin-service, registry, and some of the transact features which just enable the parent module.

vaporos (approved):

peterschwarz (approved):

PR #1158, opened 2021-01-27
Add `CircuitDisbandRequest` functionality to CLI

shannynalayna:
Makes a few updates to the current `CircuitDisbandRequest` implementation, including:

- Adds a queue for circuits that are being disbanded, similar to the `pending_consensus_proposals`, for the disband proposal to wait until all member nodes have reached consensus and sent the `DisbandedCircuit` message. Once all members have sent this message, the disband is finalized (services shut down, peer refs removed, circuit removed from routing table)
- To support the process described above, added a `DisbandedCircuit` admin message to be sent between members of a circuit being disbanded, to signal that the disbanded circuit has been accepted and committed to state
- Handling of this admin message, similar to the `MemberReady` message, in which the receiver adds the member node it received the message from to a `ready_members` list associated with the disband proposal. 
- Also adds functions to support the functionality explained above, mainly to add a circuit to the `pending_consensus_disbanded_circuits`, and check if/when members are ready to finalize the disband

Also adds the `CircuitDisbandRequest` functionality to the CLI, behind the experimental `circuit-disband` feature and updates how proposals are displayed to more accurately reflect their intention.

To test: Run two splinter nodes with `--features=stable,circuit-disband`. Create a circuit as you normally would, then use `splinter circuit disband <circuit_id>` to propose disbanding the circuit you just created. Vote to accept this proposal from the member node. Verify in the logs that the `DisbandedCircuit` was sent between the nodes. Currently, all circuits are still displayed so you will still see the disbanded circuit using the `splinter circuit list` so you can check the database to view the Circuit, where the `circuit_status` should be 2 (which is equivalent to the DB's i16 representation of the `Disbanded` variant.)

### DIFF
--- a/cli/man/splinter-circuit-disband.1.md
+++ b/cli/man/splinter-circuit-disband.1.md
@@ -1,0 +1,94 @@
+% SPLINTER-CIRCUIT-DISBAND(1) Cargill, Incorporated | Splinter Commands
+<!--
+  Copyright 2018-2021 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**splinter-circuit-disband** — Submits a request to disband the specified circuit.
+
+SYNOPSIS
+========
+**splinter circuit disband** \[**FLAGS**\] \[**OPTIONS**\] CIRCUIT-ID
+
+DESCRIPTION
+===========
+Request to disband a circuit by specifying the circuit ID of the circuit to be
+disbanded. Disbanding a circuit removes a circuit's networking functionality.
+Once all members of the circuit have accepted the request to disband the
+circuit, the circuit is only available offline. This functionality is currently
+behind the experimental `circuit-disband` feature and must be enabled to use
+this command.
+
+The `disband` command creates a new circuit proposal to reflect the disbanded
+state, with the proposed circuit's `circuit_status` field set to `Disbanded`.
+This proposal is then able to be voted on, similar to other circuit proposals.
+
+The generated ID of the existing circuit can be viewed using the
+`splinter-circuit-list` command and this circuit ID is used to specify the
+circuit to be disbanded. Once the disband request has been submitted,
+the proposal created (and other circuit proposals) can be viewed using the
+`splinter-circuit-proposals` command.
+
+The disband proposal must be accepted by all members before the existing
+circuit is updated to reflect the disbanded state. If all nodes have agreed to
+disband the circuit, the disbanded circuit may be viewed using the
+`splinter-circuit-show` command.
+
+FLAGS
+=====
+`-h`, `--help`
+: Prints help information.
+
+`-q`, `--quiet`
+: Decrease verbosity (the opposite of -v). When specified, only errors or
+  warnings will be output.
+
+`-V`, `--version`
+: Prints version information.
+
+`-v`
+: Increases verbosity (the opposite of -q). Specify multiple times for more
+  output.
+
+OPTIONS
+=======
+`-k`, `--key` PRIVATE-KEY-FILE
+: Specifies the full path to the private key file.
+
+`-U`, `--url` URL
+: Specifies the URL for the `splinterd` REST API. The URL is required unless
+  `$SPLINTER_REST_API_URL` is set.
+
+ARGUMENTS
+=========
+`CIRCUIT-ID`
+: Specify the circuit ID of the circuit to be disbanded.
+
+EXAMPLES
+========
+* The proposed circuit has ID `1234-ABCDE`.
+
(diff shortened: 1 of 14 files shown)


libsplinter/src/admin/store/diesel/operations/update_circuit.rs
```diff
@@ -61,6 +61,7 @@ impl<'a> AdminServiceStoreUpdateCircuitOperation
                     circuit::durability.eq(circuit_model.durability),
                     circuit::routes.eq(circuit_model.routes),
                     circuit::circuit_management_type.eq(circuit_model.circuit_management_type),
+                    circuit::circuit_status.eq(circuit_model.circuit_status),
```
agunde406, line 64: Does this need to be added below  to the sqlite implementation as well?

splinterd/Cargo.toml
```diff
@@ -116,6 +116,7 @@ authorization-handler-rbac = [
 biome-credentials = ["database", "splinter/biome-credentials"]
 biome-key-management = ["database", "splinter/biome-key-management"]
 biome-profile = ["splinter/biome-profile"]
+circuit-disband = ["splinter/circuit-disband"]
```
agunde406, line 119: This should be behind experimental in splintered as well, it should be stabilized at the same time as circuit-disband in libsplinter

agunde406 (changes requested):
Please include the new and updated man pages for the cli changes.

libsplinter/src/admin/store/diesel/operations/update_circuit.rs
shannynalayna: Yes it does! thanks for noticing

shannynalayna: Updated

splinterd/Cargo.toml
shannynalayna: Updated

libsplinter/src/admin/service/error.rs
```diff
@@ -298,13 +298,17 @@ impl std::fmt::Display for AdminConsensusManagerError {
 pub enum AdminError {
     ConsensusFailed(AdminConsensusManagerError),
     MessageTypeUnset,
+    #[cfg(not(feature = "circuit-disband"))]
```
peterschwarz, line 301: Why does this need to be guarded by a feature?  Seems handy to have when we introduce new messages in general

shannynalayna: I added this error to fix a compiler warning happening in the `handle_message` admin service's function for when the disband functionality isn't compiled in because there is a match statement based on the `AdminMessage`'s message type (which is totally handled when the `circuit-disband` feature is enabled) and the error is in a catch-all case at the end when that message type isn't being handled. Currently, that function has a match statement that handles all of the possible cases.

agunde406 (approved):

agunde406 (approved):

cli/man/splinter-circuit-disband.1.md
```diff
@@ -0,0 +1,94 @@
...
+Request to disband a circuit by specifying the circuit ID of the circuit to be
+disbanded. Disbanding a circuit removes a circuit's networking functionality.
+Once all members of the circuit have accepted the request to disband the
+circuit, the circuit is only available offline. This functionality is currently
+behind the experimental `circuit-disband` feature and must be enabled to use
+this command.
+
+The `disband` command creates a new circuit proposal to reflect the disbanded
+state, with the proposed circuit's `circuit_status` field set to `Disbanded`.
+This proposal is then able to voted on, similar to other circuit proposals.
```
isabeltomb, line 28: "be" should be added between "to" and "voted"

shannynalayna: Updated

agunde406 (dismissed):

agunde406 (approved):

libsplinter/src/admin/store/diesel/operations/list_nodes.rs
```diff
@@ -61,7 +61,10 @@ where
         // endpoints associated with the `node_ids` in a HashMap.
         nodes_info.into_iter().for_each(|(node, node_endpoint)| {
             if let Some(endpoint_list) = node_map.get_mut(&node.node_id) {
-                endpoint_list.push(node_endpoint.endpoint);
+                // Ensure only unique endpoints are added to the node's endpoint list
```
peterschwarz, line 64: you could sort the list and dedup

shannynalayna: Updated the last commit to use the combo of `sort` and `dedup`

peterschwarz (approved):

agunde406 (approved):

isabeltomb (approved):
Tested

PR #1162, opened 2021-01-29
Add `CircuitStatus` to `CircuitPredicate`

shannynalayna:
Adds a `CircuitStatus` circuit predicate to allow for filtering circuits and proposals based on the `circuit_status` field. This also affects the circuits returned by the admin store's `list_circuits` operation, as only `Active` circuits are returned by default. 

The `CircuitStatus` predicate is also added to the `/admin/proposals` and `/admin/circuits/` endpoints, via the `status` query parameter. 

Also adds a `circuit-status` option to the `splinter circuit list`, `splinter circuit proposals` commands. 

Also makes a small update to the data displayed via the `show` command, to display the `circuit_status` field of the circuit / proposed circuit.

### DIFF
--- a/cli/man/splinter-circuit-show.1.md
+++ b/cli/man/splinter-circuit-show.1.md
@@ -76,13 +76,14 @@ The information displayed below will appear the same on all proposed member node
 If all member nodes vote to accept the circuit, the `splinter-circuit-show`
 command will display the same information, without the `Vote` as all nodes would
 have accepted the proposal. If any of the member nodes vote to reject the circuit,
-the proposal will not be viewable by any nodes.
+the proposal will be removed and is not viewable by any nodes.
 
 ```
 $ splinter circuit show 01234-ABCDE \
   ---url URL-of-alpha-node-splinterd-REST-API
 Proposal to create: 01234-ABCDE
     Display Name: -
+    Circuit Status: Active
     Version: 2
     Management Type: mgmt001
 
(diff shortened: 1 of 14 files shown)


cli/src/action/circuit/api.rs
```diff
@@ -245,18 +246,25 @@ pub struct CircuitSlice {
...
+    pub circuit_status: Option<CircuitStatus>,
 }
 
 impl fmt::Display for CircuitSlice {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut display_string = format!("Circuit: {}\n    ", self.id,);
 
         if let Some(display_name) = &self.display_name {
-            display_string += &format!("Display Name: {}\n    ", display_name);
+            display_string += &format!("Display Name: {:?}\n    ", display_name);
```
agunde406, line 257: Was this change intentional?

shannynalayna: Oh, no it was not. Fixed

cli/src/action/circuit/api.rs
```diff
@@ -257,6 +259,12 @@ impl fmt::Display for CircuitSlice {
             display_string += "Display Name: -\n    ";
         }
 
+        if let Some(status) = &self.circuit_status {
+            display_string += &format!("Circuit Status: {:?}\n    ", status);
```
peterschwarz, line 263: This should  match on status and output a string, otherwise, it will display the qualified variant, like `CircuitStatus::Active`

peterschwarz (changes requested):

cli/src/action/circuit/api.rs
shannynalayna: Updated to include using a `From` implementation on the `CircuitStatus` that is displayed here (from the `admin::messages` module) to convert the full variant to just `Active/Disbanded/Abandoned`, depending.

cli/man/splinter-circuit-show.1.md
```diff
@@ -104,6 +105,42 @@ Proposal to create: 01234-ABCDE
                 AA01
 ```
 
+The information displayed below will appear the same on all proposed member
+nodes for a circuit proposal to disband an existing circuit.
+If all member nodes vote to accept disbanding the circuit, the
+`splinter-circuit-show` command will display the same information, without the
+`Vote` as all nodes would have accepted the proposal. If any of the member
+nodes vote to reject disbanding the circuit, the proposal will not be viewable
```
agunde406, line 113: "the proposal will not be viewable" this is kinda misleading because really the proposal will no longer exist.

agunde406 (changes requested):
Tested after rebasing on https://github.com/Cargill/splinter/pull/1158

cli/man/splinter-circuit-show.1.md
shannynalayna: Updated here to say the `the proposal will be removed and is not viewable by any nodes`, and also clarified the wording above where the same sentence was used (and the same circumstances pertain, the proposal is not only no longer viewable, but has been removed)

cli/man/splinter-circuit-show.1.md
```diff
@@ -76,13 +76,14 @@ The information displayed below will appear the same on all proposed member node
 If all member nodes vote to accept the circuit, the `splinter-circuit-show`
 command will display the same information, without the `Vote` as all nodes would
 have accepted the proposal. If any of the member nodes vote to reject the circuit,
-the proposal will not be viewable by any nodes.
+the proposal will be removed and is not be viewable by any nodes.
```
agunde406, line 79: " not be viewable by any nodes" -> " not viewable by any nodes"

agunde406 (changes requested):

cli/man/splinter-circuit-show.1.md
shannynalayna: Updated

agunde406 (approved):

agunde406 (approved):

peterschwarz (approved):

PR #1163, opened 2021-01-29
RBAC CLI: role subcommands (create, update, delete, and show)

peterschwarz:
This PR adds the `role` sub-commands `show`, `create`, `update`, `delete` to the CLI.

# Testing

## Setup a local splinter node

These operations for setup can be done in one terminal.

First, initialize your splinter instance:

```
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml --features experimental -- cert generate
$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml --features experimental -- database migrate
```

Also, generate a key and add it to the allow_keys file:
```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- keygen
```

Start your splinter node:
```
SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features experimental -- --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080
```

## Create a Role

These remainder of the operations in these test instructions can be done in a second terminal. 

List the current roles:

```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role list
ID  DISPLAY NAME
```
(if you are testing on node with existing roles, this may look different).

Add a role:
```
$  cargo run --manifest-path cli/Cargo.toml --features experimental -- \
    role create \
    --display "Status Reader" \
    --permission "status.read" \
     test_role
```

Now, list the roles again:

```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role list
ID        DISPLAY NAME
test_role Status Reader
```

## Show a Role

Display the role just created:

```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role show test_role
Id: test_role
    Display Name: Status Reader
    Permissions:
        status.read
```
## Update the role

Let's change the role to be about circuit reads:
```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role update \
   --rm-perm status.read \
   --add-perm circuit-read \
   --display "Circuit Reader" \
   test_role
```
And display it again:
```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- role show test_role
Id: test_role
    Display Name: Circuit Reader
    Permissions:
        circuit-read
```

## Delete the role

```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role delete test_role
```

Verify the deletion:
```
$ cargo run --manifest-path cli/Cargo.toml --features experimental -- \
   role show test_role
ERROR: Subcommand encountered an error: Failed to get role test_role: Role not found
```

### DIFF
--- a/cli/man/splinter-role-create.1.md
+++ b/cli/man/splinter-role-create.1.md
@@ -1,0 +1,103 @@
+% SPLINTER-ROLE-CREATE(1) Cargill, Incorporated | Splinter Commands
+<!--
+  Copyright 2018-2021 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**splinter-role-create** — Creates a role on a Splinter node
+
+SYNOPSIS
+========
+**splinter role create** \[**FLAGS**\] \[**OPTIONS**\] ROLE-ID
+
+DESCRIPTION
+===========
+Creates a role that specifies a set of permissions for accessing the REST API
+on a Splinter node. This operation only effects the node itself and not the
+wider network.
+
+FLAGS
+=====
+`-h`, `--help`
+: Prints help information
+
+`-q`, `--quiet`
+: Decrease verbosity (the opposite of -v). When specified, only errors or
+  warnings will be output.
+
+`-V`, `--version`
+: Prints version information
+
+`-v`
+: Increases verbosity (the opposite of -q). Specify multiple times for more
+  output.
+
+OPTIONS
+=======
+`-k`, `--key` PRIVATE-KEY-FILE
+: Specifies the private signing key (either a file path or the name of a
+  .priv file in $HOME/.splinter/keys).
+
+`-U`, `--url` URL
+: Specifies the URL for the `splinterd` REST API. The URL is required unless
+  `$SPLINTER_REST_API_URL` is set.
+
+`-D`, `--display` DISPLAY-NAME
+: Specifies the display name for the created role. This is a required option.
+
+`-P`, `--permission` PERMISSION
+: Specifies a permission to be included in the role. Specify multiple times for
+  more permissions. At least one permission is required.
+
+
+ARGUMENTS
+=========
+`ROLE-ID`
+: Specify the role ID of the role to be created.
+
+EXAMPLES
+========
+This example creates an `circuit_admin` role on a splinter node.
+
+* The role has ID `circuit_admin`.
+* The role has a display name of `"Circuit Admin"`
+* The role has permissions to read and write circuits
+
+
+```
+$ splinter role create \
+  --url URL-of-splinterd-REST-API \
+  --permission circuit.read \
+  --permission circuit.write \
+  --display "Circuit Admin" \
+  circuit_admin
+```
+
+This can be verified by using the `role show` command:
+
+```
+$ splinter role show \
+  --url URL-of-splinterd-REST-API \
+  circuit_admin
+ID: circuit_admin
+    Name: Circuit Administrator
+    Permissions:
+        circuit.read
+        circuit.write
+```
+
+ENVIRONMENT VARIABLES
+=====================
+**SPLINTER_REST_API_URL**
+: URL for the `splinterd` REST API. (See `-U`, `--url`.)
+
+SEE ALSO
+========
+| `splinter-role-list(1)`
+| `splinter-role-show(1)`
+|
+| Splinter documentation: https://www.splinter.dev/docs/0.5/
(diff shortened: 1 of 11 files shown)


cli/src/action/api/rbac.rs
```diff
@@ -413,6 +413,37 @@ pub fn update_role(base_url: &str, auth: &str, role_update: RoleUpdate) -> Resul
...
+        .and_then(|res| {
+            let status = res.status();
+            if status.is_success() {
+                Ok(())
+            } else {
+                let message = res
+                    .json::<super::ServerError>()
+                    .map_err(|_| {
+                        CliError::ActionError(format!(
+                            "Get role delete request failed with status code '{}', but error \
```
shannynalayna, line 432: This error message looks a little wonky - seems like a `GET` request was just made

cli/src/action/api/rbac.rs
```diff
@@ -43,6 +43,68 @@ impl fmt::Display for Role {
     }
 }
 
+/// Constructs roles for submission to a splinter node.
```
shannynalayna, line 46: Typo in commit message 'changes' -> 'change'

cli/man/splinter-role-create.1.md
```diff
@@ -0,0 +1,103 @@
...
+<!--
+  Copyright 2018-2021 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**splinter-role-create** — Displays information about a role
```
agunde406, line 11: Copy pasta

cli/man/splinter-role-update.1.md
```diff
@@ -0,0 +1,106 @@
...
+<!--
+  Copyright 2018-2021 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**splinter-role-update** — Displays information about a role
```
agunde406, line 11: copypasta

cli/man/splinter-role-update.1.md
```diff
@@ -0,0 +1,106 @@
...
+This example updates the `circuit_admin` role on a splinter node.
+
+* The role has ID `circuit_admin`.
+* The role has a display name of `"Circuit Admin"`
+* The role has the permissions circuit read and write
+* The role will be updated to have the permission `status.read`
+
+
+```
+$ splinter role create \
```
agunde406, line 75: Should say `update` not `create`

agunde406 (changes requested):
Tested

cli/man/splinter-role-create.1.md
agunde406, line 11: copy pasta

agunde406 (changes requested):

cli/src/action/api/rbac.rs
peterschwarz: Updated.

cli/src/action/api/rbac.rs
peterschwarz: Fixed

cli/man/splinter-role-update.1.md
peterschwarz: fixed

cli/man/splinter-role-update.1.md
peterschwarz: fixed

cli/man/splinter-role-create.1.md
peterschwarz: fixed

agunde406 (approved):

vaporos:
Suggest that in show, we change:

```
Id: test_role
    Display Name: Status Reader
    Permissions:
        status.read
```

to

```
test_role - Status Reader
    Permissions:
        status.read
```

or

```
test_role
    Name: Status Reader
    Permissions:
        status.read
```

vaporos:
Currently, the command does not produce an error if the we attempt to remove a permission that doesn't exist, or if we attempt to add a permission that already exists. If this is intentional, this I suggest a) the command should print what it is updating and what didn't need to change; and b) we should have an option like --strict which prevents the command from running if one of these conditions occur. I think though, in that case, it might be better to flip it around and implement --force as the option which makes it succeed even if its asked to do things that don't apply.

vaporos:
With the current command, it is not obvious how one would set specific permissions to a role that may already have permissions. This could be accomplished by adding something like "--rm-all" or "--clear" which removes all the ones that currently exists and then adds all the specified ones, so that effectively we get only the ones we are listing.

vaporos:
ERROR: Subcommand encountered an error: Failed to get role test_role: Role not found

should really be simple, like:

ERROR: Role does not exist

vaporos (approved):

peterschwarz:
Updated the error messages, will update the `update` command arguments in a followup PR.

peterschwarz:
Also, added a commit to change "Display Name" to "Name" in the list output.  The show command output has been corrected similarly, but in their original commits

shannynalayna (approved):
Tested

agunde406 (approved):

PR #1167, opened 2021-02-02
Add --rm-all to 'role update' subcommand

peterschwarz:
This PR adds a `--rm-all` flag to the `role update` subcommand.  This allows the user to remove the current permissions from the role without needing to enumerate them.

Additionally, 

- if the user adds and removes the same permission, the command will return an error.
- if the user attempts to remove a permission that does not currently belong to the role, return an error


ltseeley (approved):

ltseeley (dismissed):

ltseeley (approved):

shannynalayna (approved):
Tested

PR #1168, opened 2021-02-02
`AdminServiceEventStore` stabilization updates

shannynalayna:
Includes updates requested in the stabilization process for the `admin-service-event-store` and (as of this PR) the `admin-service-event-postgres` experimental features including:

- Moves the `event` module from `admin::service::event` to `admin::store::events`.
- Updates the memory implementation of the `AdminServiceEventStore` to store events as the module's own `AdminServiceEvent` definition(in `admin::store::events`), rather than the `AdminServiceEvent` defined in `admin::messages`. 
- Changes the `admin-service-event-store-diesel` -> `admin-service-event-store-postgres` and fixes the feature guards affected by this change. Where `#[cfg(feature = "admin-service-event-store-diesel")]` is used within the `events::store::diesel` module (which is guarded by the `admin-service-event-store` feature) now use `#[cfg(any(feature = "admin-service-event-store-postgres", feature = "sqlite"))]`


libsplinter/Cargo.toml
```diff
@@ -99,7 +99,7 @@ experimental = [
     "stable",
     # The following features are experimental:
     "admin-service-event-store",
-    "admin-service-event-store-diesel",
+    "admin-service-event-store-postgres",
```
agunde406, line 102: libsplinter/src/store should also be updated for this feature rename

agunde406 (changes requested):

libsplinter/Cargo.toml
shannynalayna: Updated the `admin-service-event-store-diesel` instances with a combo of the `admin-service-event-store` guarding the sqlite store functions, and `admin-service-event-store-postgres` in the postgres store implementation

peterschwarz (approved):

agunde406 (approved):

dnewh (approved):

dplumb94 (approved):

PR #1172, opened 2021-02-04
RBAC REST API: Fix 404 on update

peterschwarz:
This PR fixes an issue where an update on a nonexistent role or assignment doesn't properly report a 404 vs constraint violations.

Likewise, it fixes an issue where foreign key violations were not properly being reported as constraint violations, causing a 500 error, not a 409

### DIFF
--- a/libsplinter/src/rest_api/auth/authorization/rbac/store/mod.rs
+++ b/libsplinter/src/rest_api/auth/authorization/rbac/store/mod.rs
@@ -353,7 +353,7 @@ pub trait RoleBasedAuthorizationStore: Send + Sync {
     ///
     /// # Errors
     ///
-    /// Returns a `InvalidState` error if the role does not exist.
+    /// Returns a `ConstraintViolation` error if the role does not exist.
     fn update_role(&self, role: Role) -> Result<(), RoleBasedAuthorizationStoreError>;
 
     /// Removes a role.
@@ -395,7 +395,7 @@ pub trait RoleBasedAuthorizationStore: Send + Sync {
     ///
     /// # Errors
     ///
-    /// Returns a `InvalidState` error if the assignment does not exist.
+    /// Returns a `ConstraintViolation` error if the assignment does not exist.
     fn update_assignment(
         &self,
         assignment: Assignment,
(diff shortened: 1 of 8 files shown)


libsplinter/src/rest_api/auth/authorization/maintenance/mod.rs
```diff
@@ -248,7 +248,7 @@ mod tests {
             unimplemented!()
         }
 
-        fn update_role(&self, _role: Role) -> Result<(), RoleBasedAuthorizationStoreError> {
+        fn update_role(&self, _role: Role) -> Result<bool, RoleBasedAuthorizationStoreError> {
```
ltseeley, line 251: I wonder if it would make more sense to add a `ConstraintViolationType::NotFound` variant? If we decide to go with what you have here, we should probably update the data store guidelines to describe this approach.

vaporos (approved):

libsplinter/src/rest_api/auth/authorization/maintenance/mod.rs
vaporos: I'm fine with either approach.

shannynalayna (approved):

ltseeley (approved):

libsplinter/src/rest_api/auth/authorization/maintenance/mod.rs
peterschwarz: I've updated this to use the `ConstraintViolationType::NotFounnd`, which is also added in this PR.

libsplinter/src/rest_api/auth/authorization/rbac/store/mod.rs
```diff
@@ -351,6 +351,9 @@ pub trait RoleBasedAuthorizationStore: Send + Sync {
 
     /// Updates a role.
     ///
+    /// # Returns true if the record exists, and was successfully updated, false if it does not
```
vaporos, line 354: This comment doesn't match the function signature (and another instance this).

vaporos (changes requested):

vaporos (approved):

agunde406 (approved):

PR #1174, opened 2021-02-05
Merge `AdminServiceEventStore` into `AdminServiceStore`

shannynalayna:
This PR merges the `AdminServiceEventStore` into the `AdminServiceStore`. This removes the memory-backed implementation of the `AdminServiceEventStore` and merges the methods for this trait into the `AdminServiceStore` trait. 

This PR also adds the Diesel implementations to the `AdminServiceStore` `diesel` module and includes tests for the newly implemented operations. (These tests were moved from the `AdminServiceEventStore` `diesel` module.)

This change also addresses the update where the `AdminServiceEventStore` was used, mainly in the admin service. This PR also updates the shared admin service to update the feature-guarded `admin_event_store` to take the `AdminServiceStore` trait to allow for different backend configurations for the admin store and the event store. This will mainly be necessary if the `YAML` storage option is chosen (in which case the `YamlAdminServiceStore` will be used for the admin store, and the SQLite-backed admin store will be used for the event store).


libsplinter/src/admin/store/event.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright 2018-2020 Cargill Incorporated
```
agunde406, line 1: The copyright in this PR should be updated to 2021

libsplinter/src/admin/store/event.rs
```diff
@@ -53,6 +53,97 @@ impl AdminServiceEvent {
     }
 }
 
+/// Builder to be used to build a `Circuit`
```
agunde406, line 56: `Circuit` -> `AdminServiceEvent`

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
```diff
@@ -26,7 +26,7 @@ use crate::admin::service::{
     AdminCommands, AdminServiceEventSubscriber, AdminServiceStatus, AdminSubscriberError,
 };
 #[cfg(feature = "admin-service-event-store")]
-use crate::admin::store::events;
```
agunde406, line 29: Typo in commit message `This includes a removing `

agunde406 (changes requested):

libsplinter/src/admin/store/event.rs
shannynalayna: Updated

libsplinter/src/admin/store/event.rs
shannynalayna: Updated

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
shannynalayna: Ended up re-wording this commit after updating the admin service to still take an "event_store" (that uses the admin store's trait) so I believe this sentence/typo was removed entirely.

libsplinter/src/admin/store/diesel/operations/add_event.rs
```diff
@@ -0,0 +1,228 @@
+// Copyright 2018-2020 Cargill Incorporated
```
agunde406, line 1: Missed 2021 update

libsplinter/src/admin/store/diesel/operations/list_events.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2021 Cargill Incorporated
+// Copyright 2018-2020 Cargill Incorporated
```
agunde406, line 1: Missed 2021 update

libsplinter/src/admin/store/diesel/operations/list_events_since.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2021 Cargill Incorporated
+// Copyright 2018-2020 Cargill Incorporated
```
agunde406, line 1: Missed 2021 update

agunde406 (changes requested):

splinterd/src/daemon.rs
```diff
@@ -510,8 +526,8 @@ impl SplinterDaemon {
             Box::new(AllowAllKeyPermissionManager),
             Some(self.admin_timeout),
             routing_writer.clone(),
-            #[cfg(feature = "admin-service-event-store")]
-            admin_event_store,
+            #[cfg(feature = "service-arg-validation")]
```
agunde406, line 529: This should be admin-service-event-store

agunde406 (changes requested):

libsplinter/src/admin/store/event.rs
```diff
@@ -53,6 +53,97 @@ impl AdminServiceEvent {
...
+}
+
+impl AdminServiceEventBuilder {
+    /// Creates a new `AdminServiceEventBuilder`
+    pub fn new() -> Self {
+        AdminServiceEventBuilder::default()
+    }
+
+    /// Returns the event ID in the builder
+    pub fn event_id(&self) -> Option<i64> {
```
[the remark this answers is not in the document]
shannynalayna: The `AdminServiceEvent` struct defined in this module. These fields are used when handling the message. The event fields are also used within the diesel operations where the events are sorted by this field and filtered by this field.

libsplinter/src/admin/store/diesel/operations/add_event.rs
shannynalayna: Updated the copyrights from the added operations files - encompassing your other comments

splinterd/src/daemon.rs
shannynalayna: Fixed

splinterd/src/daemon.rs
```diff
@@ -244,6 +239,27 @@ impl SplinterDaemon {
             }
         };
 
+        #[cfg(feature = "admin-service-event-store")]
+        let event_store = {
+            if self.storage_type.is_some() {
```
agunde406, line 244: This should not default to memory if storage_type is some. It should still be whatever the database value is even if the admin service store gets set as YamlAdminServiceStore.

agunde406 (changes requested):

libsplinter/src/admin/store/event.rs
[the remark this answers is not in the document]
peterschwarz: The intent of my question has more to do with why their are getters on the builder vs the event itself.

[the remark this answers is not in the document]
shannynalayna: Ah, yeah those weren't used for anything so removed them from that commit.

splinterd/src/daemon.rs
shannynalayna: Updated to take the `db_url` if the storage type is yaml, and then use `memory` for the `db_url` if the storage type is memory.

peterschwarz (approved):

agunde406 (approved):

PR #1175, opened 2021-02-05
Remove service state when purging a circuit

shannynalayna:
This PR makes updates to the scabbard service's `destroy` function to enable state LMDB files to also be removed (implemented within `ScabbardState`) and makes updates to the service orchestrator to account for this update. Mainly, this PR adds a new list of `stopped_services` within the orchestrator. This makes a greater differentiation within the orchestrator between stopping and destroying a circuit by splitting up the `shutdown_service` function, as `stop` will stop the circuit, and now `destroy` also removes the state files. This allows for circuits to be disbanded or abandoned with service functionality removed, and then the purge functionality is able to completely clean up the state pertaining to a circuit (including the state files) as the service is not _completely_ removed from the orchestrator, but rather moved into the `stopped_services`. 

Before the disband functionality used the `shutdown_service` orchestrator method, which both stopped and destroyed the services but did not remove the state LMDB files.  

This also updates the `re_initialize_services` method in the admin service to fetch all circuits from the admin store (rather than just the `Active` ones) to allow for services that have previously been stopped through disbanding or abandoning to be added back to the orchestrator so the associated services are able to be purged (meaning the service is destroyed rather than just stopped, and state LMDB files will be removed). Previous to this update, the `purge` request would fail because the orchestrator was unable to find the service because it was never created between restarts.

This ultimately can be tested by running two splinter nodes, creating then disbanding a circuit and then using the `splinter circuit purge` command to remove the circuit from the node's DB as well as any scabbard state files used by the node's associated services on the circuit that was purged. The reinitializing updates may be tested in a similar way, but by also restarting your node(s) between creating+disbanding the circuit, and then purging it. The `purge` should succeed even after a re-start.

### DIFF
--- a/libsplinter/src/orchestrator/mod.rs
+++ b/libsplinter/src/orchestrator/mod.rs
@@ -73,6 +73,9 @@ pub struct ServiceOrchestrator {
     inbound_router: InboundRouter<CircuitMessageType>,
     /// `running` and `join_handles` are used to shutdown the orchestrator's background threads
     running: Arc<AtomicBool>,
+    /// A (ServiceDefinition, ManagedService) map of services that have been stopped, but yet to
+    /// be completely destroyed
+    stopped_services: Arc<Mutex<HashMap<ServiceDefinition, Box<dyn Service>>>>,
 }
 
 impl ServiceOrchestrator {
@@ -86,6 +89,7 @@ impl ServiceOrchestrator {
         channel_capacity: usize,
     ) -> Result<(Self, JoinHandles<Result<(), OrchestratorError>>), NewOrchestratorError> {
         let services = Arc::new(Mutex::new(HashMap::new()));
+        let stopped_services = Arc::new(Mutex::new(HashMap::new()));
         let mesh = Mesh::new(incoming_capacity, outgoing_capacity);
         let mesh_id = format!("{}", Uuid::new_v4());
         mesh.add(connection, mesh_id.to_string())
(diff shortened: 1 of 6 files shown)


services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -357,6 +362,16 @@ impl ScabbardState {
     pub fn clear_subscribers(&mut self) {
         self.event_subscribers.clear();
     }
+
+    pub fn remove_db_files(&self) -> Result<(), ScabbardStateError> {
+        fs::remove_file(self.state_db_path.as_path()).map_err(|err| {
```
agunde406, line 367: The lock files for the lmbd files are being left behind.

agunde406 (changes requested):

agunde406 (dismissed):
Tested

agunde406 (approved):

libsplinter/src/orchestrator/mod.rs
```diff
@@ -332,6 +332,41 @@ impl ServiceOrchestrator {
             .collect())
     }
 
+    /// Create a service that has previously been stopped according to the specified definition.
+    /// The arguments provided must match those required to create the service.
+    pub fn add_stopped_service(
```
peterschwarz, line 337: Should this just be an option to add service?

shannynalayna: Well, the other `initialize_service` option in the orchestrator that adds a service would need another argument to tell that the service is 'stopped' because the orchestrator isn't really aware about that status, and is easier to handle from the admin service's perspective (hence the separate function)

peterschwarz (approved):

PR #1176, opened 2021-02-05
Update RegistryError to use top level errors

agunde406:
Replaces the existing error types with InternalError,
InvalidState, ConstraintViolationError, and
ResourceTemporarilyUnavailableError.

Also adds a error that can be used in the REST API routes.
The Registry error can no longer be used in a web::block,
because Box<dyn Error> is guaranteed to be Send. The
new error will conver the RegistryError to either InvalidState,
which is required for returning a BadRequest, and
InternalError that will be logged and InternalServerError
is returned.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -131,10 +131,11 @@ impl AdminKeyVerifier for dyn RegistryReader {
     /// registry and the node has the given key. Otherwise, the key is not permitted.
     fn is_permitted(&self, node_id: &str, key: &[u8]) -> Result<bool, AdminKeyVerifierError> {
         let node_opt = self.fetch_node(node_id).map_err(|err| {
-            AdminKeyVerifierError::new_with_source(
-                &format!("Failed to lookup node '{}' in registry", node_id),
-                Box::new(err),
-            )
+            AdminKeyVerifierError::new(&format!(
+                "Failed to lookup node '{}' in registry: {}",
+                node_id,
+                err.to_string()
+            ))
         })?;
         Ok(match node_opt {
             Some(node) => node.has_key(&to_hex(key)),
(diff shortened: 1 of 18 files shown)


libsplinter/src/admin/service/mod.rs
```diff
@@ -131,10 +131,11 @@ impl AdminKeyVerifier for dyn RegistryReader {
     /// registry and the node has the given key. Otherwise, the key is not permitted.
     fn is_permitted(&self, node_id: &str, key: &[u8]) -> Result<bool, AdminKeyVerifierError> {
         let node_opt = self.fetch_node(node_id).map_err(|err| {
-            AdminKeyVerifierError::new_with_source(
-                &format!("Failed to lookup node '{}' in registry", node_id),
-                Box::new(err),
-            )
+            AdminKeyVerifierError::new(&format!(
```
vaporos, line 134: Wondering if we can use a pattern that doesn't require this error to be Send, allowing it to keep the new_with_source() method. Maybe handling it at the channel boundary by not sending AdminKeyVerifierError directly, but sending its display and debug instead.

agunde406: We should have a larger discussion about this pattern and update this in another PR based on that conversation.

vaporos (dismissed):

libsplinter/src/admin/service/mod.rs
ltseeley: I think we avoid having to do this in some endpoints by not using `web::block` and instead just converting the HTTP response into a future like `return Box::new(HttpResponse::Ok().finish().into_future())`

libsplinter/src/registry/diesel/mod.rs
```diff
@@ -209,6 +209,7 @@ pub mod tests {
 
         assert_eq!(node, get_node_1());
 
+        #[allow(deprecated)]
```
ltseeley, line 212: This seems like a separate change

ltseeley (approved):

vaporos (approved):

PR #1180, opened 2021-02-10
Registry database stabilization updates

agunde406:


libsplinter/src/registry/diesel/mod.rs
```diff
@@ -237,6 +237,54 @@ pub mod tests {
         assert_eq!(node, get_node_1());
     }
 
+    /// Verifies that `update_node` properly updates a node
```
ltseeley, line 240: Can you add some more details to the commit message? I'm not sure I understand what this change fixes.

libsplinter/src/registry/diesel/operations/add_node.rs
```diff
@@ -37,11 +36,6 @@ pub(in crate::registry::diesel) trait RegistryAddNodeOperation {
 #[cfg(feature = "postgres")]
 impl<'a> RegistryAddNodeOperation for RegistryOperations<'a, diesel::pg::PgConnection> {
     fn add_node(&self, node: Node) -> Result<(), RegistryError> {
-        // Verify that the node's required fields are non-empty
```
ltseeley, line 40: This change will require making the fields of the `Node` struct private; otherwise you could construct a `Node` directly without the builder that would violate these constraints.

libsplinter/src/registry/diesel/mod.rs
agunde406: Added more info to the commit message.

libsplinter/src/registry/diesel/operations/add_node.rs
agunde406: Added a new commit to change the fields to private

ltseeley (dismissed):

ltseeley (approved):

peterschwarz (approved):

PR #1171, opened 2021-02-04
RBAC CLI Authorized Identity commands

peterschwarz:
This PR adds the `role` sub-commands `show`, `create`, `update`, `delete` to the CLI.

# Testing

### Build and configure your environment:

```
$ cargo build --manifest-path cli/Cargo.toml --features experimental
$ cargo build --manifest-path splinterd/Cargo.toml --features experimental
$ export PATH=target/debug:$PATH
$ export SPLINTER_HOME=/tmp/splinter 
```

These operations for setup can be done in one terminal.

First, initialize your splinter instance:

```
$ splinter cert generate
$ splinter database migrate
```

Also, generate a key and add it to the allow_keys file:
```
$ splinter keygen
$ cat ~/.cylinder/keys/$USER.pub >> /tmp/splinter/data/allow_keys
```

Start your splinter node:
```
$ splinterd --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080
```

## Create Roles and authorized identities

These remainder of the operations in these test instructions can be done in a second terminal (once again, export your path from above). 

Add two role:

```
$  splinter role create \
    --display "Status Reader" \
    --permission "status.read" \
     status_reader
$  splinter role create \
    --display "Circuit Reader" \
    --permission "circuit.read" \
     circuit_reader
```

Show the currently authorized identities:

```
$ splinter authid list
IDENTITY                                                           TYPE ROLES
```

There are currently none, so lets create some:

```
$ splinter keygen -q test_user
$ splinter authid create --id-key $(cat ~/.cylinder/keys/test_user.pub) --role status_reader
```

Now list the keys again:
```
$ splinter authid list
IDENTITY                                                           TYPE ROLES
0261a18614bc108230310bd34f8c90ea9a65477b57feb292425f343dff279ec84c key  1
```

Check the authorized ID's details:

```
$ splinter authid show --id-key $(cat ~/.cylinder/keys/test_user.pub)
ID: 0261a18614bc108230310bd34f8c90ea9a65477b57feb292425f343dff279ec84c
    Type: key
    Roles:
        status_reader
```

Let's add the `circuit_reader` role, and show it again to verify the change

```
$ splinter authid update --id-key $(cat ~/.cylinder/keys/test_user.pub) --add-role circuit_reader
$ splinter authid show --id-key $(cat ~/.cylinder/keys/test_user.pub)
ID: 0261a18614bc108230310bd34f8c90ea9a65477b57feb292425f343dff279ec84c
    Type: key
[...]

PR #1184, opened 2021-02-11
Convert identity_type DB enum type

peterschwarz:
This change converts the data type on the columns for `identity_type` to an enum value.  In SQLite, this is a string with a check. In PostgreSQL, this is an actual enum type.  Both of these require a bit of tricky transformations when dealing with custom types, but the result is both a user-friendly query-able value (a readable string in the column vs an int) and strongly-typed values to and from the database.

# Testing

## Start a postgres instance:

```
docker run --name postgres -p 5432:5432 -e POSTGRES_PASSWORD=mypassword -e POSTGRES_DB=splinter -d postgres
```

### Build and configure your environment:

```
$ cargo build --manifest-path cli/Cargo.toml --features experimental
$ cargo build --manifest-path splinterd/Cargo.toml --features experimental
$ export PATH=target/debug:$PATH
$ export SPLINTER_HOME=/tmp/splinter 
```

These operations for setup can be done in one terminal.

First, initialize your splinter instance:

```
$ splinter cert generate
$ splinter database migrate -C postgres://postgres:mypassword@localhost:5432/splinter
```

Also, generate a key and add it to the allow_keys file:
```
$ splinter keygen
```

Start your splinter node:
```
$ splinterd --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080 \
    --database postgres://postgres:mypassword@localhost:5432/splinter
```

Follow the remainder of the test instructions from #1171.

### DIFF
--- a/libsplinter/src/migrations/diesel/sqlite/migrations/2021-02-09-042200_update-identity-enum/up.sql
+++ b/libsplinter/src/migrations/diesel/sqlite/migrations/2021-02-09-042200_update-identity-enum/up.sql
@@ -1,0 +1,32 @@
+-- Copyright 2018-2021 Cargill Incorporated
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+-- -----------------------------------------------------------------------------
+
+ALTER TABLE identities RENAME TO _identities_old;
+
+CREATE TABLE identities (
+    identity      TEXT PRIMARY KEY,
+    identity_type TEXT CHECK( identity_type IN ('key','user') ) NOT NULL
+);
+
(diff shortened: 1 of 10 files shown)


libsplinter/src/migrations/diesel/sqlite/migrations/2021-02-09-042200_update-identity-enum/up.sql
```diff
@@ -0,0 +1,31 @@
...
+CREATE TABLE identities (
+    identity      TEXT PRIMARY KEY,
+    identity_type TEXT CHECK( identity_type IN ('key','user') ) NOT NULL
+);
+
+INSERT INTO identities (identity, identity_type)
+  SELECT identity, (case
+                      when 1 then 'key'
+                      when 2 then 'user'
+                      else 'unknown'
```
agunde406, line 27: If "unknown" was ever set what would happen?

ltseeley (approved):
LGTM

shannynalayna (approved):

libsplinter/src/migrations/diesel/sqlite/migrations/2021-02-09-042200_update-identity-enum/up.sql
peterschwarz: Given this code isn't released, I'm not sure that it would matter, but the migration will fail.  I think I will change this to default to 'user', as that currently can be an arbitrary value

agunde406 (approved):

ltseeley (approved):

shannynalayna (approved):

PR #1188, opened 2021-02-12
Implement `circuit-abandon` in the admin service

shannynalayna:
Adds the implementation of the `circuit-abandon` feature. Mainly, this enables a circuit member to leave a circuit. This will "turn off" the circuit from that node's perspective and update the circuit status of the circuit to `Abandoned` for that node. Other nodes apart of the circuit will receive an `ABANDONED_CIRCUIT` admin message when this happens but will individually also need to abandon the circuit to deactivate it for their node. 

This PR adds the feature to libsplinter, which will be fully testable once the CLI commands are up in a PR (following this PR). For now, you can run `cargo test --features experimental abandon` to run through the unit tests for the implementation.


libsplinter/src/admin/service/shared.rs
```diff
@@ -1017,6 +1023,94 @@ impl AdminServiceShared {
...
+                    circuit_id
+                )),
+            )));
+        }
+        let (abandoned_proto_circuit, abandoned_store_circuit) = self
+            .make_abandoned_circuit(&stored_circuit)
+            .map_err(|err| ServiceError::UnableToHandleMessage(Box::new(err)))?;
+        // Updating the corresponding `active` circuit from the admin store
+        // and then removing the corresponding `CircuitProposal` from the
+        // disband request
```
agunde406, line 1079: disband -> abandon

libsplinter/src/admin/service/shared.rs
```diff
@@ -1017,6 +1023,94 @@ impl AdminServiceShared {
...
+            .get_circuit(circuit_id)
+            .map_err(|err| {
+                ServiceError::UnableToHandleMessage(Box::new(AdminSharedError::SplinterStateError(
+                    format!("error occurred when trying to get circuit {}", err),
+                )))
+            })?
+            .ok_or_else(|| {
+                ServiceError::UnableToHandleMessage(Box::new(AdminSharedError::SplinterStateError(
+                    format!(
+                        "Received purged request for a circuit that does not exist: circuit id {}",
```
agunde406, line 1040: purged -> abandon

libsplinter/src/admin/service/shared.rs
```diff
@@ -1017,6 +1023,94 @@ impl AdminServiceShared {
...
+                ServiceError::UnableToHandleMessage(Box::new(MarshallingError::ProtobufError(err)))
+            })?;
+            for member in stored_circuit.members().iter() {
+                if member != &self.node_id {
+                    network_sender.send(&admin_service_id(member), &envelope_bytes)?;
+                }
+            }
+        }
+
+        if stored_circuit.circuit_status() != &StoreCircuitStatus::Active {
```
agunde406, line 1066: This check should be before you send the messages to the other admin services

agunde406: Actually does this check even need to happen? here it should have already been validated in validate_abandon_circuit, also the error message is wrong. It should say "Attempting to abandon circuit that is not active"

agunde406 (changes requested):

libsplinter/src/admin/service/shared.rs
shannynalayna: Looking again, no this check doesn't need to happen again since it should all be validated within the `validate_abandon_circuit` method

libsplinter/src/admin/service/shared.rs
shannynalayna: Updated

libsplinter/src/admin/service/shared.rs
shannynalayna: Fixed the second half of this comment to say that it's updating the " `active` circuit in the admin store to have an `Abandoned` `circuit_status`. "

libsplinter/src/admin/service/shared.rs
shannynalayna: At this point it should be validated, so I removed that check and kept that message where it is

libsplinter/src/admin/service/shared.rs
```diff
@@ -2409,6 +2511,92 @@ impl AdminServiceShared {
...
+            .is_permitted(signer_public_key, PROPOSER_ROLE)
+            .map_err(|_| {
+                AdminSharedError::ValidationFailed(format!(
+                    "{} is not permitted to propose change for node {}",
+                    to_hex(signer_public_key),
+                    requester_node_id
+                ))
+            })?;
+
+        // Verifying the circuit is `Disbanded` and able to be purged
```
agunde406, line 2565: Comment is wrong, should say "Verifying the circuit is `Active` and able to be abandoned"

agunde406 (changes requested):

libsplinter/src/admin/service/shared.rs

agunde406 (dismissed):

libsplinter/src/admin/service/shared.rs
```diff
@@ -5830,6 +6101,473 @@ mod tests {
...
+            )
+            .expect("unable to add circuit to store");
+
+        if let Ok(()) = admin_shared.validate_abandon_circuit(
+            "01234-ABCDE",
+            PUB_KEY,
+            "node_a",
+            ADMIN_SERVICE_PROTOCOL_VERSION,
+        ) {
+            panic!("Should have been invalid because the protocol version is invalid");
```
isabeltomb, line 6483: based on the description of the test this message should probably be something like 
`"Should have been invalid because the requester is not permitted for the admin service"`

isabeltomb (approved):

agunde406 (approved):

PR #1190, opened 2021-02-15
Add AdminService builder

peterschwarz:
This change replaces the `AdminService::new` constructor with a builder, `AdminServiceBuilder`.

In order to cleanly do this, it moves the start of a thread from the constructor to the `Service::start` implementation.  This thread-start was in the constructor to avoid a race condition where notifications would be lost if they can before the admin service subscribed. The fix here is to queue the notifications that are collected by the subscriber map, such that they are available to the first subscriber.  This makes it so order doesn't matter, at least as far as the first subscriber is concerned.

The original constructor is deprecated.

### DIFF
--- a/libsplinter/src/admin/service/builder.rs
+++ b/libsplinter/src/admin/service/builder.rs
@@ -1,0 +1,244 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Builder for the AdminService
+
+#[cfg(feature = "service-arg-validation")]
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use cylinder::Verifier as SignatureVerifier;
+
+use crate::admin::store::AdminServiceStore;
+use crate::circuit::routing::RoutingTableWriter;
+use crate::error::InvalidStateError;
+use crate::keys::KeyPermissionManager;
+use crate::orchestrator::ServiceOrchestrator;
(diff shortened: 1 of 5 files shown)


libsplinter/src/admin/service/builder.rs
```diff
@@ -0,0 +1,232 @@
...
+    ) -> Self {
+        self.routing_table_writer = Some(routing_table_writer);
+        self
+    }
+
+    /// Constructs the AdminServce.
+    ///
+    /// # Errors
+    ///
+    /// Returna an [InvalidStateError] if any required properties are missing.
```
shannynalayna, line 149: s/Returna/Returns

libsplinter/src/peer/notification.rs
```diff
@@ -163,4 +198,140 @@ pub mod tests {
...
+            }
+        );
+    }
+
+    /// Test that the subscriber map obeys its queue limit for messages, until there is at least
+    /// one subscriber.
+    ///
+    /// Procedure:
+    ///
+    /// 1. Create a SubscriberMap with a with limit of 1
```
shannynalayna, line 281: Duplicate 'with'

libsplinter/src/admin/service/builder.rs
```diff
@@ -0,0 +1,232 @@
...
+        })?;
+
+        let routing_table_writer = self.routing_table_writer.ok_or_else(|| {
+            InvalidStateError::with_message(
+                "An admin service requires an routing_table_writer".into(),
+            )
+        })?;
+
+        #[cfg(feature = "admin-service-event-store")]
+        let admin_event_store = admin_store.clone_boxed();
```
agunde406, line 203: The event store needs to be able to be a different type in the case that the admin store is set to YAML. The builder should have a separate with_admin_event_store function until we make YamlAdminServiceStore read only.
This will require updating the depreacted new method and the daemon to still pass it in.

agunde406 (changes requested):

libsplinter/src/admin/service/builder.rs
peterschwarz: Fixed

libsplinter/src/peer/notification.rs
peterschwarz: Fixed

libsplinter/src/admin/service/builder.rs
peterschwarz: Fixed

agunde406 (approved):

shannynalayna (approved):

PR #1191, opened 2021-02-16
Separate Service Purge from Destory

peterschwarz:
This change separates service::destroy from the purge operation.  These are distinct operations: one is an object life-cycle operation, the other is specific operation for cleaning up persistent state.

This split fixes an issue where destroy was called in the `shutdown_all_services` method of the `Orchestrator`, resulting in deleting scabbard data and, at node restart, making the services brand new instances, including no contracts.

# Test

Build gameroom with stable features and start it as normal.  Go through the process of creating a gameroom.

Stop and start the acme node:

```
$ docker-compose -f examples/gameroom/docker-compose.yaml stop splinterd-node-acme
$ docker-compose -f examples/gameroom/docker-compose.yaml start splinterd-node-acme
```

Before this fix: the gameroom would not be usable, with an error message of 

```
splinterd-node-acme     | [2021-02-15 19:54:54.767] T["consensus-a001"] ERROR [splinter::consensus::two_phase] Error while creating proposal: proposal manager error occurred: scabbard state error: transaction failed: "Contract does not exist: xo, 0.3.3"
```

After this fix: the gameroom can still be used to create games or make moves.

### DIFF
--- a/libsplinter/src/admin/service/mod.rs
+++ b/libsplinter/src/admin/service/mod.rs
@@ -622,6 +622,10 @@ impl Service for AdminService {
         }
     }
 
+    fn purge(&mut self) -> Result<(), crate::error::InternalError> {
+        Ok(())
+    }
+
     fn handle_message(
         &self,
         message_bytes: &[u8],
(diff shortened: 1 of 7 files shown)


shannynalayna (approved):
Tested restarting and was able to use the gameroom, also tested purging after a restart and that also worked

agunde406 (approved):
tested

PR #1193, opened 2021-02-17
Remove splinterd compile err for no DB backend

ltseeley:
Removes the compile error in splinterd that would occur when no database
backend features are enabled. This prevented building splinterd with no
default features. Now, if no database backend is enabled, DB URLs will
always be invalid.


vaporos (approved):
Tested that 'just build' worked.

peterschwarz (approved):

PR #1196, opened 2021-02-18
Update ServiceProcessor, ServiceOrchestrator to have builders, shutdown handles

peterschwarz:
This change implements the new ShutdownHandle crate for the ServiceProcessor, replacing its previous combination of handle and JoinHandles, when the "shutdown" feature is enabled.

Likewise it makes several changes to the daemon code to handle this experimental feature:

- Removes the need to send handles to the ctrl-c handler
- Removes the separate threads used to startup both the admin and health service processors
- Reworks the health service feature to include the requirement on the "shutdown" feature, as both are experimental. 
- Removes the need for an external atomic bool for shutdown, as this could put the processor into an inconsistent state if one is triggered, but the other isn't.

Likewise, it implements ShutdownHandle for ServiceOrchestrator, but in doing so, it splits the orchestrator into an builder, a runnable (not started) and a started instance.

### DIFF
--- a/libsplinter/src/service/mod.rs
+++ b/libsplinter/src/service/mod.rs
@@ -53,6 +53,7 @@ pub use factory::ServiceFactory;
 pub use processor::registry::StandardServiceNetworkRegistry;
 pub use processor::JoinHandles;
 pub use processor::ServiceProcessor;
+#[cfg(not(feature = "shutdown"))]
 pub use processor::ShutdownHandle;
 
 pub use error::{
--- a/libsplinter/src/service/mod.rs
+++ b/libsplinter/src/service/mod.rs
@@ -53,6 +53,7 @@ pub use factory::ServiceFactory;
 pub use processor::registry::StandardServiceNetworkRegistry;
 pub use processor::JoinHandles;
 pub use processor::ServiceProcessor;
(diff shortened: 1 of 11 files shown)


ltseeley (dismissed):

libsplinter/src/service/mod.rs
```diff
@@ -53,6 +53,7 @@ pub use factory::ServiceFactory;
 pub use processor::registry::StandardServiceNetworkRegistry;
 pub use processor::JoinHandles;
 pub use processor::ServiceProcessor;
+#[cfg(not(feature = "shutdown"))]
```
agunde406, line 57: typo in commit message splitnerd -> splinterd

libsplinter/src/service/processor/mod.rs
```diff
@@ -53,7 +53,6 @@ const TIMEOUT_SEC: u64 = 2;
 /// Includes the service senders and join_handles for the service threads.
```
agunde406, line 53: typo in commit message: shudown -> shutdown

agunde406 (changes requested):

vaporos (changes requested):

vaporos (approved):

agunde406 (approved):

PR #1197, opened 2021-02-18
Add type for `requester` in `AdminServiceEvent`

shannynalayna:
Adds a type, `PublicKey`, to represent the `requester` field in an
`AdminServiceEvent`. Previously, this was just `Vec<u8>`, so adding the
`PublicKey` type around this field makes the field's purpose more
obvious.

### DIFF
--- a/libsplinter/src/admin/store/event.rs
+++ b/libsplinter/src/admin/store/event.rs
@@ -20,6 +20,9 @@ use super::CircuitProposal;
 use crate::admin::service::messages;
 use crate::error::InvalidStateError;
 
+/// Represents the `requester`'s public key associated with an `AdminServiceEvent`
+pub type PublicKey = Vec<u8>;
+
 #[derive(Debug, PartialEq, Eq, Clone)]
 /// Representation of an `AdminServiceEvent` defined by the admin messages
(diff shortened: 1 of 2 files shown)


libsplinter/src/admin/store/event.rs
```diff
@@ -32,9 +32,9 @@ pub struct AdminServiceEvent {
 /// Native representation of the `AdminServiceEvent` enum variants
 pub enum EventType {
     ProposalSubmitted,
-    ProposalVote { requester: Vec<u8> },
-    ProposalAccepted { requester: Vec<u8> },
-    ProposalRejected { requester: Vec<u8> },
+    ProposalVote { requester: messages::PublicKey },
```
peterschwarz, line 35: This should also be a type def here, not the messages value, since this is the native struct.

peterschwarz (changes requested):

libsplinter/src/admin/store/event.rs
shannynalayna: Updated to add a native type definition for this struct

peterschwarz (approved):

agunde406 (approved):

PR #1200, opened 2021-02-19
Require explicitly enabling biome credentials for splinterd

ltseeley:
Adds the `--enable-biome-credentials` flag for splinterd. This change requires breaking up the Biome REST API into providers for each of the main components (credentials, key management, and profile). Some more work will need to be done in the future to cleanly separate all functionality, but this is an incremental improvement.

A future PR will add a `PATCH /biome/keys` endpoint for updating a users' set of encrypted keys when Biome credentials are not being used.

# Testing

## Setup

1. Build the CLI and splinterd, and configure your environment:

```bash
$ cargo build --manifest-path cli/Cargo.toml --features experimental
$ cargo build --manifest-path splinterd/Cargo.toml --features experimental
```

2. Configure environment variables:

```bash
$ export PATH=target/debug:$PATH
$ export SPLINTER_HOME=/tmp/splinter
```

3. Create certs and migrate the DB:

```bash
$ splinter cert generate
$ splinter database migrate
```

## Verifying Biome Credentials

1. Start splinterd with Biome credentials enabled:

```bash
$ splinterd --rest-api-endpoint http://localhost:8080 --tls-insecure --enable-biome-credentials -vv
```

2. In a new terminal, register and login a new user with Biome credentials:

```bash
$ curl -X POST \
  -H "Content-Type: application/json" \
  -d '{"username": "test", "hashed_password": "123"}' \
  http://localhost:8080/biome/register
$ curl -X POST \
  -H "Content-Type: application/json" \
  -d '{"username": "test", "hashed_password": "123"}' \
  http://localhost:8080/biome/login
```

3. Use the token returned in the last command to get the user's keys (should be empty):

```bash
$ curl -H "Authorization: Bearer <token>" http://localhost:8080/biome/keys
```

4. Shutdown splinterd

## Verifying Key Management Without Biome Credentials

1. Start splinterd WITHOUT Biome credentials enabled, but with OAuth enabled (this example uses Google OAuth):

```bash
$ splinterd --rest-api-endpoint http://localhost:8080 \
  --oauth-provider google \
  --oauth-client-id <client-id> \
  --oauth-client-secret <client-secret> \
  --tls-insecure -vv
```

2. In a web browser, navigate to http://localhost:8080/oauth/login?redirect_url=test and login with OAuth. This will result in a 404, but the access token will be displayed in the query string in the browser's URL bar.

3. Using the access token from the last step, get the list of keys for your user:

```bash
$ curl -H "Authorization: Bearer <token>" http://localhost:8080/biome/keys
```


libsplinter/src/biome/key_management/mod.rs
```diff
@@ -14,6 +14,8 @@
```
shannynalayna, line 14: Commit message mentions the `/biome/profile` route, seems to be copy pasta from previous commit

libsplinter/src/biome/credentials/rest_api/actix_web_1/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    refresh_token_store: Option<Arc<dyn RefreshTokenStore>>,
+    credentials_store: Option<Arc<dyn CredentialsStore>>,
+}
+
+impl BiomeCredentialsRestResourceProviderBuilder {
+    /// Sets a KeyStore for the BiomeCredentialsRestResourceProvider
+    ///
+    /// # Arguments
+    ///
+    /// * `pool`: ConnectionPool to database that will serve as backend for KeyStore
```
shannynalayna, line 142: The arg in this comment doesn't match the function signature

libsplinter/src/biome/credentials/rest_api/actix_web_1/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        self.token_secret_manager = Some(Arc::new(secret_manager));
+        self
+    }
+
+    /// Sets a SecretManager for the refresh tokens for the BiomeCredentialsRestResourceProvider
+    ///
+    /// # Arguments
+    ///
+    /// * `secret_manager`: the SecretManager to be used for fetching and generating secrets to
+    ///   sign and verify JWT tokens
```
shannynalayna, line 197: Comment here looks like copypasta from the above function

shannynalayna (commented):
Also this error comes up running `just lint`: 

```
error[E0599]: no method named `with_enable_biome_credentials` found for struct `config::partial::PartialConfig` in the current scope
  --> splinterd/src/config/default.rs:81:14
   |
81 |             .with_enable_biome_credentials(Some(false));
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `config::partial::PartialConfig`
   | 
  ::: splinterd/src/config/partial.rs:34:1
   |
34 | pub struct PartialConfig {
   | ------------------------ method `with_enable_biome_credentials` not found for this

error: aborting due to previous error
```

libsplinter/src/biome/key_management/mod.rs
ltseeley: Fixed

libsplinter/src/biome/credentials/rest_api/actix_web_1/mod.rs
ltseeley: Fixed

libsplinter/src/biome/credentials/rest_api/actix_web_1/mod.rs
ltseeley: Fixed

shannynalayna (approved):
Tested

peterschwarz (approved):

PR #1207, opened 2021-02-25
Remove Serialize/Deserialize from Node struct

agunde406:
The struct itself should be derive Serialize/Deserialize
itself. Instead, the different components should implement
their own structs to handler serialization.

To remove Serialize/Deserialize from Node, a YamlNode for
the yaml implementation had to be added, as well as a
NewNode struct to be used by the REST API add/update
routes.

### DIFF
--- a/cli/src/action/registry/mod.rs
+++ b/cli/src/action/registry/mod.rs
@@ -16,7 +16,7 @@
 mod api;
 
 use clap::ArgMatches;
-use splinter::registry::Node;
+use splinter::registry::{Node, YamlNode};
 #[cfg(feature = "registry")]
 use std::collections::HashMap;
 use std::fs::File;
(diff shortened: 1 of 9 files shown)


cli/src/action/registry/mod.rs
```diff
@@ -16,7 +16,7 @@
 mod api;
 
 use clap::ArgMatches;
-use splinter::registry::Node;
+use splinter::registry::{Node, YamlNode};
```
ltseeley, line 19: Typos in commit message: s/should be/shouldn't/ and s/handler/handle/

agunde406: Fixed

ltseeley (approved):

ltseeley (approved):

peterschwarz (approved):

PR #1210, opened 2021-02-25
Fix scabbard hash mismatch bug

ltseeley:
This PR fixes a bug that cause hash mismatches to occur when a sufficiently high rate of batches was submitted across a network of scabbard services.


vaporos:
We should consider backward compatibility options before this is merged.

ltseeley:
Going to break this up into smaller PRs

PR #1213, opened 2021-03-01
Scabbard v2

ltseeley:
Adds versioning to the scabbard service and creates a new version 2 of scabbard that uses version 2 of the two-phase commit consensus algorithm. This fixes a bug where hash mismatch errors occurred because multiple proposals were created at the same height.

# Testing

1. Build and run gameroom:

```bash
$ docker-compose -f examples/gameroom/docker-compose.yaml build
$ docker-compose -f examples/gameroom/docker-compose.yaml up
```

2. In a new terminal window, get alice and bob's keys and save them in your current directory:

```bash
$ docker-compose -f examples/gameroom/docker-compose.yaml run generate-registry bash
# cat /registry/alice.priv
<alice's private key>
# cat /registry/alice.pub
<alice's public key>
# cat /registry/bob.priv
<bob's private key>
# cat /registry/bob.pub
<bob's public key>
# exit
$ echo "<alice's private key>" > alice.priv
$ echo "<alice's public key>" > alice.pub
$ echo "<bob's private key>" > bob.priv
$ echo "<bob's public key>" > bob.pub
```

3. Create a new circuit:

```bash
$ export SPLINTER_CIRCUIT_TEMPLATE_PATH=./cli/packaging/
$ splinter circuit propose \
  --node acme-node-000::tcps://splinterd-node-acme:8044 \
  --node bubba-node-000::tcps://splinterd-node-bubba:8044 \
  --template scabbard_circuit_template \
  --template-arg SIGNER_PUB_KEY=$(cat alice.pub) \
  --url http://0.0.0.0:8088 \
  -k ./alice.priv \
  --management test
$ export CIRCUIT_ID=<circuit ID>
$ splinter circuit vote $CIRCUIT_ID --accept --url http://0.0.0.0:8089 -k ./bob.priv
```

4. Download [smallbank_0.0.1.tar.gz](https://github.com/Cargill/splinter/files/6065243/smallbank_0.0.1.tar.gz), change the extension to `.scar`, and move the file to your current directory.

5. Setup smallbank for the new circuit:

```bash
$ scabbard cr create smallbank \
  --owners $(cat alice.pub) \
  --key ./alice.priv \
  --url http://0.0.0.0:8088 \
  --service-id $CIRCUIT_ID::a000
$ scabbard contract upload smallbank:0.0.1 \
  --path . \
  --key ./alice.priv \
  --url http://0.0.0.0:8088 \
 --service-id $CIRCUIT_ID::a000
$ scabbard  ns create 332514 \
    --owners $(cat alice.pub) \
    --key ./alice.priv \
    --url http://0.0.0.0:8088 \
    --service-id $CIRCUIT_ID::a000
$ scabbard perm 332514 smallbank --read --write \
  --key ./alice.priv \
  --url http://0.0.0.0:8088 \
  --service-id $CIRCUIT_ID::a000
$ scabbard contract list -U 'http://0.0.0.0:8088' \
  --service-id $CIRCUIT_ID::a000 --key ./alice.priv
```

6. Using the transact CLI, run a smallbank workload targetting both nodes:

```bash
$ transact workload \
  --target-rate 1-5 \
  --key ./alice.priv \
  --workload smallbank \
  --update 5 \
  --smallbank-num-accounts 5 \
  --targets "http://0.0.0.0:8089/scabbard/$CIRCUIT_ID/a001;http://0.0.0.0:8088/scabbard/$CIRCUIT_ID/a000"
```

Verify that no hash mismatch errors occur

### DIFF
--- a/services/scabbard/libscabbard/src/service/consensus.rs
+++ b/services/scabbard/libscabbard/src/service/consensus.rs
@@ -21,7 +21,8 @@ use std::time::Duration;
 use protobuf::Message;
 use splinter::consensus::{
     error::{ConsensusSendError, ProposalManagerError},
-    two_phase::v1::TwoPhaseEngine,
+    two_phase::v1::TwoPhaseEngine as TwoPhaseEngineV1,
+    two_phase::v2::TwoPhaseEngine as TwoPhaseEngineV2,
     ConsensusEngine, ConsensusMessage, ConsensusNetworkSender, PeerId, Proposal, ProposalId,
     ProposalManager, ProposalUpdate, StartupState,
 };
@@ -32,6 +33,7 @@ use crate::protos::scabbard::{ProposedBatch, ScabbardMessage, ScabbardMessage_Ty
 use super::error::{ScabbardConsensusManagerError, ScabbardError};
 use super::shared::ScabbardShared;
 use super::state::ScabbardState;
+use super::ScabbardVersion;
 
 /// Component used by the service to manage and interact with consenus
 pub struct ScabbardConsensusManager {
@@ -45,6 +47,7 @@ impl ScabbardConsensusManager {
     /// consensus, and start consensus in a separate thread.
     pub fn new(
         service_id: String,
+        version: ScabbardVersion,
         shared: Arc<Mutex<ScabbardShared>>,
         state: Arc<Mutex<ScabbardState>>,
         // The coordinator timeout for the two-phase commit consensus engine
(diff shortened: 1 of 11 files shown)


agunde406 (dismissed):
tested

services/scabbard/libscabbard/src/service/consensus.rs
```diff
@@ -413,6 +413,7 @@ mod tests {
             VecDeque::new(),
             Some(Box::new(service_sender.clone())),
             peer_services.clone(),
+            "0".to_string(),
```
peterschwarz, line 416: What does this `"0"` mean? If it is a service id, it would be nice to use a string that indicates as much, e.g., "svc0" (to conform to the length requirement)

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -196,3 +230,70 @@ impl ScabbardShared {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use cylinder::{secp256k1::Secp256k1Context, VerifierFactory};
+    use splinter::service::{ServiceMessageContext, ServiceSendError};
+
+    /// Verifies that the `is_coordinator` and `coordinator_service_id` methods work properly.
+    #[test]
+    fn coordinator() {
```
peterschwarz, line 243: The test isn't clear as to why the asserts are correct.

services/scabbard/libscabbard/src/service/mod.rs
```diff
@@ -393,6 +421,30 @@ impl Service for Scabbard {
                     ))
                     .map_err(|err| ServiceError::UnableToHandleMessage(Box::new(err)))
             }
+            ScabbardMessage_Type::NEW_BATCH => {
+                match self.version {
+                    ScabbardVersion::V1 => {
```
peterschwarz, line 426: Is there any point where they negotiate their versions, or would these be the point at which an admin would see that the services have a version mismatch?

peterschwarz (changes requested):

services/scabbard/libscabbard/src/service/consensus.rs
ltseeley: Updated this test to use "svc_" service IDs.

services/scabbard/libscabbard/src/service/shared.rs
ltseeley: Added more details to the test's doc comment

services/scabbard/libscabbard/src/service/mod.rs
ltseeley: No, the version is determined by the service arguments provided in the circuit definition. It's up to the circuit creator to ensure they're the same, which is handled automatically by the scabbard circuit template.

peterschwarz: It would be good if we could verify that sort of thing at circuit creation time, but that is probably beyond the scope of this PR.

agunde406 (approved):

peterschwarz (approved):

PR #1216, opened 2021-03-02
Add initial circuit creation integration tests

shannynalayna:
Adds some helper functions as well as four initial circuit creation tests between 2- and 3-nodes. This includes tests for when a circuit is proposed then accepted, and when rejected. Currently these tests are ignored and will fail if run, but should compile.

### DIFF
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -49,6 +49,9 @@ serde_derive = "1.0.80"
 tempdir = "0.3"
 toml = "0.5"
 
+[dev-dependencies]
+openssl = { version = "0.10" }
+
 [dependencies.scabbard]
 path = "../services/scabbard/libscabbard"
 features = [
--- a/splinterd/tests/admin/circuit_lifecycle.rs
+++ b/splinterd/tests/admin/circuit_lifecycle.rs
@@ -1,0 +1,782 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Integration tests for creating a circuit between multiple nodes.
+
+use std::collections::HashMap;
+use std::time::{Duration, Instant};
+
+use cylinder::{secp256k1::Secp256k1Context, Context, Signer};
+use openssl::hash::{hash, MessageDigest};
+use protobuf::Message;
+
+use crate::framework::network::Network;
+use splinter::admin::client::{AdminServiceClient, ProposalSlice};
+use splinter::admin::messages::{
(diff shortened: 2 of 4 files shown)


splinterd/Cargo.toml
```diff
@@ -41,7 +41,7 @@ cylinder = "0.2"
 flexi_logger = "0.14"
 health = { path = "../services/health", optional = true }
 log = "0.4"
-openssl = { version = "0.10", optional = true }
```
peterschwarz, line 44: This might be better suited to be additionally added as a dev dependency or, there should be a feature that depends on this used by the tests.

shannynalayna: Updated to add `open-ssl` as a dev-dependency

agunde406 (changes requested):

splinterd/tests/admin/circuit_list.rs
```diff
@@ -37,6 +37,23 @@ fn single_node_network(rest_api_variant: RestApiVariant) {
...
 
+/// Creates a multi-node network and confirms that the admin service's REST API is available
+/// by listing circuits (which will be empty).
+fn multi_node_network(rest_api_variant: RestApiVariant) {
+    let mut network = Network::new()
+        .with_default_rest_api_variant(rest_api_variant)
+        .add_nodes_with_defaults(3)
+        .unwrap();
+
+    let client = network.node(0).unwrap().admin_service_client();
```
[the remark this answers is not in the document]
shannynalayna: I updated this test to check the `list_circuits` endpoint of all three nodes in the network.

splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -0,0 +1,665 @@
...
+        "node_b".to_string(),
+        format!("http://localhost:{}", node_b.rest_api_port()),
+    );
+    let circuit_id = "ABCDE-01234";
+    // Create the `CircuitManagementPayload` to be sent to a node
+    let circuit_payload_bytes =
+        make_create_circuit_payload(&circuit_id, "node_a", node_info.clone(), node_a_signer);
+    // Submit the `CircuitManagementPayload` to the first node
+    let res = node_a_client.submit_admin_payload(circuit_payload_bytes);
+    println!("{:?}", res);
```
agunde406, line 221: println should be removed

splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -0,0 +1,665 @@
...
+    assert!(res.is_ok());
+
+    // Wait for the proposal to be committed for the second node
+    let proposal_b;
+    let start = Instant::now();
+    loop {
+        if Instant::now().duration_since(start) > Duration::from_secs(60) {
+            panic!("Failed to detect proposal in time");
+        }
+        let proposals = node_b_client.list_proposals(None, None).unwrap().data;
```
agunde406, line 231: All of the unwraps should be expects to give better idea of what happened if there is a failure.

agunde406 (changes requested):

splinterd/tests/admin/circuit_lifecycle.rs
shannynalayna: Updated most of the unwraps to expects to better explain (kept some unwraps where expect seemed unnecessary and added comments as to why the unwrap is used instead of expect)

splinterd/tests/admin/circuit_lifecycle.rs
shannynalayna: Fixed

agunde406 (dismissed):

agunde406 (approved):

isabeltomb (approved):

PR #1218, opened 2021-03-02
Remove `AdminServiceEventStore`-related features

shannynalayna:
Removes instances of the `admin-service-event-store` and `admin-service-event-store-postgres`, as well as code guarded with `not(feature = "admin-service-event-store")`.  Also makes a small fix that came up when running `just lint` after the feature was removed to add a feature guard for an import statement in the `add_event` module.


libsplinter/Cargo.toml
```diff
@@ -105,7 +105,6 @@ experimental = [
     # The following features are experimental:
     "admin-service-client",
     "admin-service-event-store",
```
vaporos, line 107: Did you intend to also remove admin-service-event-store?

shannynalayna: Yes, I tried to separate removing the two features and forgot this line when I updated the PR - fixed now

libsplinter/src/admin/rest_api/actix/ws_register_type.rs
```diff
@@ -317,24 +231,10 @@ struct JsonAdminEvent {
...
-            event: raw_evt.1,
-            event_id: None,
-        }
-    }
-}
-
 // Conversion for the `JsonAdminEvent` sent to subscribers when the `AdminServiceStore` is
 // used to store admin events.
-// `timestamp` is set to the `UNIX_EPOCH` value to allow for backward-compatibility, as the
+// `timestamp` is set to the current time to allow for backward-compatibility, as the
```
vaporos, line 236: Is this change intentional?

shannynalayna: Yes, because that comment was not valid before as the timestamp was being set to the current time (using `SystemTime::now()`) rather than the `UNIX_EPOCH` time value.

libsplinter/src/admin/store/diesel/operations/add_event.rs
```diff
@@ -21,13 +21,15 @@ use diesel::{dsl::insert_into, prelude::*};
 use super::AdminServiceStoreOperations;
 
 use crate::admin::service::messages;
+#[cfg(feature = "sqlite")]
```
vaporos, line 24: Should the two implementations here actually differ?

shannynalayna: They do, in how the event is initially added, then the event_id is retrieved. If you look at how the postgres impl is getting the `event_id`: https://github.com/Cargill/splinter/blob/main/libsplinter/src/admin/store/diesel/operations/add_event.rs#L63, versus the sqlite impl: https://github.com/Cargill/splinter/blob/main/libsplinter/src/admin/store/diesel/operations/add_event.rs#L155

shannynalayna: Also, generally the `add_*` methods can't use the generic connection, if you look at the other add operations implemented for the admin store. (Most have separate implementations for the postgres and sqlite feature, though the implementations don't differ greatly)

peterschwarz (approved):

vaporos (approved):

isabeltomb (approved):

agunde406 (approved):

ltseeley (approved):

RyanLassigBanks (approved):

dplumb94 (approved):

dnewh (approved):

rbuysse (approved):

PR #1219, opened 2021-03-03
Update ShutdownHandle.wait_for_shutdown to take Box<Self> 

agunde406:
Ideally, wait_for_shutdown would consume the object itself.
Before talking &mut self would result in the structs being
left in an invalid state. However, since ShutdownHandle
is a trait it must take a Box of itself instead.

The resulting implementation allows for consuming the struct
on wait_for_shutdown, but does require wrapping some structs
in a Box only to shutdown.


vaporos (approved):

agunde406:
Closing in favor of https://github.com/Cargill/splinter/pull/1222

PR #1224, opened 2021-03-04
Modify OpenIdProfileProvider for azure account profile photos

isabeltomb:
Modify the OpenIdProfileProvider to make a call to the microsoft graph api endpoint for retrieving profile pictures if Azure is being used as the OpenID provider for auth.

The microsoft graph api `/photo/$value` endpoint returns the binary data for the photo. This data is converted to a base64 string and set as the 'picture' field in the OpenIdProfileResponse struct.

### DIFF
--- a/libsplinter/src/oauth/profile/openid.rs
+++ b/libsplinter/src/oauth/profile/openid.rs
@@ -14,6 +14,7 @@
 
 //! A profile provider that looks up OpenID profile information
 
+use base64::encode;
 use reqwest::{blocking::Client, StatusCode};
 use serde::Deserialize;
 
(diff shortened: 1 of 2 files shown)


libsplinter/src/oauth/profile/openid.rs
```diff
@@ -55,10 +56,33 @@ impl ProfileProvider for OpenIdProfileProvider {
...
+        // microsoft graph api endpoint with the access token to retrieve the
+        // binary data for the authenticated user's profile photo
+        if self.userinfo_endpoint.contains("graph.microsoft.com") {
+            let picture_response = Client::builder()
+                .build()
+                .map_err(|err| InternalError::from_source(err.into()))?
+                .get("https://graph.microsoft.com/beta/me/photo/$value")
+                .header("Authorization", format!("Bearer {}", access_token))
+                .send()
+                .map_err(|err| InternalError::from_source(err.into()))?;
```
peterschwarz, line 73: I'm not sure we should propagate these failures if the picture can't be downloaded. Perhaps log a warning - it doesn't seem catastrophic enough to give up on the whole profile.

isabeltomb: Updated this to log a warning and set the picture field of the struct to an empty string if the microsoft graph api call fails. The empty string in the picture field will make the ui display the default profile picture icon.

libsplinter/src/oauth/profile/openid.rs
```diff
@@ -55,10 +56,38 @@ impl ProfileProvider for OpenIdProfileProvider {
...
+            let picture_response = match Client::builder()
+                .build()
+                .map_err(|err| InternalError::from_source(err.into()))?
+                .get("https://graph.microsoft.com/beta/me/photo/$value")
+                .header("Authorization", format!("Bearer {}", access_token))
+                .send()
+            {
+                Ok(image) => Some(encode(
+                    image
+                        .bytes()
```
peterschwarz, line 76: Probably would do the same with this error, too

isabeltomb: Fixed

peterschwarz (approved):

PR #1225, opened 2021-03-05
Move threading::shutdown into threading::lifecycle

agunde406:
This move provides a location for future threading
lifecycle traits to be added.


vaporos:
I think we should either pub use this as threading::lifecycle::ShutdownHandle or even define it in the lifecycle module. We have two other lifecycle traits in mind but no implementation stuff related to them (in this module) so it may not need submodules.

vaporos (approved):

peterschwarz (approved):

PR #1228, opened 2021-03-08
Add disband integration tests

shannynalayna:
This PR updates the current circuit-creation tests to grab the node ID from the node (as this is now implemented) rather than using a placeholder of `node_*`. This also updates how the node information was stored and handed to the helper functions to allow for the order of the nodes' information to be maintained. 

This PR also adds tests for the circuit-disband process, including scenarios between 2 and 3 nodes.

### DIFF
--- a/splinterd/tests/admin/circuit_lifecycle.rs
+++ b/splinterd/tests/admin/circuit_lifecycle.rs
@@ -522,7 +522,7 @@ pub fn test_2_party_circuit_creation() {
 
     commit_2_party_circuit(circuit_id, node_a, node_b);
 
-    shutdown!(network).unwrap();
+    shutdown!(network).expect("Unable to shutdown network");
 }
 
 /// Test that a 3-party circuit may be created on a 3-node network.
--- a/splinterd/tests/admin/circuit_lifecycle.rs
+++ b/splinterd/tests/admin/circuit_lifecycle.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Integration tests for creating a circuit between multiple nodes.
+//! Integration tests for the lifecycle of a circuit between multiple nodes.
 
 use std::collections::HashMap;
 use std::time::{Duration, Instant};
@@ -28,8 +28,8 @@ use splinter::admin::messages::{
     RouteType, SplinterNode, SplinterNodeBuilder, SplinterService, SplinterServiceBuilder, Vote,
 };
 use splinter::protos::admin::{
-    CircuitCreateRequest, CircuitManagementPayload, CircuitManagementPayload_Action,
-    CircuitManagementPayload_Header,
+    CircuitCreateRequest, CircuitDisbandRequest, CircuitManagementPayload,
+    CircuitManagementPayload_Action, CircuitManagementPayload_Header,
 };
 use splinterd::node::RestApiVariant;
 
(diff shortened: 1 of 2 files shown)


splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -38,7 +37,7 @@ use splinterd::node::RestApiVariant;
 fn make_create_circuit_payload(
     circuit_id: &str,
     requester: &str,
-    node_info: HashMap<String, String>,
+    node_info: &[(String, String)],
```
peterschwarz, line 40: If you changed this to a `&mut dyn Iterator<Item = (...)>`, you would have had to change far less code.

peterschwarz (commented):
This might be easier to do if we were to modify Network to have `node(&self, idx: usize) -> &Node` and `node_mut(&mut self, idx: usize) -> &mut Node`.  The tests mainly need references, and we could just pass a list of `&Node`s to the functions for creating circuits. It also will remove the need to do some clones (though, in tests this is less of a problem).

shannynalayna:
> This might be easier to do if we were to modify Network to have `node(&self, idx: usize) -> &Node` and `node_mut(&mut self, idx: usize) -> &mut Node`. The tests mainly need references, and we could just pass a list of `&Node`s to the functions for creating circuits. It also will remove the need to do some clones (though, in tests this is less of a problem).

Yeah, it probably would be. A lot of the reasoning for saving values was required based on the mutable ref to the nodes that is taken and then can't be used after another node is borrowed.

splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -455,20 +494,25 @@ pub fn test_2_party_circuit_creation() {
...
+    let mut node_info = Vec::new();
     // Get the node and node's client for the first node
-    let node_a = network.node(0).expect("Unable to get node_a");
+    let node_a = network.node(0).expect("Unable to get first node");
     let node_a_client = node_a.admin_service_client();
     let node_a_signer = node_a.admin_signer().clone_box();
+    // Store node information to use to build payloads
+    node_info.push((
+        node_a.node_id().to_string(),
+        format!("http://localhost:{}", node_a.rest_api_port()),
```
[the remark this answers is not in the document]
shannynalayna: This was initially saved in order to build the information needed by the circuit (to build the variable information in the circuit). This was used in place of having to pass the nodes into this function to retrieve the information as the test wouldn't be able to pull out multiple mutable references from the network. Making the change mentioned in your comment above will help solve this.

splinterd/tests/framework/network.rs
```diff
@@ -78,7 +78,7 @@ impl Network {
         self
     }
 
-    pub fn node(self: &mut Network, n: usize) -> Result<&mut Node, InvalidArgumentError> {
+    pub fn node_mut(self: &mut Network, n: usize) -> Result<&mut Node, InvalidArgumentError> {
```
peterschwarz, line 81: I think we can just change this method to not be `mut`, since it's currently not necessary (which would not require the commit that makes use of it just for the sake of using it). Easy enough to add back.

peterschwarz (changes requested):

splinterd/tests/framework/network.rs
shannynalayna: Updated to remove the `node_mut` method and the extra commit that used this method.

peterschwarz (approved):

agunde406 (approved):

PR #1231, opened 2021-03-09
Remove separate list for `disband` circuit proposals in admin service

shannynalayna:
This PR removes the separate list that held circuit proposals associated with a `CircuitDisbandRequest`. The circuit proposals associated with a disband request are now put in the `uninitialized_circuits` list, where other circuits that have been committed to state but must have their services handled. This also removes the separate methods used to handle the votes for a proposal associated with a `disband` request to verify whether the services must be initialized or stopped based on the associated circuit proposal.

This requires checking if the associated proposal is in state (for the remote nodes receiving the request) or finding the proposal stored in the `uninitialized_circuits` list if the circuit has already been committed to state but the services have yet to be stopped (a scenario encountered by the requesting node).

### DIFF
--- a/libsplinter/protos/admin.proto
+++ b/libsplinter/protos/admin.proto
@@ -310,8 +310,7 @@ message AdminMessage {
         CONSENSUS_MESSAGE = 1;
         PROPOSED_CIRCUIT = 2;
         MEMBER_READY = 3;
-        DISBANDED_CIRCUIT = 4;
-        ABANDONED_CIRCUIT = 5;
+        ABANDONED_CIRCUIT = 4;
 
         SERVICE_PROTOCOL_VERSION_REQUEST = 100;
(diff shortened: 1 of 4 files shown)


libsplinter/protos/admin.proto
```diff
@@ -310,8 +310,7 @@ message AdminMessage {
         CONSENSUS_MESSAGE = 1;
         PROPOSED_CIRCUIT = 2;
         MEMBER_READY = 3;
-        DISBANDED_CIRCUIT = 4;
```
peterschwarz, line 313: This commit seems like two very different changes.  They should probably be split up.

shannynalayna: Updated the change to the admin.proto file into a separate commit

agunde406 (approved):
tested

peterschwarz (approved):

PR #1236, opened 2021-03-11
Implement NetworkSubsystem for Nodes used in the integrations tests

agunde406:
The NetworkSubsystem contains the configuration for all of the network components of a Node including
the ConnectionManager, PeerManager, Dispatcher Handlers and network listeners.

The PR includes a bug fix for MemoryStorageFactory so that all sqlite memory implementations use the same instance.

At this point the test_2_party_circuit_creation passes (after a little reorg). The other tests fail due issues unrelated to the testing framework.

### DIFF
--- a/splinterd/src/node/builder/network.rs
+++ b/splinterd/src/node/builder/network.rs
@@ -1,0 +1,90 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Builder for the NetworkSubsystem
+
+use std::time::Duration;
+
+use rand::{thread_rng, Rng};
+use splinter::error::InternalError;
+use splinter::transport::multi::MultiTransport;
+use splinter::transport::socket::TcpTransport;
+
+use crate::node::runnable::network::RunnableNetworkSubsystem;
+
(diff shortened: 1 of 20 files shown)


peterschwarz (changes requested):

splinterd/src/node/runnable/network.rs
```diff
@@ -0,0 +1,365 @@
...
+            .into_iter()
+            .map(|mut network_listener| {
+                let connection_connector_clone = connection_connector.clone();
+                thread::spawn(move || {
+                    let endpoint = network_listener.endpoint();
+                    for connection_result in network_listener.incoming() {
+                        let connection = match connection_result {
+                            Ok(connection) => connection,
+                            Err(AcceptError::ProtocolError(msg)) => {
+                                println!("Failed to accept connection on {}: {}", endpoint, msg);
```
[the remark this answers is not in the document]
agunde406: ya accidentally put that change in the wrong commit

splinterd/src/node/builder/network.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn with_network_endpoints(mut self, network_endpoints: Vec<String>) -> Self {
+        self.network_endpoints = Some(network_endpoints);
+        self
+    }
+
+    pub fn build(mut self) -> Result<RunnableNetworkSubsystem, InternalError> {
+        let node_id = self
+            .node_id
+            .take()
+            .unwrap_or_else(|| format!("n{}", thread_rng().gen::<u16>().to_string()));
```
peterschwarz, line 70: I wonder if this generation should be provided in one place - otherwise, different subsystems might end up with different values.

agunde406: Ive already removed it from the admin service (admin service subsystem errors if the node id is not provided) and it gets generated in the network subsystem and passed to the admin service builder. I can move it up to the node builder though.

agunde406: Added a commit to move node_id generations to the NodeBuilder and pass it to the subsystem.

peterschwarz (approved):

peterschwarz (approved):

shannynalayna (approved):

PR #1245, opened 2021-03-15
Update circuit integration tests

shannynalayna:
This PR updates the circuit integration tests in the following ways:
- Updates how the tests wait for the change to be committed. Now, all nodes are checked before moving out of the loop to ensure all nodes are on the same page before moving on. Previously, only the nodes that hadn't submitted the payload would be checked in the wait loop. This resulted in issues where that submitter node's state wasn't up-to-date and the test would continue on, assuming it was without validating.
- Updates the happy path circuit create and circuit disband tests to include validation that a duplicate proposal will be rejected. This allows us to validate the case that duplicate proposals are rejected without running separate tests for these cases. 
- Also updates the comments for these tests to account for these updates
- Un-ignores 2 tests, `test_3_party_circuit_creation` and `test_2_party_circuit_lifecycle` as these tests are now passing with these changes

### DIFF
--- a/splinterd/tests/admin/circuit_lifecycle.rs
+++ b/splinterd/tests/admin/circuit_lifecycle.rs
@@ -943,7 +943,6 @@ pub fn test_3_party_circuit_creation_proposal_rejected() {
 /// 10. Validate the disbanded circuit is still available to each node, though disbanded, and that
 ///    the disbanded circuit is the same for each node
 #[test]
-#[ignore]
 pub fn test_2_party_circuit_lifecycle() {
     // Start a 2-node network
     let mut network = Network::new()
--- a/splinterd/tests/admin/circuit_lifecycle.rs
+++ b/splinterd/tests/admin/circuit_lifecycle.rs
@@ -596,7 +596,6 @@ pub fn test_2_party_circuit_creation() {
 /// 10. Wait until the circuit becomes available for one of the other nodes, using `list_circuits`
 /// 11. Validate the circuit is available to every node
 #[test]
-#[ignore]
 pub fn test_3_party_circuit_creation() {
     // Start a 3-node network
     let mut network = Network::new()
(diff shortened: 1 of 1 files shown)


splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -665,22 +714,21 @@ pub fn test_2_party_circuit_creation_proposal_rejected() {
...
         }
-        if node_a
+        let proposals_a = node_a
             .admin_service_client()
             .list_proposals(None, None)
-            .expect("Unable to list proposals from first node")
-            .data
-            .is_empty()
-        {
+            .expect("Unable to list pro")
```
isabeltomb, line 720: End of message missing

splinterd/tests/admin/circuit_lifecycle.rs
```diff
@@ -875,10 +934,12 @@ pub fn test_3_party_circuit_creation_proposal_rejected() {
 /// 2. Create and submit a `CircuitDisbandRequest` from the first node
 /// 3. Wait until the disband proposal is available to the second node, using `list_proposals`
 /// 4. Verify the same disband proposal is available on each node
+/// 5. Submit the same `CircuitDisbandRequest` from the second step to the second node
+/// 6. Validate this duplicate disband proposal is rejected
 /// 5. Create and submit a `CircuitProposalVote` from the second node to accept the disband proposal
 /// 6. Wait until the circuit is no longer available as an active circuit on the first node,
```
isabeltomb, line 940: Numbering is off

isabeltomb (changes requested):

splinterd/tests/admin/circuit_lifecycle.rs
shannynalayna: Updated the error message here

splinterd/tests/admin/circuit_lifecycle.rs
shannynalayna: Updated the numbering here

isabeltomb (approved):

isabeltomb (approved):

agunde406 (approved):

PR #1248, opened 2021-03-16
Registry client

isabeltomb:
- Add a `RegistryClient` trait and a reqwest based implementation of the trait, the client is behind an experimental feature `registry-client` and the reqwest based implementation is behind an experimental feature `registry-client-reqwest`
- Add `registry_client` to Node so that the registry client can be fetched from a node in registry integration testing.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -117,6 +117,8 @@ experimental = [
     "client-reqwest",
     "https-bind",
     "oauth-profile",
+    "registry-client",
+    "registry-client-reqwest",
     "rest-api-actix-web-3",
     "service-arg-validation",
     "service-network",
@@ -151,6 +153,8 @@ memory = ["sqlite"]
 oauth = ["biome", "oauth2", "reqwest", "rest-api"]
 postgres = ["diesel/postgres", "diesel_migrations"]
 registry = []
+registry-client = ["registry"]
+registry-client-reqwest = ["registry-client", "reqwest"]
 registry-remote = ["reqwest", "registry"]
 rest-api = [
     "actix",
--- a/libsplinter/src/registry/client/mod.rs
+++ b/libsplinter/src/registry/client/mod.rs
@@ -1,0 +1,109 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


libsplinter/src/registry/client/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Delete a node from the registry.
+    ///
+    /// # Arguments
+    ///
+    /// * `identity` - The id of the node to be deleted
+    fn delete_node(&self, identity: &str) -> Result<(), InternalError>;
+}
+
+#[cfg(feature = "registry")]
```
vaporos, line 65: This cfg and the one below are unnecessary as the feature guard exists around the entire module.

libsplinter/src/registry/mod.rs
```diff
@@ -24,6 +24,8 @@
 //! [`RegistryWriter`]: trait.RegistryWriter.html
 //! [`RwRegistry`]: trait.RwRegistry.html
 
+#[cfg(feature = "registry")]
```
vaporos, line 27: This should be changed to "registry-client".

libsplinter/src/registry/client/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Traits and implementations useful for communicating with the registry as a client.
+
+use std::fmt;
+
+mod reqwest;
```
vaporos, line 19: This should be protected by a feature called "registry-client-reqwest".

libsplinter/Cargo.toml
```diff
@@ -151,7 +151,7 @@ https-bind = ["actix-web/ssl"]
 memory = ["sqlite"]
 oauth = ["biome", "oauth2", "reqwest", "rest-api"]
 postgres = ["diesel/postgres", "diesel_migrations"]
-registry = []
+registry = ["reqwest"]
```
vaporos, line 154: Replace this with something like:

registry = []
registry-client = ["registry"]
registry-client-reqwest = ["registry-client", "reqwest"]

libsplinter/src/registry/client/mod.rs
isabeltomb: That got copied over from the original client, removed

libsplinter/src/registry/mod.rs
isabeltomb: Fixed

libsplinter/src/registry/client/mod.rs
isabeltomb: Fixed

libsplinter/Cargo.toml
isabeltomb: Fixed

shannynalayna (approved):

vaporos (approved):

PR #1259, opened 2021-03-18
Separate admin service integration tests by concern

shannynalayna:
This PR separates the content currently in the `circuit_lifecycle` module into separate modules based on the concern. This includes separating the helper functions into `payload` and `circuit_commit`, which contains functions to create the `CircuitManagementPayload` bytes used in the integration tests and the functions to commit 2- and 3-party circuits, respectively. This also includes separating the tests defined in this module into a `circuit_disband` and `circuit_create` module (based on what the test is validating). This will allow for tests to be added around other circuit operations that also fit under the umbrella of `circuit_lifecycle`, but without overloading the single file and making it overwhelming to viewers.
This re-organization does not change the functionality of the tests.

This PR does make one small functional change: fixes an if-statement to check if 2 lists are empty in the `3_party_circuit_lifecycle` test that intermittently produced an error as an empty list was able to pass this if-statement, and then a `None` value is unwrapped. (As the test assumes if this if-statement has been passed, both lists are not empty)


splinterd/tests/admin/mod.rs
```diff
@@ -14,6 +14,8 @@
 
 //! Admin service integration tests.
 
+#[cfg(feature = "node")]
```
agunde406, line 17: Could this features guard be moved up a level into admin_service.rs. Nothing in this file is not guarded by "node".

agunde406 (changes requested):

splinterd/tests/admin/mod.rs
shannynalayna: Probably, I don't remember specific reasoning for guarding each individually

shannynalayna: I added a commit to move that feature guard to the top-level module import

agunde406 (approved):

isabeltomb (approved):

PR #1261, opened 2021-03-19
Add external registries to testing framework

isabeltomb:
Add the ability to add a yaml file as an external registry to the testing framework. If any external registries are added a unified registry is created from the internal registry and external read-only registries.

A yaml registry file can be added by passing the registry file names in a vector to the `with_external_registries` method when creating a new Network.

Example:
```
let network = Network::new()
        .with_default_rest_api_variant(RestApiVariant::ActixWeb1)
        .with_external_registries(vec![<file names in the format "file://file-path">])
        .add_nodes_with_defaults(1)
```

### DIFF
--- a/splinterd/src/node/runnable/admin.rs
+++ b/splinterd/src/node/runnable/admin.rs
@@ -24,6 +24,7 @@ use splinter::circuit::routing::RoutingTableWriter;
 use splinter::error::InternalError;
 use splinter::orchestrator::ServiceOrchestratorBuilder;
 use splinter::peer::PeerManagerConnector;
+use splinter::registry::{LocalYamlRegistry, RegistryReader, UnifiedRegistry};
 use splinter::rest_api::actix_web_1::RestResourceProvider as _;
 use splinter::service::ServiceProcessorBuilder;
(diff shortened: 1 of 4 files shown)


splinterd/src/node/runnable/admin.rs
```diff
@@ -51,7 +53,52 @@ impl RunnableAdminSubsystem {
         let mut service_transport = self.service_transport;
         let routing_writer = self.routing_writer;
 
-        let registry = store_factory.get_registry_store();
+        let mut registry = store_factory.get_registry_store();
+
+        if let Some(external_registries) = self.registries {
+            let read_only_registries = external_registries
+                .iter()
+                .filter_map(|registry| {
```
peterschwarz, line 61: I’m not sure `filter_map` is the right way to go here. We should probably propagate the errors that can occur on the individual registries. 

Better to `map` and then `collect::<Result<Vec<_>, _>>()?`

peterschwarz (changes requested):

peterschwarz (approved):

agunde406 (approved):

PR #1263, opened 2021-03-22
Add initial registry tests

isabeltomb:
Add initial registry integration tests to test the get, list, add, update, and delete functionality of the registry

### DIFF
--- a/splinterd/tests/admin/registry.rs
+++ b/splinterd/tests/admin/registry.rs
@@ -1,0 +1,453 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


splinterd/tests/admin/registry.rs
```diff
@@ -0,0 +1,451 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Integration tests for the registry REST API endpoints
+
+use crate::framework::network::Network;
```
vaporos, line 17: The order of use statements here should be:

```

use splinter::registry::client::RegistryNode;
use splinterd::node::RestApiVariant;

use crate::framework::network::Network;
```

Basically less-to-more specific, in groups (std, crates, crate::, super, children modules).

isabeltomb: Noted and fixed

vaporos (approved):

peterschwarz (approved):

PR #1268, opened 2021-03-23
Add Scabbard transactions to circuit integration tests

shannynalayna:
This PR adds the necessary plumbing to create and submit a batch to the Scabbard client's `submit` endpoint, and also adds validation in the currently implemented tests for the submission of scabbard transactions. 

This PR also makes an update to the `test_3_party_circuit_lifecycle` test, which allows this test to pass so this PR also removes the `#[ignore]` from this test so that it is able to run in CI.

### DIFF
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -51,6 +51,8 @@ toml = "0.5"
 
 [dev-dependencies]
 openssl = { version = "0.10" }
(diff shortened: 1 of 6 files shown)


shannynalayna:
Closing in favor of #1282,  also a bug was uncovered in the circuit-disband tests so this will be reworked once the bug in these function as expected (noted in #1284)

PR #1279, opened 2021-03-31
Update scabbard to handle both json lists and csv for arguments

agunde406:
This PR also updates the `splinter circuit propose   --service-arg` to take the value as provided, before it was wrapped in a Vec but this caused issues with setting the version service arg because that should be in a vec. 

The Scabbard factory can now handle peer_services and admin_keys in either json fmt or csv:

`--service-arg "*::admin_keys=admin_keys=key1,key2"`
or 
`--service-arg "*::admin_keys=[\"key1\", \"key2\"]"`

### DIFF
--- a/services/scabbard/libscabbard/src/service/factory.rs
+++ b/services/scabbard/libscabbard/src/service/factory.rs
@@ -156,15 +156,16 @@ impl ServiceArgValidator for ScabbardArgValidator {
         let admin_keys_str = args
             .get("admin_keys")
             .ok_or_else(|| ServiceArgValidationError("admin_keys argument not provided".into()))?;
-        let admin_keys: Vec<String> = serde_json::from_str(admin_keys_str).map_err(|err| {
+
+        let admin_keys = parse_list(admin_keys_str).map_err(|err| {
             ServiceArgValidationError(format!("failed to parse admin_keys list: {}", err,))
         })?;
 
         for key in admin_keys {
             let key_bytes = parse_hex(&key).map_err(|_| {
                 ServiceArgValidationError(format!(
-                    "{} is not a valid hex-formatted public key",
-                    key
+                    "{:?} is not a valid hex-formatted public key",
+                    key,
                 ))
             })?;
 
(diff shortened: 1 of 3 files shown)


services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -156,15 +156,25 @@ impl ServiceArgValidator for ScabbardArgValidator {
...
+        };
 
         for key in admin_keys {
-            let key_bytes = parse_hex(&key).map_err(|_| {
+            let key_bytes = parse_hex(&key).map_err(|err| {
                 ServiceArgValidationError(format!(
-                    "{} is not a valid hex-formatted public key",
-                    key
+                    "{:?} is not a valid hex-formatted public key {}",
+                    key, err
```
vaporos, line 177: Is err relevant here? If it is, should there be a : before {}?

vaporos (approved):

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -156,15 +156,25 @@ impl ServiceArgValidator for ScabbardArgValidator {
         let admin_keys_str = args
             .get("admin_keys")
             .ok_or_else(|| ServiceArgValidationError("admin_keys argument not provided".into()))?;
-        let admin_keys: Vec<String> = serde_json::from_str(admin_keys_str).map_err(|err| {
```
peterschwarz, line 159: Typo in commit message: s/facotry/factory

services/scabbard/libscabbard/src/service/factory.rs
agunde406: I actually meant to remove this. I added it when I was debugging. I will remove it.

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -156,15 +156,25 @@ impl ServiceArgValidator for ScabbardArgValidator {
...
-            ServiceArgValidationError(format!("failed to parse admin_keys list: {}", err,))
-        })?;
+
+        let admin_keys = {
+            if admin_keys_str.starts_with('[') {
```
peterschwarz, line 161: Can we pull this list parsing out to a function? (you do the same operation three times here)

peterschwarz (changes requested):

services/scabbard/libscabbard/src/service/factory.rs
agunde406: fixed

services/scabbard/libscabbard/src/service/factory.rs
agunde406: fixed

services/scabbard/libscabbard/src/service/factory.rs
agunde406: added

peterschwarz (approved):

vaporos (approved):

peterschwarz (approved):

PR #1281, opened 2021-03-31
Splinterd database bug fix

isabeltomb:
Currently splinterd creates a database if one doesn't exist, allowing splinterd to start successfully and later log an error because the necessary tables aren't in the database.

This PR adds the following to keep this from happening:
- When using a file for the database source, check that the file exists and that it is not empty. Return an error and shut down splinterd if either case is true.

### **Testing:**

Start splinter locally without a database:
If the file `/tmp/splinter/data/splinter_state.db` exists delete it then run the following commands
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml --features stable -- cert generate --skip`
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features stable -- --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080`

Check that splinterd reports the daemon failed to start because it was unable to set up storage

Then run successfully with a database:

SQLite:
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml --features stable -- database migrate`
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features stable -- --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080`

Postgres:
`$ docker run --name postgres -p 5432:5432 -e POSTGRES_PASSWORD=mypassword -e POSTGRES_DB=splinter -d postgres`
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path cli/Cargo.toml -- database migrate -C postgres://postgres:mypassword@localhost:5432/splinter`
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features stable -- --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080 --database postgres://postgres:mypassword@localhost:5432/splinter`

Memory:
`$ SPLINTER_HOME=/tmp/splinter cargo run --manifest-path splinterd/Cargo.toml --features stable -- --tls-insecure --enable-biome -vv --rest-api-endpoint http://127.0.0.1:8080 --database memory`

### DIFF
--- a/libsplinter/src/store/mod.rs
+++ b/libsplinter/src/store/mod.rs
@@ -95,6 +95,12 @@ pub fn create_store_factory(
         }
         #[cfg(feature = "sqlite")]
         ConnectionUri::Sqlite(conn_str) => {
+            if (conn_str != ":memory:") && !std::path::Path::new(&conn_str).exists() {
+                return Err(InternalError::with_message(format!(
+                    "Database file '{}' does not exist",
+                    conn_str
+                )));
+            }
             let connection_manager =
                 ConnectionManager::<diesel::sqlite::SqliteConnection>::new(&conn_str);
             let mut pool_builder =


splinterd/src/daemon.rs
```diff
@@ -982,6 +984,35 @@ fn start_health_service(
 fn create_store_factory(
     db_url: &str,
 ) -> Result<Box<dyn splinter::store::StoreFactory>, StartError> {
+    // If the given db_url is a file path check that the file exists and is not empty
+    if !db_url.starts_with("postgres://") & (db_url != "memory") & (db_url != ":memory:") {
```
vaporos, line 988: This should use && (short-circuit logical and) instead of & (bitwise and).

splinterd/src/daemon.rs
```diff
@@ -982,6 +984,35 @@ fn start_health_service(
 fn create_store_factory(
     db_url: &str,
 ) -> Result<Box<dyn splinter::store::StoreFactory>, StartError> {
+    // If the given db_url is a file path check that the file exists and is not empty
```
vaporos, line 987: The logic here is similar to that which happens in db_url.parse() in the following statement, which determines the type of database URI which has been provided (postgres or sqlite). It is probably easier to maintain consistency if the  check is done against the connection_uri (and only for sqlite variant).

It may make sense to place this check within create_store_factory() to ensure that it it is not possible to create a factory which accidentally creates empty files.

splinterd/src/daemon.rs
```diff
@@ -982,6 +984,35 @@ fn start_health_service(
...
+            )));
+        } else {
+            let mut file = File::open(db_url).map_err(|err| {
+                StartError::StorageError(format!(
+                    "Unable to open database file '{}' : {}",
+                    db_url, err
+                ))
+            })?;
+            let mut buffer = Vec::new();
+            file.read_to_end(&mut buffer).map_err(|err| {
```
vaporos, line 1002: This will not work for very large databases as it will be constrained by the size of memory.

It is probably sufficient here to check for the existence of the file (Path::new().exists()) in order to ensure we are not going to accidentally create an empty database.

vaporos (changes requested):

splinterd/src/daemon.rs
isabeltomb: Moved this to `create_store_factory()`

splinterd/src/daemon.rs
isabeltomb: Removed this from the check so now only the existence of the file is checked

peterschwarz (approved):

vaporos (approved):

PR #1282, opened 2021-04-01
Add circuit-abandon integration tests

shannynalayna:
This PR adds the following to the circuit-integration tests: 
- Updates the `commit_*_circuit` functions to handle multiple circuits, this enables multiple circuits to be committed in the circuit integration tests
- Updates the helper function that creates the `CircuitCreateRequest` to take a list of public keys to be used as the `admin_keys` service arg for each node's service to allow service transactions from any of the nodes on the circuit.
- Adds plumbing to submit scabbard transactions from the circuit integration tests
- Updates the `ReqwestAdminServiceClient` to handle `circuit_status` filters. Initially, the client would add filters to the url with `filter={}`, but the `circuit_status` filter is expected as `status={}` in the url thus requiring the update. 
-  Adds initial circuit-abandon integration tests, which validate the process of abandoning a circuit between 2- and 3-nodes.

Additionally, this PR contains the Scabbard plumbing introduced in #1268 but does not contain the service transaction updates to the existing circuit-integration tests.

### DIFF
--- a/splinterd/tests/admin/circuit_commit.rs
+++ b/splinterd/tests/admin/circuit_commit.rs
@@ -53,6 +53,18 @@ pub(in crate::admin) fn commit_2_party_circuit(circuit_id: &str, node_a: &Node, 
         node_a.node_id(),
         node_info,
         &*node_a.admin_signer().clone_box(),
+        &vec![
+            node_a
+                .admin_signer()
+                .public_key()
+                .expect("Unable to get first node's public key")
+                .as_hex(),
+            node_b
+                .admin_signer()
+                .public_key()
+                .expect("Unable to get second node's public key")
+                .as_hex(),
+        ],
     );
     // Submit the `CircuitManagementPayload` to the first node
     let res = node_a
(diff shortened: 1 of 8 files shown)


splinterd/tests/admin/payload.rs
```diff
@@ -236,3 +238,25 @@ fn setup_circuit(
...
+) -> Batch {
+    let owners = vec![signer
+        .public_key()
+        .expect("Unable to get signer's public key")
+        .as_hex()];
+    CreateContractRegistryActionBuilder::new()
+        .with_name(name.into())
+        .with_owners(owners)
+        .into_payload_builder()
+        .expect("Unable to create payload builder for `CreateContractRegistryActionBuilder`")
```
peterschwarz, line 303: I think that we should probably propagate these as InternalError, and have the `expect` done by the caller.

splinterd/tests/admin/circuit_commit.rs
```diff
@@ -64,25 +64,19 @@ pub(in crate::admin) fn commit_2_party_circuit(circuit_id: &str, node_a: &Node,
     assert!(res.is_ok());
 
     // Wait for the proposal to be committed for the second node
-    let mut proposal_b;
```
peterschwarz, line 55: This commit might not be needed after #1284 goes in.

shannynalayna: Yep 👍

shannynalayna: I did have to add a commit to change the circuit's `management_type` so we don't have to skip through events or save an event ID if we attempt to commit multiple circuits. I found that as is, the `circuit_commit` function would receive old events from previously committed circuits (since the events are starting from the beginning) which would lead the test to run into issues. So, making that management type unique makes it so we can subscribe to the specific circuit's events.

splinterd/tests/admin/payload.rs
shannynalayna: Ok, updated these to propagate these as `InternalError`

peterschwarz (approved):

peterschwarz (approved):

agunde406 (approved):

PR #1283, opened 2021-04-02
Implement Node stop in integration test framework

isabeltomb:
Implement Node::stop() in the integration test framework

- Add a `clone_box` method to the `StoreFactory` trait and implement `Clone` for `StoreFactory`
- Add a `store_factory` field to `AdminSubsystem` in the integration test framework so that a store factory can be retrieved from a node and used in the creation of a `RunnableNode` when a node is stopped, this allows the same store factory remain with a node throughout the node lifecycle 
- Add a `stop` method to node that will shut down a node and create a `RunnableNode` from the node data
- Create a `NetworkNode` enum in `Network` to allow for the node list in `Network` to hold both stopped and running nodes and maintain the original order the nodes were created in
- Add methods to `Network` for starting and stopping nodes
- Add a test that starts a network of three nodes and runs through stopping and then starting each node

### **Testing:**
Run the `test_stop_start_nodes_multiple` test in splinterd/tests/admin/node_lifecycle.rs

### DIFF
--- a/splinterd/src/node/running/mod.rs
+++ b/splinterd/src/node/running/mod.rs
@@ -31,6 +31,8 @@ use splinter::rest_api::actix_web_1::RestApiShutdownHandle;
 use splinter::rest_api::actix_web_3::RestApi;
 use splinter::threading::lifecycle::ShutdownHandle;
 
+use super::{running::admin::AdminSubsystem, NodeBuilder, RestApiVariant, RunnableNode};
+
 pub(super) enum NodeRestApiVariant {
     ActixWeb1(RestApiShutdownHandle, JoinHandle<()>),
     ActixWeb3(RestApi),
@@ -89,6 +91,58 @@ impl Node {
             format!("http://localhost:{}", self.rest_api_port),
             "foo".to_string(),
         ))
+    }
+
+    pub fn stop(mut self) -> Result<RunnableNode, InternalError> {
+        self.signal_shutdown();
+
+        let Node {
+            admin_signer,
+            admin_subsystem,
+            rest_api_variant,
+            node_id,
+            rest_api_port,
+            mut network_subsystem,
+        } = self;
+
+        let rest_api_variant = match rest_api_variant {
+            // shutdown node
+            NodeRestApiVariant::ActixWeb1(shutdown_handle, join_handle) => {
+                shutdown_handle
+                    .shutdown()
+                    .map_err(|e| InternalError::from_source(Box::new(e)))?;
+                join_handle.join().map_err(|_| {
+                    InternalError::with_message(
+                        "REST API thread panicked, join() failed".to_string(),
+                    )
+                })?;
+
+                RestApiVariant::ActixWeb1
+            }
+            NodeRestApiVariant::ActixWeb3(rest_api) => {
+                rest_api
+                    .wait_for_shutdown()
+                    .map_err(|err| InternalError::from_source(Box::new(err)))?;
+
+                RestApiVariant::ActixWeb3
+            }
+        };
+
+        let AdminSubsystem { store_factory, .. } = admin_subsystem;
+
+        // shutdown admin subsystem and network subsystem
+        admin_subsystem.admin_service_shutdown.wait_for_shutdown()?;
+        network_subsystem.signal_shutdown();
+        network_subsystem.wait_for_shutdown()?;
+
+        // create a runnable node from the node data
+        NodeBuilder::new()
+            .with_node_id(node_id)
+            .with_rest_api_variant(rest_api_variant)
+            .with_admin_signer(admin_signer.to_owned())
+            .with_rest_api_port(rest_api_port.into())
+            .with_store_factory(store_factory)
+            .build()
     }
 }
 
--- a/splinterd/tests/admin/node_lifecycle.rs
+++ b/splinterd/tests/admin/node_lifecycle.rs
@@ -1,0 +1,161 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


splinterd/tests/admin/node_lifecycle.rs
```diff
@@ -0,0 +1,161 @@
...
+/// 14. Start the third node and check that it can now be retrieved from the network
+/// 15. Check that the third node is running by getting its registry client and using it
+/// 16. Shutdown
+#[test]
+pub fn test_stop_start_nodes_multiple() {
+    // create network with three nodes
+    let mut network = Network::new()
+        .with_default_rest_api_variant(RestApiVariant::ActixWeb1)
+        .add_nodes_with_defaults(3)
+        .expect("Unable to start single node ActixWeb1 network");
```
shannynalayna, line 49: Copy pasta in this expect message

splinterd/tests/admin/node_lifecycle.rs
```diff
@@ -0,0 +1,161 @@
...
+        .registry_client()
+        .get_node(node_2.node_id())
+        .expect("Unable to list nodes from registry");
+    assert_eq!(
+        node_2.network_endpoints(),
+        registry_response.unwrap().endpoints
+    );
+
+    // Stop the third node
+    network = network.stop(2).expect("Unable to stop the second node");
```
shannynalayna, line 93: Copy pasta in the expect message here again. This comment applies to the following expects as well, I noticed a few other expect messages that don't pertain to the thing being expected (below this line in this file).

shannynalayna (changes requested):

splinterd/tests/admin/node_lifecycle.rs
isabeltomb: Fixed

splinterd/tests/admin/node_lifecycle.rs
isabeltomb: Fixed

shannynalayna (dismissed):

splinterd/tests/framework/network.rs
```diff
@@ -20,16 +20,21 @@ use cylinder::{secp256k1::Secp256k1Context, Context};
...
 pub struct Network {
     default_rest_api_variant: RestApiVariant,
-    nodes: Vec<Node>,
+    nodes: Vec<NetworkNode>,
     temp_dirs: HashMap<String, TempDir>,
     external_registries: Option<Vec<String>>,
 }
 
+pub enum NetworkNode {
+    RunningNode(Node),
```
vaporos, line 34: This could be simplified to Node(Node), RunnableNode(RunnableNode).

libsplinter/src/store/memory.rs
```diff
@@ -37,6 +37,7 @@ use super::sqlite::ForeignKeyCustomizer;
 use super::StoreFactory;
 
 /// A `StoryFactory` backed by memory.
+#[derive(Clone)]
```
vaporos, line 40: I'm concerned that cloning the factory, while harmless in some circumstances, may better enable patterns which exhaust the connection pool size--when re-using an already allocated store may be a better option.

splinterd/src/node/running/mod.rs
```diff
@@ -90,6 +92,26 @@ impl Node {
...
+            NodeRestApiVariant::ActixWeb1(_, _) => RestApiVariant::ActixWeb1,
+            NodeRestApiVariant::ActixWeb3(_) => RestApiVariant::ActixWeb3,
+        };
+
+        let runnable_node = NodeBuilder::new()
+            .with_node_id(self.node_id().to_string())
+            .with_rest_api_variant(rest_api_variant)
+            .with_admin_signer(self.admin_signer.to_owned())
+            .with_rest_api_port(self.rest_api_port().into())
+            .with_store_factory(self.admin_subsystem.store_factory.to_owned())
```
vaporos, line 107: This clone could be avoided by taking the store factory out of the admin subsystem (which in this function may not need a lot of Option to accomplish because it consumes self).

vaporos: See - https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_structures.html

splinterd/tests/framework/network.rs
isabeltomb: Updated

libsplinter/src/store/memory.rs
isabeltomb: Removed this commit

splinterd/src/node/running/mod.rs
isabeltomb: Updated this method to take store factory out of admin subsystem and then do the shutdown procedure

shannynalayna (approved):

splinterd/tests/framework/network.rs
```diff
@@ -70,22 +75,31 @@ impl Network {
...
+                                    .with_endpoints(endpoints.to_vec())
+                                    .with_key(pub_key.as_hex())
+                                    .build()
+                                    .map_err(|e| InternalError::from_source(Box::new(e)))?,
+                            )
+                            .map_err(|e| InternalError::from_source(Box::new(e)))?;
+                    }
+                }
+                _ => {
+                    return Err(InternalError::with_message(
```
peterschwarz, line 99: This could probably be an `unreachable!()` stmt.

Currently, this method isn't written to support multiple calls correctly, as it iterates over the whole list of nodes.  We  should probably correct that in a future PR.

isabeltomb: Replaced with `unreachable!()`

vaporos (approved):

shannynalayna (approved):

peterschwarz (approved):

PR #1284, opened 2021-04-02
Add AdminServiceEventClient

peterschwarz:
This PR adds the AdminServiceEventClient trait and an implementation backed by client websockets.  It updates the circuit tests to make use of these clients to verify that the correct events have occurred.

### DIFF
--- a/libsplinter/src/admin/client/event/mod.rs
+++ b/libsplinter/src/admin/client/event/mod.rs
@@ -94,3 +94,40 @@ pub trait AdminServiceEventClient {
     /// This should block until an event is available.
     fn next_event(&self) -> Result<AdminServiceEvent, NextEventError>;
 }
+
+impl AdminServiceEventClient for Box<dyn AdminServiceEventClient> {
+    fn try_next_event(&self) -> Result<Option<AdminServiceEvent>, NextEventError> {
+        (**self).try_next_event()
+    }
+
+    fn next_event(&self) -> Result<AdminServiceEvent, NextEventError> {
+        (**self).next_event()
+    }
+}
+
+pub struct BlockingAdminServiceEventIterator<T>
+where
+    T: AdminServiceEventClient,
+{
+    client: T,
+}
+
+impl<T> BlockingAdminServiceEventIterator<T>
+where
+    T: AdminServiceEventClient,
+{
+    pub fn new(client: T) -> BlockingAdminServiceEventIterator<T> {
+        Self { client }
+    }
+}
+
+impl<T> Iterator for BlockingAdminServiceEventIterator<T>
+where
+    T: AdminServiceEventClient,
+{
+    type Item = AdminServiceEvent;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.client.next_event().ok()
+    }
+}
--- a/libsplinter/src/admin/client/event/mod.rs
+++ b/libsplinter/src/admin/client/event/mod.rs
@@ -1,0 +1,89 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Client traits to receive AdminServiceEvents.
+
+mod error;
+
+use super::ProposalSlice;
+
+pub use error::NextEventError;
+
+/// A public key for the private key that signed an admin proposal.
+#[derive(Clone, PartialEq)]
+#[repr(transparent)]
+pub struct PublicKey(pub Vec<u8>);
+
+impl std::fmt::Debug for PublicKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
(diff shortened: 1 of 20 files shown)


shannynalayna (approved):

libsplinter/src/admin/client/event/ws.rs
```diff
@@ -0,0 +1,439 @@
...
+                },
+            ),
+            ProposalRejected((proposal, pub_key_bytes)) => (
+                proposal,
+                EventType::ProposalRejected {
+                    requester: PublicKey(pub_key_bytes),
+                },
+            ),
+            CircuitReady(proposal) => (proposal, EventType::CircuitReady),
+            CircuitDisbanded(proposal) => (proposal, EventType::CircuitReady),
```
shannynalayna, line 335: This result should be `EventType::CircuitDisbanded`, which I believe was causing the error you ran into with the disband tests.

shannynalayna (changes requested):

libsplinter/src/admin/client/event/ws.rs
peterschwarz: Fixed

shannynalayna (dismissed):

libsplinter/src/admin/client/event/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// An admin service event client will provide events as they are returned from an admin service.
+///
+/// It provides two methods, a blocking method and a non-blocking method, depending on the caller's
+/// tolerance for blocking a thread.
+pub trait AdminServiceEventClient {
+    /// Returns the next event, if one is available.
+    ///
+    /// This should be a non-blocking function.
+    fn try_next_event(&self) -> Result<AdminServiceEvent, TryNextEventError>;
```
vaporos, line 88: Could this just be Result<Option<AdminServiceEvent>, NextEventError>>?

libsplinter/src/admin/client/event/ws.rs
```diff
@@ -0,0 +1,439 @@
+// Copyright 2018-2021 Cargill Incorporated
```
vaporos, line 1: Consider moving this to client::event::ws::actix_web_1 module and guarding with the actix_web_1 feature.

libsplinter/src/admin/client/event/mod.rs
peterschwarz: The intent was to mirror the API of `recv` and `try_recv` on `Receiver`, but without indicating too tightly that implementations might be using channels under the hood.

splinterd/src/node/runnable/mod.rs
```diff
@@ -20,6 +20,7 @@ pub(super) mod network;
 use std::net::{Ipv4Addr, SocketAddr};
 
 use splinter::error::InternalError;
+use splinter::events::Reactor;
```
vaporos, line 23: This has actix requirements and should be handled separately based on actix web 1 vs 3.

splinterd/Cargo.toml
```diff
@@ -132,6 +132,8 @@ database-sqlite = ["splinter/sqlite"]
 health-service = ["health"]
 https-bind = ["splinter/https-bind"]
 node = [
+    "authorization",
```
vaporos, line 135: This should probably pull in the libsplinter dependencies directly, since this local features are there mostly to guard splinterd code fragments.

libsplinter/src/admin/service/shared.rs
```diff
@@ -1648,6 +1648,43 @@ impl AdminServiceShared {
...
+            match protocol {
+                0 => {
+                    if pending_protocol_payload
+                        .missing_protocol_ids
+                        .iter()
+                        .any(|missing_protocol_id| missing_protocol_id == service_id)
+                    {
+                        warn!(
+                            "Dropping circuit request including service {}, \
+                             due to protocol mismatch",
```
vaporos, line 1663: Add the mismatched protocol versions to this message for debugging purposes.

peterschwarz: It's actually in a message in the caller of this function.  Any bad protocol at the call site is translated to `0`

splinterd/Cargo.toml
peterschwarz: The daemon code won't compile, if it is just the splinter features, so you can't compile with just `--features node` without it

libsplinter/src/admin/client/event/mod.rs
peterschwarz: Updated to return an option

libsplinter/src/admin/client/event/ws.rs
peterschwarz: Moved this to `client::event::ws::actix_web_client`, to provide a similar separation. This being purely client code, and not relying on actix-web, it did not make sense to put in the actix_web_1 feature

splinterd/src/node/runnable/mod.rs
peterschwarz: I've added  AdminServiceEventClientVariant enums, which acts similarly to the RestApiVariant enums

libsplinter/src/events/error.rs
```diff
@@ -21,7 +21,7 @@ use tokio::io;
 pub enum ParseError {
     MalformedMessage(Box<dyn error::Error + Send + Sync + 'static>),
     /// This type provides a sendable alternative to source errors that may not be Sync and Send.
-    MalformedReducedToString(String)
+    MalformedReducedToString(String),
```
shannynalayna, line 24: It looks like this belongs in a previous commit

peterschwarz: Fixed

shannynalayna (dismissed):

shannynalayna (approved):

vaporos (approved):

PR #1285, opened 2021-04-05
Add `Biome` endpoints to `RunningNode`

shannynalayna:
This PR adds the set-up within the RunningNode to allow for test nodes to access a `BiomeClient`. This is in support of the Biome integration tests.

### DIFF
--- a/splinterd/src/node/builder/mod.rs
+++ b/splinterd/src/node/builder/mod.rs
@@ -23,12 +23,11 @@ use std::time::Duration;
 use cylinder::{secp256k1::Secp256k1Context, Context, Signer};
 use rand::{thread_rng, Rng};
 use splinter::error::InternalError;
-use splinter::rest_api::actix_web_1::{AuthConfig, RestApiBuilder as RestApiBuilder1};
+use splinter::rest_api::actix_web_1::RestApiBuilder as RestApiBuilder1;
 use splinter::rest_api::actix_web_3::RestApiBuilder as RestApiBuilder3;
 use splinter::rest_api::auth::{
     authorization::{AuthorizationHandler, AuthorizationHandlerResult},
-    identity::{Identity, IdentityProvider},
-    AuthorizationHeader,
+    identity::Identity,
 };
 use splinter::rest_api::BindConfig;
 use splinter::store::StoreFactory;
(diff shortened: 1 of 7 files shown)


isabeltomb (approved):

splinterd/src/node/builder/mod.rs
```diff
@@ -58,6 +60,7 @@ pub struct NodeBuilder {
     network_subsystem_builder: NetworkSubsystemBuilder,
     node_id: Option<String>,
     auth_configs: Vec<AuthConfig>,
+    biome_actix1_resources: Vec<Actix1Resource>,
```
peterschwarz, line 63: These are tied to the RestApIVariant, so they shouldn't necessarily be top-level items.  Like with other resources, they may need to be moved to later steps, then the build step.

splinterd/tests/framework/network.rs
```diff
@@ -56,6 +60,17 @@ impl Network {
             let temp_dir = TempDir::new("scabbard_data")
                 .map_err(|e| InternalError::from_source(Box::new(e)))?;
 
+            // Create a store factory to allow for the Biome endpoints to be used
+            let store_factory = Box::new(MemoryStoreFactory::new()?);
+            // Build the `BiomeResourceProvider` to allow the node to access `Biome` endpoints
+            let mut biome_resource_provider = BiomeResourceProviderBuilder::default()
```
peterschwarz, line 66: These should be with-in the node some how.  Probably we should have a `with_biome_enabled` function (for now) on the Node builder, and then they are constructed internally at the appropriate time.

peterschwarz (changes requested):

splinterd/tests/framework/network.rs
shannynalayna: I changed this to add the `with_biome_enabled` to the `NodeBuilder` struct and since the resources that are built need access to the store factory, we can internally construct the resources in the `RunnableNode` module.

splinterd/src/node/builder/mod.rs
shannynalayna: I altered this so that the biome resources are being constructed at the `RunnableNode` stage (since the resources need access to the store factory) so I was able to take these resources out from the top-level and replaced this with just a `biome_enabled` flag for the `NodeBuilder`

splinterd/src/node/runnable/biome.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl BiomeResourceProvider {
+    /// Take the available REST Resources from the Biome resource provider.
+    pub fn take_actix1_resources(&mut self) -> Vec<Actix1Resource> {
+        let mut replaced = vec![];
+        std::mem::swap(&mut self.actix1_resources, &mut replaced);
+        replaced
+    }
+
+    pub fn create_biome_actix1_resources(
```
peterschwarz, line 46: Suggest renaming this to `new`, as it returns a provider, not biome_actix1_resources

shannynalayna: Updated this method to just `new`

peterschwarz (approved):

isabeltomb (approved):

PR #1296, opened 2021-04-15
Improve cli usability

leebradley:
This change makes two minor cli usability improvements.

First, I've corrected what looks like a bug in some of splinterd's error messages. The configuration build step prepends `missing required argument:` to all ConfigError results, despite the fact that many possible ConfigErrors have nothing to do with missing required arguments. When they do, the ConfigError already says this when formatted to a string, making for an oddly redundant error message. With this change, configuration errors during configuration object build will instead be preceded with: `error occurred building config object:`

Second, I've added an explicit check for the existence of the state directory. splinterd already requires a state directory. This was changed in favor of the current behavior of throwing an error at the first attempt to save state (for instance, while creating the node_id file).

Signed-off-by: Lee Bradley bradley@bitwise.io

### DIFF
--- a/splinterd/src/error.rs
+++ b/splinterd/src/error.rs
@@ -26,6 +26,7 @@ use crate::daemon::StartError;
 #[derive(Debug)]
 pub enum UserError {
     TransportError(GetTransportError),
+    #[cfg(feature = "metrics")]
     MissingArgument(String),
     InvalidArgument(String),
     ConfigError(ConfigError),
@@ -61,6 +62,7 @@ impl Error for UserError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match self {
             UserError::TransportError(err) => Some(err),
+            #[cfg(feature = "metrics")]
             UserError::MissingArgument(_) => None,
             UserError::InvalidArgument(_) => None,
             UserError::ConfigError(err) => Some(err),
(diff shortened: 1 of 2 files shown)


splinterd/src/config/mod.rs
```diff
@@ -807,6 +809,17 @@ impl Config {
             debug!("Config: whitelist: {:?} (source: {:?})", list, source,);
         }
     }
+
+    pub fn verify_files(&self) -> Result<(), ConfigError> {
```
vaporos, line 813: This check should be done outside of the config module, as it has nothing to do with verifying the validity of the configuration, but rather the validity of the environment. In other words, the config module's primary concern is config data, but no how that data is applied.

vaporos (changes requested):
Please expand on the explanation within the commit messages -- the PR description is fine, but PRs are not considered as long-lived as the commit history itself.

Not sure the PR description matches the commits; there is a commit for linting fixes but not a commit for changing "missing required argument:" (or it isn't obvious in the diff).

splinterd/src/error.rs
```diff
@@ -26,6 +26,7 @@ use crate::daemon::StartError;
 #[derive(Debug)]
 pub enum UserError {
     TransportError(GetTransportError),
+    #[cfg(feature = "metrics")]
```
vaporos, line 29: This change does not match the commit message.

leebradley: Added an explanation of the linting fix in commit message

vaporos (approved):

agunde406 (approved):

peterschwarz (approved):

PR #1306, opened 2021-04-21
Implement simple backpressure for Scabbard batch submit

agunde406:
This PR contains several bug fixes for the following error that can occur under load
```
[2021-04-13 13:06:28.911] T["Orchestrator Outgoing"] ERROR [splinter::orchestrator::runnable]
 erminating orchestrator  outgoing thread due to error: an orchestration error occurred: 
connection 20ae95a4-b2f0-4a99-8174-29c123420c00 send queue is full
```
The above error breaks the ability to handle any further batches. 

It adds very simple back pressure to scabbard batch submit that will return TooManyRequests if 30 batches are pending. This will continue to be returned until the queue gets to half that amount. This value is arbitrary. There is not any dynamic adjustments to this number. Because only the coordinator keeps track of pending batches, two new scabbard messages were added to that will be used to tell the other services if they should be returning TooManyRequests or accepting the batches. 

The following fixes were moved to https://github.com/Cargill/splinter/pull/1309

It also logs an error and continues instead of existing if the orchestrator outgoing send queue fills up. The message is dropped if this occurs. 

Finally, the capacity of the orchestrator messages queues are increased to 512. Before they were set to 8, which is very small for message queues

### DIFF
--- a/services/scabbard/libscabbard/src/service/mod.rs
+++ b/services/scabbard/libscabbard/src/service/mod.rs
@@ -67,7 +67,7 @@ const SERVICE_TYPE: &str = "scabbard";
 const DEFAULT_COORDINATOR_TIMEOUT: u64 = 30; // 30 seconds
 
 /// Specifies the version of scabbard to use.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq)]
 pub enum ScabbardVersion {
     V1,
     V2,
@@ -130,6 +130,8 @@ impl Scabbard {
             peer_services,
             service_id.clone(),
             signature_verifier,
+            #[cfg(feature = "back-pressure")]
+            version,
         );
 
         let (state_db_path, receipt_db_path) =
@@ -188,6 +190,20 @@ impl Scabbard {
             .to_string())
     }
 
+    /// Get whether the service is currently accepting batches
+    #[cfg(feature = "back-pressure")]
+    pub fn accepting_batches(&self) -> Result<bool, ScabbardError> {
+        let shared = self
+            .shared
+            .lock()
+            .map_err(|_| ScabbardError::LockPoisoned)?;
+
+        match self.version {
+            ScabbardVersion::V1 => Ok(true),
+            ScabbardVersion::V2 => Ok(shared.accepting_batches()),
+        }
+    }
+
     pub fn add_batches(&self, batches: Vec<BatchPair>) -> Result<Option<String>, ScabbardError> {
         let mut shared = self
             .shared
@@ -210,10 +226,10 @@ impl Scabbard {
                 link.push_str(&format!("{},", batch.batch().header_signature()));
 
                 match self.version {
-                    ScabbardVersion::V1 => shared.add_batch_to_queue(batch),
+                    ScabbardVersion::V1 => shared.add_batch_to_queue(batch)?,
                     ScabbardVersion::V2 => {
                         if shared.is_coordinator() {
-                            shared.add_batch_to_queue(batch);
+                            shared.add_batch_to_queue(batch)?;
                         } else {
                             let batch_bytes = batch
                                 .into_bytes()
@@ -438,7 +454,9 @@ impl Service for Scabbard {
                                 BatchPair::from_bytes(message.get_new_batch()).map_err(|err| {
                                     ServiceError::UnableToHandleMessage(Box::new(err))
                                 })?;
-                            shared.add_batch_to_queue(batch);
+                            shared.add_batch_to_queue(batch).map_err(|err| {
+                                ServiceError::UnableToHandleMessage(Box::new(err))
+                            })?;
                         } else {
                             warn!("Ignoring new batch; this service is not the coordinator");
                         }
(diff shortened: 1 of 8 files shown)


services/scabbard/libscabbard/src/service/rest_api/actix/batches.rs
```diff
@@ -68,6 +70,21 @@ pub fn make_add_batches_to_queue_endpoint() -> ServiceEndpoint {
                             }
                         };
 
+                        match scabbard.pending_batches() {
+                            Ok(num) => {
+                                if num >= DEFAULT_PENDING_BATCH_LIMIT {
+                                    warn!("Dropping submitted batch, too many pending batches");
```
peterschwarz, line 76: We're not dropping it, we're rejecting the batch with the appropriate error.

peterschwarz (changes requested):

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -114,7 +183,9 @@ impl ScabbardShared {
             );
         }
 
-        batch
+        self.update_accepting_batches()?;
```
peterschwarz, line 186: Should this also be added to the `add_batch_to_queue` function?

agunde406: I reworked the solution a little bit to move the updating_accepting_batches into add_batch_to_queue/pop_batch_from_queue only

peterschwarz (approved):

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -95,15 +106,84 @@ impl ScabbardShared {
...
+    fn update_accepting_batches(&mut self) -> Result<(), ScabbardError> {
+        // only the coordinator should change accepting batches and
+        // back pressure is not supported by V1
+        if !self.is_coordinator() || self.scabbard_version == ScabbardVersion::V1 {
+            return Ok(());
+        };
+
+        if self.accepting_batches {
+            // check if batch queue has grown too large
+            if self.pending_batches_count() >= DEFAULT_PENDING_BATCH_LIMIT {
```
vaporos, line 132: Use of wrapper functions -- like pending_batches_count() -- requires the reader to reference the function unnecessarily to see the definition, then retain that definition in mind while reading the code.

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -95,15 +106,84 @@ impl ScabbardShared {
...
+    /// set whether we are accepting new batches
+    pub fn set_accepting_batches(&mut self, accepting: bool) {
+        self.accepting_batches = accepting;
+    }
+
+    /// Check whether the pending batch queue has either gotten too big and back pressure has
+    /// should be enabled.
+    /// If back pressure has been enabled, check to see if the queue has reduced in size enough
+    /// to start accepting batches again.
+    fn update_accepting_batches(&mut self) -> Result<(), ScabbardError> {
```
vaporos, line 123: There are two decisions coded within this function, but those decision points occur in different points of the code, and are unrelated in terms of the algorithm. This should be refactored.

For example, this function is called when adding a batch -- but then this function checks for a change in both increasing and decreasing batch counts. When adding a batch, it is unnecessary/incorrect to check for the decreasing case.

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -95,15 +106,84 @@ impl ScabbardShared {
...
     }
 
-    pub fn add_batch_to_queue(&mut self, batch: BatchPair) {
+    /// Returns the number of pending batches
+    pub fn pending_batches_count(&self) -> usize {
+        self.batch_queue.len()
+    }
+
+    /// set whether we are accepting new batches
+    pub fn set_accepting_batches(&mut self, accepting: bool) {
```
vaporos, line 115: If this does not need to be pub it should be removed and the boolean accessed directly.

libsplinter/src/orchestrator/mod.rs
```diff
@@ -29,7 +29,7 @@ use protobuf::Message;
 
 use crate::channel;
 use crate::error::InternalError;
-use crate::mesh::{Envelope, Mesh, RecvTimeoutError as MeshRecvTimeoutError};
+use crate::mesh::{Envelope, Mesh, RecvTimeoutError as MeshRecvTimeoutError, SendError};
```
vaporos, line 32: This fix seems unrelated, could go in another separate PR?

libsplinter/src/orchestrator/builder.rs
```diff
@@ -20,9 +20,9 @@ use crate::transport::Connection;
 
 use super::runnable::RunnableServiceOrchestrator;
 
-const DEFAULT_INCOMING_CAPACITY: usize = 8;
-const DEFAULT_OUTGOING_CAPACITY: usize = 8;
-const DEFAULT_CHANNEL_CAPACITY: usize = 8;
+const DEFAULT_INCOMING_CAPACITY: usize = 512;
```
vaporos, line 23: This commit is independent and could go into a separate PR.

vaporos (changes requested):

services/scabbard/libscabbard/src/service/shared.rs
agunde406: It does need to be pub, non-coordinators use this when receiving the TOO_MANY_REQUESTS/ACCEPTING_REQUESTS message from the coordinators

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -95,15 +106,84 @@ impl ScabbardShared {
...
+        // only the coordinator should change accepting batches and
+        // back pressure is not supported by V1
+        if !self.is_coordinator() || self.scabbard_version == ScabbardVersion::V1 {
+            return Ok(());
+        };
+
+        if self.accepting_batches {
+            // check if batch queue has grown too large
+            if self.pending_batches_count() >= DEFAULT_PENDING_BATCH_LIMIT {
+                self.set_accepting_batches(false);
```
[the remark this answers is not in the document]
agunde406: All of shared is wrapped in a lock.

services/scabbard/libscabbard/src/service/shared.rs
agunde406: I'll update so this code is directly in the add/pop batch queue functions instead

services/scabbard/libscabbard/src/service/shared.rs
agunde406: removed

services/scabbard/libscabbard/src/service/shared.rs
agunde406: updated, removes update_accepting_batches

libsplinter/src/orchestrator/mod.rs
agunde406: moved to https://github.com/Cargill/splinter/pull/1309

libsplinter/src/orchestrator/builder.rs
agunde406: moved to https://github.com/Cargill/splinter/pull/1309

vaporos:
Because this changes the protocol, we should upgrade the scabbard version.

We should also consider upgrading the REST API protocol version because the contract has changed.

vaporos:
This change should live behind an experimental feature so we can test it prior to stabilization.

agunde406:
I added the experimental features. As per our discussion, the protocol version decisions will be put off until stabilization

services/scabbard/libscabbard/src/service/mod.rs
```diff
@@ -67,7 +67,7 @@ const SERVICE_TYPE: &str = "scabbard";
 const DEFAULT_COORDINATOR_TIMEOUT: u64 = 30; // 30 seconds
 
 /// Specifies the version of scabbard to use.
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq)]
```
peterschwarz, line 70: Is this derive necessary? If it's only for tests, we might consider using something like

```rust
#[cfg_attr(test, derive(PartialEq))]
```

agunde406: its required by this check https://github.com/Cargill/splinter/pull/1306/files#diff-8407c21da2f6e6d430d3275579d5723b307d0dab6cb18fda50c9edee59cf9be6R139

peterschwarz (dismissed):

peterschwarz (approved):

vaporos (approved):

PR #1307, opened 2021-04-22
Biome profile stabilization update

isabeltomb:
Based on stabilization comments adds API doc comments to `ProfileBuilder` and adds a module doc comment to the store module

### DIFF
--- a/libsplinter/src/biome/profile/store/mod.rs
+++ b/libsplinter/src/biome/profile/store/mod.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Defines a basic representation of a profile.
+
 #[cfg(any(feature = "postgres", feature = "sqlite"))]
 pub(in crate::biome) mod diesel;
 pub mod error;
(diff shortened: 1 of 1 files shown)


agunde406 (approved):

libsplinter/src/biome/profile/store/mod.rs
```diff
@@ -70,7 +71,7 @@ impl Profile {
     }
 }
 
-/// Builder for profile.
+/// Builder for profile, user_id and subject are required
```
peterschwarz, line 76: I would suggest putting this on a separate line - it's a detail and the first line is the summary in the list of structs in the doc.

I would also add a note in the `with_` methods for the required items on the order of `This is a required field for the final Profile struct.`

peterschwarz (changes requested):

libsplinter/src/biome/profile/store/mod.rs
isabeltomb: Fixed

peterschwarz (approved):

agunde406 (approved):

PR #1312, opened 2021-04-26
Add wait_for method and node event client

leebradley:
Add wait_for convenience method, which allows a node to wait for a
particular admin event type rather than manually looping through every
event.

Add persistent node event client for nodes that will listen for all
event types and track the current event state. This event client is
currently only used by the wait_for convenience method.

### DIFF
--- a/libsplinter/src/admin/client/event/mod.rs
+++ b/libsplinter/src/admin/client/event/mod.rs
@@ -19,8 +19,11 @@ mod error;
 mod ws;
 
 use super::ProposalSlice;
+use std::cmp;
+use std::thread;
+use std::time::{Duration, Instant};
 
-pub use error::NextEventError;
+pub use error::{NextEventError, WaitForError};
 #[cfg(feature = "admin-service-event-client-actix-web-client")]
 pub use ws::actix_web_client::{
     AwcAdminServiceEventClient, AwcAdminServiceEventClientBuilder,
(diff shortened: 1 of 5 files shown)


libsplinter/src/admin/client/event/mod.rs
```diff
@@ -93,6 +94,15 @@ pub trait AdminServiceEventClient {
     ///
     /// This should block until an event is available.
     fn next_event(&self) -> Result<AdminServiceEvent, NextEventError>;
+
+    /// Wait for the given event up until a timeout
+    ///
+    /// This should block until an event is available.
+    fn wait_for(
+        &self,
+        event_type: &EventType,
```
peterschwarz, line 103: This argument needs to contain more detail then just the type.  More likely then not, the test will be waiting on a specific circuit change.

Perhaps an enum like
```rust
enum EventQuery {
    CircuitReady(String), // or a struct variant with circuit id
    CircuitRejected(String),
    CircuitDisbanded(String),
    // ... etc
    // The voting ones would be similar but include the PublicKey
}
```

libsplinter/src/admin/client/event/ws/actix_web_client.rs
```diff
@@ -303,6 +307,32 @@ impl AdminServiceEventClient for AwcAdminServiceEventClient {
         evt_result
             .map_err(|e| NextEventError::InternalError(InternalError::from_source(Box::new(e))))
     }
+
+    fn wait_for(
```
peterschwarz, line 311: This function is implementation-agnostic, so it can be pulled up to the trait as a default implementation. Any implementer that wants to override it can do so.

peterschwarz (changes requested):

libsplinter/src/admin/client/event/mod.rs
leebradley: That's a good point.

I created an `EventQuery` enum that's consumed by `wait_for` and uses the filters you list above. To cover all possible uses cases, `wait_for` calls a new function `wait_for_filter` that will accept any arbitrary closure to act as an event filter.

Under the hood, the `EventQuery` enum just transforms itself into an event filter closure.

libsplinter/src/admin/client/event/mod.rs
```diff
@@ -79,6 +82,54 @@ impl AdminServiceEvent {
...
+    ProposalVote { circuit_id: &'a str, key: PublicKey },
+    ProposalAccepted { circuit_id: &'a str, key: PublicKey },
+    ProposalRejected { circuit_id: &'a str, key: PublicKey },
+    CircuitReady { circuit_id: &'a str },
+    CircuitDisbanded { circuit_id: &'a str },
+}
+
+impl<'a> EventQuery<'a> {
+    /// Transform the EventQuery into an event filter
+    pub fn filter(self) -> impl Fn(&AdminServiceEvent) -> bool + 'a {
```
peterschwarz, line 97: This probably doesn't need to be public

libsplinter/src/admin/client/event/ws/actix_web_client.rs
```diff
@@ -159,6 +159,7 @@ impl RunnableAwcAdminServiceEventClient {
         let last_event_id = Arc::new(AtomicU64::new(last_event_id.unwrap_or(0)));
         let received_id = last_event_id.clone();
         let received_sender = event_sender.clone();
+
```
peterschwarz, line 162: As this is formatting related it should be part of a separate commit (or the change should be removed)

peterschwarz (dismissed):

shannynalayna (dismissed):

shannynalayna (approved):

peterschwarz (approved):

PR #1316, opened 2021-04-28
Retry messages whose connection_id does not have a peer_id

agunde406:
Before this commit, if a message was received whose connection id
did not have a correlating peer id, the peer interconnect would
drop this message. This however is possible if the sending
node finishes the peering processes before we do, causing an
important message, such as the admin protocol agreement message
to be dropped. If this message was sent at a later time it would
have been successful.

To handle this issues a message that is missing a peer id will
be sent to a separate thread and retried at a future time, giving
the peering processes a chance to complete. The message will
be given 3 chances for the peer id to be established before the
message will be dropped. A pacemaker will send a retry message
to this thread to trigger the checks.

To avoid this queue of pending messages to grow uncontrollably,
the queue is limited to 100 messages. If more then 100 messages
are in that queue, the oldest message will be dropped and
replaced with the new pending message.


libsplinter/src/peer/interconnect.rs
```diff
@@ -107,6 +117,16 @@ impl ShutdownHandle for PeerInterconnect {
...
     /// being joined.
     fn signal_shutdown(&mut self) {
+        self.pacemaker_shutdown_signaler.shutdown();
+
+        if self
+            .pending_shutdown_sender
+            .send(RetryIncoming::Shutdown)
+            .is_err()
+        {
+            warn!("Peer Interconnect is no longer running");
```
vaporos, line 127: This should probably have more context because this statement in the log context won't have much meaning.

agunde406: added more context

libsplinter/src/peer/interconnect.rs
```diff
@@ -258,6 +294,25 @@ where
...
+        let pending_recv_peer_lookup = peer_lookup_provider.peer_lookup();
+        let recv_pending_join_handle = thread::Builder::new()
+            .name("PeerInterconnect Pending Recv".into())
+            .spawn(move || {
+                if let Err(err) = run_pending_recv_loop(
+                    &*pending_recv_peer_lookup,
+                    pending_incoming_receiver,
+                    pending_network_dispatcher_sender,
+                ) {
+                    error!("Shutting down peer interconnect pending recevier: {}", err);
```
vaporos, line 306: Typo - receiver

vaporos (approved):

peterschwarz (approved):

vaporos (approved):

PR #1318, opened 2021-04-28
Add ability to remove a proposal within the admin service

shannynalayna:
This change adds a payload and handling to enable users to remove a proposal. This also adds tests around the validation of the payload and the process of submitting the payload. 

The process of removing a proposal includes (basically in order):
- the payload is validated
- a message is sent to the other proposed circuit members that another member has removed the proposal, but does not make any effect on other members' proposal record
- Removes the proposal from the requesting nodes' admin store
- updates the metrics to account for the change in proposals for the requesting node
- Removes the peer references associated with the proposal for the requesting node

For testing, you can checkout this branch: https://github.com/shannynalayna/splinter/tree/shannynalayna-proposal-removal-CLI, it adds a `splinter-circuit-remove-proposal` command that can be used to test the feature out.

### DIFF
--- a/libsplinter/src/admin/service/shared.rs
+++ b/libsplinter/src/admin/service/shared.rs
@@ -6498,13 +6498,15 @@ mod tests {
         shutdown(mesh, cm, pm);
     }
 
-    /// Tests that a request to abandon a circuit returns an error if the requester is not
-    /// permitted for the admin service.
+    /// Tests that a request to abandon a circuit is completed successfully, resulting in the
+    /// updated circuit status.
     ///
     /// 1. Set up `AdminServiceShared`
-    /// 2. Add a `Active` circuit to the admin store
-    /// 3. Call `validate_abandon_circuit` with the circuit and requester info
-    /// 4. Validate the call to `validate_abandon_circuit` returns an error
+    /// 2. Add an `Active` circuit to the admin store
+    /// 3. Create a `CircuitAbandon` message
+    /// 4. Submit the `CircuitAbandon` request to the node's admin service
+    /// 5. Fetch the circuit that was abandoned in the previous step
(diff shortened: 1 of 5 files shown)


libsplinter/src/admin/service/shared.rs
```diff
@@ -6456,6 +6456,7 @@ mod tests {
         circuit.set_comments("test circuit".to_string());
         circuit.set_display_name("test_display".into());
         circuit.set_circuit_status(admin::Circuit_CircuitStatus::ACTIVE);
+        circuit.set_circuit_version(2);
```
agunde406, line 7009: typo in commit message, verion -> version

libsplinter/src/admin/service/shared.rs
```diff
@@ -1070,6 +1072,68 @@ impl AdminServiceShared {
...
+                            &circuit_id
+                        )),
+                    ))
+                })?;
+            // Update the metrics because the proposal has been removed for this node
+            #[cfg(feature = "admin-service-count")]
+            self.update_metrics()
+                .map_err(|err| ServiceError::UnableToHandleMessage(Box::new(err)))?;
+
+            // Remove the peer refs belonging to the removed proposal proposal
```
agunde406, line 1122: "removed proposal proposal" -> double proposal

agunde406 (changes requested):

libsplinter/src/admin/service/shared.rs

libsplinter/src/admin/service/shared.rs
shannynalayna: Fixed

isabeltomb (approved):
Tested, looks good

agunde406 (approved):

PR #1319, opened 2021-04-30
Added user configurable logging

Caleb-Hill:
Added a "log" element in the Config struct. Logging can be tailored by
module to log different information, at different levels, to different
places. Currently stdOut, File, and Rotating File are supported for
logging targets.

### DIFF
--- a/splinterd/src/config/mod.rs
+++ b/splinterd/src/config/mod.rs
@@ -32,6 +32,8 @@ pub use crate::config::clap::ClapPartialConfigBuilder;
 pub use crate::config::default::DefaultPartialConfigBuilder;
 pub use crate::config::env::EnvPartialConfigBuilder;
 pub use crate::config::toml::TomlPartialConfigBuilder;
+#[cfg(feature = "log-config")]
+use crate::logging::{AppenderConfig, LoggerConfig, RootConfig};
 pub use builder::{ConfigBuilder, PartialConfigBuilder};
 pub use error::ConfigError;
 pub use partial::{ConfigSource, PartialConfig};
@@ -95,6 +97,14 @@ pub struct Config {
     metrics_username: Option<(String, ConfigSource)>,
     #[cfg(feature = "metrics")]
     metrics_password: Option<(String, ConfigSource)>,
+    #[cfg(feature = "log-config")]
+    root_logger: (RootConfig, ConfigSource),
+    #[cfg(feature = "log-config")]
+    appenders: Option<Vec<(AppenderConfig, ConfigSource)>>,
+    #[cfg(feature = "log-config")]
+    loggers: Option<Vec<(LoggerConfig, ConfigSource)>>,
+    #[cfg(feature = "log-config")]
+    verbosity: (log::Level, ConfigSource),
 }
 
 impl Config {
@@ -554,6 +564,40 @@ impl Config {
         }
     }
 
+    #[cfg(feature = "log-config")]
+    pub fn root_logger(&self) -> &RootConfig {
+        &self.root_logger.0
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn root_logger_source(&self) -> &ConfigSource {
+        &self.root_logger.1
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn appenders(&self) -> Option<Vec<AppenderConfig>> {
+        self.appenders
+            .as_ref()
+            .map(|configs| configs.iter().map(|c| c.0.clone()).collect())
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn loggers(&self) -> Option<Vec<LoggerConfig>> {
+        self.loggers
+            .as_ref()
+            .map(|configs| configs.iter().map(|c| c.0.clone()).collect())
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn verbosity(&self) -> log::Level {
+        self.verbosity.0
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn verbosity_source(&self) -> &ConfigSource {
+        &self.verbosity.1
+    }
+
     #[allow(clippy::cognitive_complexity)]
     /// Displays the configuration value along with where the value was sourced from.
     pub fn log_as_debug(&self) {
@@ -778,6 +822,29 @@ impl Config {
             {
                 debug!("Config: metrics_password: <HIDDEN> (source: {:?})", source,);
             }
+        }
+        #[cfg(feature = "log-config")]
+        {
+            let loggers = self.loggers.as_ref().unwrap_or(&vec![]).to_owned();
+            for logger in loggers {
+                debug!("Config: logger: {:?} (source: {:?})", logger.0, logger.1);
+            }
+            for appender in self.appenders.as_ref().unwrap_or(&vec![]).to_owned() {
+                debug!(
+                    "Config: appender: {:?} (source: {:?})",
+                    appender.0, appender.1
+                );
+            }
+            debug!(
+                "Config: root_logger: {:?} (source: {:?})",
+                self.root_logger(),
+                self.root_logger_source()
+            );
+            debug!(
+                "Config: verbosity: {:?} (source: {:?})",
+                self.verbosity(),
+                self.verbosity_source()
+            );
         }
     }
 
--- a/splinterd/src/config/partial.rs
+++ b/splinterd/src/config/partial.rs
@@ -15,7 +15,12 @@
 //! An intermediate representation of the configuration values, used to take the
 //! configuration values from different sources into a common representation.
 
+#[cfg(feature = "log-config")]
+use std::collections::HashMap;
 use std::time::Duration;
+
+#[cfg(feature = "log-config")]
+use crate::logging::{RootConfig, UnnamedAppenderConfig, UnnamedLoggerConfig};
 
 /// `ConfigSource` displays the source of configuration values, used to identify which of the various
 /// config modules were used to create a particular `PartialConfig` object.
@@ -88,6 +93,14 @@ pub struct PartialConfig {
     metrics_username: Option<String>,
     #[cfg(feature = "metrics")]
     metrics_password: Option<String>,
+    #[cfg(feature = "log-config")]
+    root_logger: Option<RootConfig>,
+    #[cfg(feature = "log-config")]
+    appenders: Option<HashMap<String, UnnamedAppenderConfig>>,
+    #[cfg(feature = "log-config")]
+    loggers: Option<HashMap<String, UnnamedLoggerConfig>>,
+    #[cfg(feature = "log-config")]
+    verbosity: Option<log::Level>,
 }
 
 impl PartialConfig {
@@ -150,6 +163,14 @@ impl PartialConfig {
             metrics_username: None,
             #[cfg(feature = "metrics")]
             metrics_password: None,
+            #[cfg(feature = "log-config")]
+            appenders: None,
+            #[cfg(feature = "log-config")]
+            loggers: None,
+            #[cfg(feature = "log-config")]
+            root_logger: None,
+            #[cfg(feature = "log-config")]
+            verbosity: None,
         }
     }
 
@@ -327,6 +348,23 @@ impl PartialConfig {
     #[cfg(feature = "metrics")]
     pub fn metrics_password(&self) -> Option<String> {
         self.metrics_password.clone()
+    }
+
+    #[cfg(feature = "log-config")]
+    pub fn appenders(&self) -> Option<HashMap<String, UnnamedAppenderConfig>> {
+        self.appenders.clone()
+    }
+    #[cfg(feature = "log-config")]
+    pub fn loggers(&self) -> Option<HashMap<String, UnnamedLoggerConfig>> {
+        self.loggers.clone()
+    }
+    #[cfg(feature = "log-config")]
+    pub fn root_logger(&self) -> Option<RootConfig> {
+        self.root_logger.clone()
+    }
+    #[cfg(feature = "log-config")]
+    pub fn verbosity(&self) -> Option<log::Level> {
+        self.verbosity
     }
 
     /// Adds a `config_dir` value to the `PartialConfig` object.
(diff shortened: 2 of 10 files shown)


splinterd/Cargo.toml
```diff
@@ -41,6 +41,7 @@ cylinder = "0.2"
 flexi_logger = "0.14"
 health = { path = "../services/health", optional = true }
 log = "0.4"
+log4rs = {version ="1.0", features = ["pattern_encoder","delete_roller","compound_policy","size_trigger","file_appender","toml_format","config_parsing"]}
```
vaporos, line 44: This should be optional for now and pulled in with the experimental feature.

Note the spacing is incorrect -- "{version" vs. "{ version" and speaces after commas, etc. to match the style used elsewhere.

Version should be set to "1" not "1.0" as they are functionally equivalent and we prefer the shorter.

Are all of these features needed explicitly? This is very strange.

splinterd/Cargo.toml
```diff
@@ -152,6 +154,9 @@ service-arg-validation = [
 ]
 service-endpoint = []
 ws-transport = ["splinter/ws-transport"]
+log-config = ["log-redirect","log-filter"]
+log-redirect = []
```
vaporos, line 158: log-filter and log-redirect are not used in this PR - please remove.

splinterd/src/config/mod.rs
```diff
@@ -799,6 +819,14 @@ impl Config {
                 debug!("Config: metrics_password: <HIDDEN> (source: {:?})", source,);
             }
         }
+        #[cfg(feature = "log-config")]
+        {
+            debug!(
+                "Config: enable_biome_credentials: {:?} (source: {:?})",
```
vaporos, line 825: This appears to be a cut-n-paste error

splinterd/src/config/toml.rs
```diff
@@ -16,6 +16,8 @@
 
 use crate::config::PartialConfigBuilder;
 use crate::config::{ConfigError, ConfigSource, PartialConfig};
+#[cfg(feature = "log-config")]
+use log4rs::config::RawConfig as LogConfig;
```
vaporos, line 20: Remove use of log4rs from within crate::config and define the appropriate structs explicitly. This is important for two reasons: 1) log4rs is fundamentally unrelated to configuration loading; and 2) the config file is a versioned API and changes to it need to be handled in a backward-compatible way -- so we can not introduce an external data structure as part of our API definition.

Then, write an Into trait allowing conversion of our LogConfig struct to log4rs's RawConfig. This should maybe go in a crate::log module.

splinterd/src/main.rs
```diff
@@ -25,8 +25,11 @@ mod error;
 mod routes;
 mod transport;
 
-use flexi_logger::{style, DeferredNow, LogSpecBuilder, Logger};
+use flexi_logger::{style, DeferredNow};
```
vaporos, line 28: flexi_logger should be completely disabled when log4rs is enabled.

splinterd/src/main.rs
```diff
@@ -443,25 +446,52 @@ fn main() {
...
-    log_spec_builder.module("tokio", log::LevelFilter::Warn);
-    #[cfg(feature = "https-bind")]
-    log_spec_builder.module("h2", log::LevelFilter::Warn);
-
-    Logger::with(log_spec_builder.build())
-        .format(log_format)
-        .log_target(flexi_logger::LogTarget::StdOut)
-        .start()
-        .expect("Failed to create logger");
+    #[cfg(feature = "log-config")]
```
vaporos, line 449: This doesn't seem to take clap's log level into account. Similarly, hyper, tokio, h2 log-levels aren't adjusted.

splinterd/src/main.rs
```diff
@@ -443,25 +446,52 @@ fn main() {
...
+                    Some(
+                        log4rs::init_raw_config(log_config.clone())
+                            .expect("Failed to create logger"),
+                    )
+                } else {
+                    None
+                }
+            }
+            Err(e) => {
+                println!("Could not create config: {}", e);
```
vaporos, line 470: Errors must be printed with eprintln! and should begin with "Error: ".

splinterd/src/main.rs
```diff
@@ -500,13 +530,18 @@ fn setup_metrics_recorder(config: &Config) -> Result<(), UserError> {
     Ok(())
 }
 
-fn start_daemon(matches: ArgMatches) -> Result<(), UserError> {
-    // get provided config file or search default location
-    let config_file = matches
+fn get_config_file<'a>(matches: &'a ArgMatches) -> &'a str {
```
vaporos, line 533: This change could be a separate commit prior to the larger commit.

splinterd/src/main.rs
vaporos, line 449: Switching to log4rs but maintaining the same functionality could be a commit prior to the larger config commit.

vaporos (changes requested):

splinterd/Cargo.toml
Caleb-Hill: fixed

splinterd/Cargo.toml
Caleb-Hill: fixed

splinterd/src/config/mod.rs
Caleb-Hill: fixed

splinterd/src/config/toml.rs
Caleb-Hill: fixed/updated

splinterd/src/main.rs
Caleb-Hill: fixed

splinterd/src/main.rs
Caleb-Hill: fixed

splinterd/src/main.rs
Caleb-Hill: fixed

splinterd/src/main.rs
Caleb-Hill: fixed

Caleb-Hill: fixed/refactored

splinterd/src/main.rs
Caleb-Hill: Changed to an expect to keep with the current behavior of stopping the app if configs are broken. Maybe it makes sense to start the app without logging but don't think its a good idea

splinterd/src/config/mod.rs
peterschwarz: This looks like it hasn't been fixed

peterschwarz (changes requested):
I wonder if the log config should also have a set of partial configs, where it can generate these from cli args or a toml file (like the general configuration). That would simplify the usage in main and have the config resolution occur when collapsing the partial configs.

Caleb-Hill:
> I wonder if the log config should also have a set of partial configs, where it can generate these from cli args or a toml file (like the general configuration). That would simplify the usage in main and have the config resolution occur when collapsing the partial configs.

While generally I think this is a good idea representing the log configs as command line args would be hugely complicated. Not undoable but really hard for anyone to use for sure. If you'd like me to refactor so most of the main.rs changes as they stand now are managed in partial.rs I can do that.

I fixed the wrong debug message thing, it won't let me comment on it for whatever reason.

splinterd/src/config/mod.rs
Caleb-Hill: Double super fixed

peterschwarz:
> While generally I think this is a good idea representing the log configs as command line args would be hugely complicated. Not undoable but really hard for anyone to use for sure. If you'd like me to refactor so most of the main.rs changes as they stand now are managed in partial.rs I can do that.

The intent isn't to represent all of the possible options in every config, but to use the same mechanism, such that adding other logging overrides to the CLI args would be simple.  I'm not requesting that all the configuration options are added to the CLI args.

Caleb-Hill:
> > While generally I think this is a good idea representing the log configs as command line args would be hugely complicated. Not undoable but really hard for anyone to use for sure. If you'd like me to refactor so most of the main.rs changes as they stand now are managed in partial.rs I can do that.
> 
> The intent isn't to represent all of the possible options in every config, but to use the same mechanism, such that adding other logging overrides to the CLI args would be simple. I'm not requesting that all the configuration options are added to the CLI args.

Yeah that makes sense, Ill update the PR once I make the change

splinterd/src/config/error.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[cfg(feature = "log-config")]
+use log4rs::config::runtime::ConfigErrors;
```
peterschwarz, line 16: Is this change still relevant?

splinterd/src/config/error.rs
```diff
@@ -21,12 +23,17 @@ use toml::de::Error as TomlError;
 #[derive(Debug)]
 /// General error type used during `Config` contruction.
 pub enum ConfigError {
-    ReadError { file: String, err: io::Error },
+    ReadError {
```
peterschwarz, line 26: This change is not mentioned in the commit message - could be a separate commit, or was added by accident.

splinterd/src/main.rs
```diff
@@ -460,33 +468,63 @@ fn main() {
...
-    let log_config = config.build(Root::builder().appender("stdout").build(log_level));
-    if let Ok(lc) = log_config {
-        let _handle = log4rs::init_config(lc);
-    } else {
-        unreachable!();
-        // log_config  should always return an Ok variant.
-        // Err is returned if multiple logger facades are registered which shouldn't ever happen
-    }
+            if let Ok(lc) = conf {
+                log4rs::init_config(lc).unwrap()
```
peterschwarz, line 519: Remove the unwrap calls, here and elsewhere

splinterd/src/main.rs
```diff
@@ -538,7 +577,7 @@ fn get_log_filter_level(matches: &ArgMatches) -> log::LevelFilter {
     }
 }
 
-fn start_daemon(matches: ArgMatches) -> Result<(), UserError> {
+fn start_daemon(matches: ArgMatches, _log_handle: Handle) -> Result<(), UserError> {
```
peterschwarz, line 580: Why the unused param?

Caleb-Hill: Its only used with experimental turned on and throws a warning otherwise

splinterd/src/config/error.rs
Caleb-Hill: Ahh yeah rustfmt  did it for me, Ill pull it out

splinterd/src/config/error.rs
Caleb-Hill: Doesn't look like it is, Ill take it out

splinterd/src/logging/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct AppenderConfig {
+    pub name: String,
+    #[serde(default = "default_pattern")]
+    pub encoder: String,
+    pub target: LogTarget,
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub enum LogTarget {
+    StdOut,
```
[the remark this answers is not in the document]
Caleb-Hill: added some aliases that seemed reasonable-ish

splinterd/src/main.rs
Caleb-Hill: No more just raw unwrap calls

splinterd/src/config/mod.rs
```diff
@@ -553,6 +563,37 @@ impl Config {
             None
         }
     }
+    #[cfg(feature = "log-config")]
+    pub fn root_logger(&self) -> &RootConfig {
+        &self.root_logger.0
+    }
+    #[cfg(feature = "log-config")]
```
vaporos, line 570: Missing a empty line above this line.

splinterd/src/config/partial.rs
```diff
@@ -785,4 +821,45 @@ impl PartialConfig {
...
+    /// Adds a `verbosity` value to the `PartialConfig` object.
+    ///
+    /// # Arguments
+    ///
+    ///  * `level` - The log level the system will use for the root_logger logger
+    pub fn with_verbosity(mut self, level: Option<log::Level>) -> Self {
+        self.verbosity = level;
+        self
+    }
+    #[cfg(feature = "log-config")]
```
vaporos, line 835: Missing a newline above this line (also, same with functions below in this chunk).

splinterd/src/logging/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    StdErr,
```
vaporos, line 37: s/StdErr/stderr/ and s/StdOut/Stdout/

splinterd/src/logging/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct LogConfigPartial {
+    root: Option<RootConfig>,
+    appenders: Option<Vec<AppenderConfig>>,
+    loggers: Option<Vec<LoggerConfig>>,
+}
+
+pub fn default_pattern() -> String {
```
vaporos, line 49: This should at most be a const declaration.

splinterd/src/logging/mod.rs
```diff
@@ -28,13 +36,20 @@ pub struct LogConfig {
 pub struct LoggerConfig {
     name: String,
     appenders: Vec<String>,
-    filter: Level,
+    level: Level,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Deserialize, Clone, Debug)]
+pub struct UnNamedLoggerConfig {
```
peterschwarz, line 43: It's unnamed, not un named, so remove the capital N

splinterd/src/config/toml.rs
```diff
@@ -195,6 +203,21 @@ impl PartialConfigBuilder for TomlPartialConfigBuilder {
                 .with_metrics_username(self.toml_config.metrics_username)
                 .with_metrics_password(self.toml_config.metrics_password)
         }
+        #[cfg(feature = "log-config")]
+        {
+            if let Some(mut loggers) = self.toml_config.loggers {
+                if let Some(un_named) = loggers.remove("root") {
```
peterschwarz, line 209: unnamed, not un_named

splinterd/src/main.rs
```diff
@@ -462,33 +469,65 @@ fn main() {
...
+                    encoder: String::from(DEFAULT_PATTERN),
+                    kind: LogTarget::Stdout,
+                }],
+                loggers: vec![],
+            };
+            log4rs::init_config(
+                //Always produces a valid config try_into_config just returns a Result
+                default_config
+                    .try_into()
+                    .expect("Default logging config failure"),
```
peterschwarz, line 492: It might be more clear to use an `unreachable!()` statement like below

splinterd/src/logging/mod.rs
```diff
@@ -28,13 +36,20 @@ pub struct LogConfig {
...
+    appenders: Vec<String>,
+    #[serde(alias = "filter")]
```
peterschwarz, line 45: Should we alias this?  What if we want to use "filter" for ignoring modules?

peterschwarz (approved):

splinterd/src/logging/mod.rs
Caleb-Hill: It needs to be a function for its use within serde, so I have made the function private and added a public const with the actual pattern for use outside of this file.

splinterd/src/logging/mod.rs
Caleb-Hill: fixed

splinterd/src/config/partial.rs
Caleb-Hill: fixed all the spacing issues

vaporos (approved):

PR #1320, opened 2021-05-01
Add file as default registry uri scheme

leebradley:
Improve command line user experience by allowing files to be passed as
registry arguments without requiring the "file://" prefix. This
requirement was not mentioned in the cli --help or in the resulting
error message.

Update message on supported registry uri schemes.


leebradley:
CI is failing on main branch as well, so it appears unrelated. Once main is fixed I'll rebase and kick off CI again.

splinterd/src/daemon.rs
```diff
@@ -1458,13 +1456,61 @@ fn create_registry(
...
+// Parses a registry argument, returning the uri scheme (defaulting to file) and remaining uri data
+fn parse_registry_arg(registry: &str) -> (&str, &str) {
     let mut iter = registry.splitn(2, "://");
-    let scheme = iter
+    let first = iter
         .next()
         .expect("str::split cannot return an empty iterator");
-    let path = iter.next().ok_or("No URI scheme provided")?;
-    Ok((scheme, path))
+    match iter.next() {
```
peterschwarz, line 1465: Non-blocking: Since you are updating this, it might make sense to remove the expect and follow this style: https://github.com/Cargill/splinter/blob/main/libsplinter/src/rest_api/auth/mod.rs#L177-L183  adding a case of 
```rust
    (None, None) => unreachable!()
```
makes it much more clear then the above expect.

leebradley: Done :+1:

vaporos (approved):

PR #1324, opened 2021-05-04
Fix test race conditions

peterschwarz:
This fixes four test race conditions:

- in the first case, it allows for multiple heartbeats to arrive before the expected message (within a set of attempts)
- In the second, it replaces the shared state + sleeps with a channel to receive the expected output from a separate thread.
- In the third, the admin service was not started before the admin event client attempts to connect. Waiting until it is in the started state fixed the issue (returning an error if it is not started in a reasonably long window)
- In the fourth, it fixes an issue with mesh, where connections could be added before the internal channel is registered

### DIFF
--- a/splinterd/src/node/runnable/admin.rs
+++ b/splinterd/src/node/runnable/admin.rs
@@ -19,7 +19,7 @@ use std::time::Duration;
 use cylinder::Verifier;
 use scabbard::service::ScabbardFactory;
 use splinter::admin::rest_api::CircuitResourceProvider;
-use splinter::admin::service::AdminServiceBuilder;
+use splinter::admin::service::{AdminCommands, AdminServiceBuilder, AdminServiceStatus};
 use splinter::circuit::routing::RoutingTableWriter;
 use splinter::error::InternalError;
 use splinter::events::Reactor;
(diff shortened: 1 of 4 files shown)


agunde406 (dismissed):

isabeltomb (dismissed):

rbuysse (approved):

splinterd/src/node/runnable/admin.rs
```diff
@@ -171,6 +177,25 @@ impl RunnableAdminSubsystem {
...
+            let status = commands
+                .admin_service_status()
+                .map_err(|e| InternalError::from_source(Box::new(e)))?;
+
+            if status == AdminServiceStatus::Running {
+                break;
+            }
+            if attempts.next().is_none() {
+                return Err(InternalError::with_message(format!(
+                    "Admin service has completely started after {} secs",
```
vaporos, line 192: Is the intended message here that the admin service hasn't started?

vaporos (changes requested):

splinterd/src/node/runnable/admin.rs
peterschwarz: Fixed

vaporos (approved):

agunde406 (approved):

PR #1326, opened 2021-05-04
Add `splinter-circuit-remove-proposal` subcommand

shannynalayna:
This PR adds the `splinter-circuit-remove-proposal` subcommand to the Splinter CLI. This enables users to submit a `ProposalRemoveRequest` to the admin service to delete a circuit proposal.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -74,6 +74,7 @@ experimental = [
     "health",
     "https-certs",
     "permissions",
+    "proposal-removal",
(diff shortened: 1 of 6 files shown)


rbuysse (approved):

isabeltomb (commented):
Non blocking comment, this can be done in a separate PR but the the `splinter-circuit-propose` man page should probably be updated to reference this new command

isabeltomb (approved):

PR #1329, opened 2021-05-05
0.4 Revert transact version to 0.2 instead of 0.3 in scabbard cli

agunde406:
The API for SmartContractArchive::from_scar_file changed
between the version enforcing different requirements of
the the used scar files.

To stay compatible with 0.4.1 splinter, the version is
moved back to 0.2. in the scabbard cli. This is does however
result in a different version of the transact error being
returned by the sabre sdk builders that are used to created
the transactions. As such the errors must be wrapped into
a different type.

This changes allows these steps https://www.splinter.dev/docs/0.4/tutorials/configuring_splinter_nodes.html#uploading-a-smart-contract to work correctly

### DIFF
--- a/services/scabbard/cli/src/main.rs
+++ b/services/scabbard/cli/src/main.rs
@@ -778,8 +778,20 @@ fn run() -> Result<(), CliError> {
                     .with_contract(smart_contract.contract)
                     .into_payload_builder()?
                     .into_transaction_builder(&*signer)?
-                    .into_batch_builder(&*signer)?
-                    .build(&*signer)?;
+                    .into_batch_builder(&*signer)
+                    .map_err(|err| {
+                        CliError::action_error_with_source(
+                            "Unable to get batch builder for CreateContractAction",
+                            Box::new(err),
+                        )
+                    })?
+                    .build(&*signer)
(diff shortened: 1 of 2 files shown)


services/scabbard/cli/src/main.rs
```diff
@@ -778,8 +778,20 @@ fn run() -> Result<(), CliError> {
                     .with_contract(smart_contract.contract)
                     .into_payload_builder()?
                     .into_transaction_builder(&*signer)?
-                    .into_batch_builder(&*signer)?
-                    .build(&*signer)?;
```
peterschwarz, line 782: Doesn't this mean there is already a From implementation?

agunde406: Yes but that is for the 0.2 version of transact error which is used above in SmartContractArchive. This is the 0.3 transact error because that's the version sabre sdk is using.

peterschwarz: Oh, I see.  Makes sense now.

peterschwarz (approved):

vaporos (approved):

PR #1337, opened 2021-05-10
Add integration tests for disbanding while a node is stopped

shannynalayna:
This adds previously designed integration tests for disbanding a circuit, specifically when the non-requesting node is stopped throughout the process. This validates that the nodes are able to continuously communicate as expected despite restarts. Currently, when the integration framework nodes are restarted, their endpoints change. The tests added are ignored because of this bug, which will be addressed in the future. 

This also makes a small update to the `admin_service_event_client` to add an optional `last_event_id` argument when instantiating a new admin service event client to allow for nodes to catch up after being restarted.

### DIFF
--- a/splinterd/src/node/runnable/mod.rs
+++ b/splinterd/src/node/runnable/mod.rs
@@ -159,6 +159,7 @@ impl RunnableNode {
                     format!("http://localhost:{}", rest_api_port),
                     "foo".to_string(),
                     "*".to_string(),
+                    None,
                 )
                 .map_err(|e| InternalError::from_source(Box::new(e)))?,
         );
--- a/splinterd/src/node/running/admin.rs
+++ b/splinterd/src/node/running/admin.rs
@@ -54,6 +54,7 @@ impl AdminSubsystem {
         splinter_url: String,
         authorization: String,
         event_type: String,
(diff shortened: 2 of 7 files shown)


isabeltomb (dismissed):

splinterd/tests/admin/submit.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use splinter::admin::client::ProposalSlice;
+use splinterd::node::Node;
+
+use crate::admin::payload::{make_circuit_disband_payload, make_circuit_proposal_vote_payload};
+use splinter::error::InternalError;
+
+// Propose a circuit disband from the specified node. The payload is returned as it is used again
+// in some tests.
+pub(super) fn propose_disband(node: &Node, circuit_id: &str) -> Result<Vec<u8>, InternalError> {
```
vaporos, line 25: These super granular wrappers will make it harder to modify/maintain the code, and my preference would be to avoid them.

shannynalayna: Fair, removed the commit that added this module

isabeltomb (dismissed):

isabeltomb (approved):

agunde406 (approved):

PR #1349, opened 2021-05-14
Add protocols for proto::network messages 

agunde406:
There are some instance around the dispatchers that still require
the protobuf. Also some instance of the protos are left
until the circuit protocols are implemented. They will be
updated in a future commit.

### DIFF
--- a/libsplinter/src/protocol/network.rs
+++ b/libsplinter/src/protocol/network.rs
@@ -1,0 +1,120 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


Caleb-Hill (dismissed):

libsplinter/src/protocol/network.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use crate::protos::network;
+use crate::protos::prelude::*;
+
+/// The network message envelope
+#[derive(Debug)]
+pub enum NetworkMessage {
+    NetworkEcho(NetworkEcho),
+    NetworkHeartbeat(NetworkHeartbeat),
+    Circuit(Vec<u8>),
```
peterschwarz, line 26: Is this `Vec<u8>` because we don't have the protocol structs for Circuit messages yet?

agunde406: Yes, once the circuit ones are defined the Vec<u8> will be replaced with the CircuitMessage protocol.

peterschwarz (dismissed):

peterschwarz (approved):

vaporos (approved):

PR #1360, opened 2021-05-24
Add v1 Trust Peer Authorization

agunde406:
This PR  includes the message handlers for the new authorization
message, as well as separating out the authorization state machine into
several state machines to make room for new actions/states in the new
trust procedure. This will make it easier in the future to add a
new state machine for challenge authorization.

The v0 Trust is still supported along with v1. If a ConnectRequest
is the first message received v0 trust will be used. If
AuthProtocolRequest is the first message v1 Trust will be used.
This is required to stay backwards compatible with 0.4.

It is also now required to track both our local authorization state
and the state of the remote as separate state machines. This is
because authorization will be happening in parallel.

The new implementation is behind a new experimental features
`trust-authorization`. The `challenge-authorization` features
now depends on this new feature.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -134,6 +134,7 @@ experimental = [
     "rest-api-actix-web-3",
     "service-arg-validation",
     "service-network",
+    "trust-authorization",
     "ws-transport",
     "zmq-transport",
 ]
@@ -194,6 +195,7 @@ service-arg-validation = []
 service-network = []
 sqlite = ["diesel/sqlite", "diesel_migrations"]
 store-factory = []
+trust-authorization = []
 ws-transport = ["tungstenite"]
 zmq-transport = ["zmq"]
 
--- a/libsplinter/src/network/auth/handlers/v1_handlers.rs
+++ b/libsplinter/src/network/auth/handlers/v1_handlers.rs
@@ -1,0 +1,456 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Message handlers for v1 authorization messages
+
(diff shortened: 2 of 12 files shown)


libsplinter/src/protocol/authorization.rs
```diff
@@ -229,8 +229,8 @@ pub struct AuthComplete;
 /// returned.
 #[derive(Debug)]
 pub struct AuthProtocolRequest {
-    auth_protocol_min: i32,
-    auth_protocol_max: i32,
+    pub auth_protocol_min: u32,
+    pub auth_protocol_max: u32,
```
shannynalayna, line 233: Should these `pub` updates be in the next commit?

agunde406: updated the commits to separate adding the pub and the change to u32.

libsplinter/Cargo.toml
```diff
@@ -161,7 +162,7 @@ biome-credentials = ["bcrypt", "biome"]
 biome-key-management = ["biome"]
 biome-notifications = ["biome"]
 biome-profile = ["biome"]
-challenge-authorization = []
+challenge-authorization = ["trust-authorization"]
```
peterschwarz, line 165: Why is this dependency needed in this PR?

agunde406: I guess it is not needed yet, but will be as soon as the challenge authorization implementation is added. Ill remove it form this PR.

agunde406: updated

libsplinter/src/network/auth/handlers/v1_handlers.rs
```diff
@@ -0,0 +1,456 @@
...
+            }
+            Ok(next_state) => panic!("Should not have been able to transition to {}", next_state),
+        }
+        Ok(())
+    }
+}
+
+/// Return the supported protocol version that matches the min/max provided. If there is no
+/// matching protocol version 0 is returned.
+fn supported_protocol_version(min: u32, max: u32) -> u32 {
```
Caleb-Hill, line 154: Should this return a Result instead of an int with a magic value?

peterschwarz: Probably, but I think this matches how these protocol version matches are checked elsewhere in the code base.  We should add a story to make that sort of conversion, but it shouldn't block this PR

peterschwarz (approved):

libsplinter/src/network/auth/handlers/v1_handlers.rs
agunde406: I'll add a story.

Caleb-Hill (approved):

shannynalayna (approved):

PR #1361, opened 2021-05-24
Make crates publishing more reliable

rbuysse:
The amount of time it takes for a published crate to appear in the
index varies significantly. Replacing the finite sleep with a while
loop ensures that the latest version of Splinter is available as a
dependency before compiling libscabbard for publishing.

### DIFF
--- a/ci/publish-splinter-crates
+++ b/ci/publish-splinter-crates
@@ -49,11 +49,18 @@ ENV PATH=$PATH:/root/.cargo/bin
 WORKDIR /project/splinter/libsplinter
 
 CMD cargo login $CARGO_CRED \
+ && echo "Publshing version $REPO_VERSION" \
  && rm -f ../Cargo.lock ./Cargo.lock \
  && cargo clean \
  && cargo test \
  && cargo publish \
- && sleep 10 \
+ && bash -c '\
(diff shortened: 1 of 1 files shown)


agunde406 (approved):

ci/publish-splinter-crates
```diff
@@ -49,11 +49,18 @@ ENV PATH=$PATH:/root/.cargo/bin
...
 CMD cargo login $CARGO_CRED \
+ && echo "Publshing version $REPO_VERSION" \
  && rm -f ../Cargo.lock ./Cargo.lock \
  && cargo clean \
  && cargo test \
  && cargo publish \
- && sleep 10 \
+ && bash -c '\
+    while true; do echo "Waiting for Splinter publishing to complete"; \
+    cargo search splinter | grep "splinter = " | grep -q $REPO_VERSION; \
```
peterschwarz, line 59: Will this timeout at any point?

rbuysse: No, but given that releases are a manual activity, publishing is generally supervised so an operator can cancel it if necessary.

vaporos (approved):

peterschwarz (approved):

PR #1362, opened 2021-05-25
Update keygen --system to check for env variable

agunde406:
If --system is passed to keygen, the command will
check the SPLINTER_CONFIG_DIR and SPLINTER_HOME environment
variable to determine where the key should be generated.
The key will be added in one of the following locations:

SPLINTER_CONFIG_DIR/keys
SPLINTER_HOME/etc/keys
/etc/splinter/keys

### DIFF
--- a/cli/man/splinter-keygen.1.md
+++ b/cli/man/splinter-keygen.1.md
@@ -94,6 +94,19 @@ writing file: "/etc/splinter/keys/splinterd.priv"
 writing file: "/etc/splinter/keys/splinterd.pub"
 ```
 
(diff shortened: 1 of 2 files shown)


vaporos (approved):

peterschwarz (commented):
`--system` now places the key in the correct place, but still names the keys with the user name - is the desired effect? Also, how does this command now compare to the other command `splinter admin keygen`?

peterschwarz (dismissed):

peterschwarz (approved):

shannynalayna (approved):

PR #1363, opened 2021-05-25
Add the splinter upgrade command

leebradley:
Add "splinter upgrade" command that imports data from a yaml state store
to a database.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -51,6 +51,7 @@ features = ["admin-service", "registry"]
 [dev-dependencies]
 gag = "0.1"
 serial_test = "0.3"
+tempdir = "0.3"
 
 [features]
 default = [
@@ -72,6 +73,7 @@ experimental = [
     "authorization-handler-rbac",
     "challenge-authorization",
     "health",
+    "upgrade",
     "https-certs",
     "permissions",
     "proposal-removal",
@@ -86,6 +88,8 @@ circuit-template = ["splinter/circuit-template"]
 registry = []
 
 health = []
+
+upgrade = ["database", "splinter/store-factory"]
 
 https-certs = []
 permissions = []
--- a/cli/src/action/database/mod.rs
+++ b/cli/src/action/database/mod.rs
@@ -15,7 +15,12 @@
 #[cfg(feature = "sqlite")]
 mod sqlite;
 
+#[cfg(feature = "upgrade")]
+mod upgrade;
+
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::{env, fs};
 
 use clap::ArgMatches;
 use diesel::{connection::Connection as _, pg::PgConnection};
@@ -29,6 +34,9 @@ use self::sqlite::{get_default_database, sqlite_migrations};
 
 use super::Action;
 
+const SPLINTER_HOME_ENV: &str = "SPLINTER_HOME";
+const SPLINTER_STATE_DIR_ENV: &str = "SPLINTER_STATE_DIR";
+const DEFAULT_STATE_DIR: &str = "/var/lib/splinter";
 pub struct MigrateAction;
 
 impl Action for MigrateAction {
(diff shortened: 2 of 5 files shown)


cli/src/action/database/mod.rs
```diff
@@ -63,6 +77,227 @@ impl Action for MigrateAction {
...
+        get_db_store: impl FnOnce() -> Result<Box<dyn AdminServiceStore>, CliError>,
+    ) -> Result<(), CliError> {
+        writeln!(stdout, "Upgrading splinterd state")?;
+
+        write!(stdout, "Loading YAML datastore... ")?;
+        fn invalid_utf8() -> CliError {
+            CliError::ActionError("'state_dir' is not a valid UTF-8 string".to_string())
+        }
+
+        let circuits_location = Path::new(&state_dir).join("circuits.yaml");
```
agunde406, line 128: The state files can be added as const above.

cli/src/action/database/mod.rs
```diff
@@ -63,6 +77,227 @@ impl Action for MigrateAction {
...
+        };
+
+        let state_dir = match state_dir {
+            Some(state_dir) => state_dir.to_string(),
+            None => Self::env_state_dir()?,
+        };
+
+        Self::import_command(&mut io::stdout(), &state_dir, || {
+            // Verify database connection before setting up pool
+            PgConnection::establish(&database_uri[..]).map_err(|err| {
```
agunde406, line 287: Commit message says that only Postgress is supported, however this command should support sqlite as well.

cli/src/main.rs
```diff
@@ -948,7 +948,25 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
                         ),
                 ),
-        )
+        );
+
+        app = app.subcommand(
+            SubCommand::with_name("upgrade")
+                .about("Upgrade splinter")
+                .arg(
+                    Arg::with_name("state_dir")
```
agunde406, line 957: Please add `.value_name("state-dir")`

cli/src/action/database/mod.rs
```diff
@@ -23,12 +23,26 @@ use diesel::{connection::Connection as _, pg::PgConnection};
 use splinter::migrations::run_postgres_migrations;
 
 use crate::error::CliError;
+use splinter::admin::store::error::AdminServiceStoreError;
+use splinter::admin::store::yaml::YamlAdminServiceStore;
+use splinter::admin::store::AdminServiceStore;
+use std::env;
```
agunde406, line 29: Please rearrange your use statements into blocks in the order std, external crate, internal crate. We try to standardize this format for the use statements.

peterschwarz: With a new line in between, so the formatter doesn't resort them.

This is one thing we can't (currently) catch with the linter, but is a team preference.

leebradley:
Made changes in response to feedback.

Major changes:
- Added experimental feature `upgrade`
   - To facilitate this, I moved the importer to its own module under `cli::action::database::upgrade`
- Split out commit for `SplinterEnvironment`
- Added support for sqlite

Other changes:
- Replaced instances of `write!` and `writeln!` with `info!`
  - Because there's no easy way to capture or validate the output of the `info!` macro, I modified the unit test to validate using database output instead of the command output
- Added value-name of "state-dir" for state argument
- Added more information to argument help
- Moved some static values to consts

I also found that because `YamlAdminServiceStore` creates invalid circuit objects (apparently they lack endpoints, and this is by design?), it corrupts the data in a way that makes reading those circuits impossible. I work around this by filling in those endpoints during the import process.

This is ready for another round of review.

cli/Cargo.toml
```diff
@@ -46,11 +46,12 @@ whoami = "0.7.0"
```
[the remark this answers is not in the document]
leebradley: Fixed. Put everything behind an `update` feature

cli/src/action/database/mod.rs
```diff
@@ -63,6 +77,227 @@ impl Action for MigrateAction {
...
+            })?;
+        Ok(store_factory.get_admin_service_store())
+    }
+
+    fn import_command(
+        stdout: &mut dyn Write,
+        state_dir: &str,
+        get_db_store: impl FnOnce() -> Result<Box<dyn AdminServiceStore>, CliError>,
+    ) -> Result<(), CliError> {
+        writeln!(stdout, "Upgrading splinterd state")?;
```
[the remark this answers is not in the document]
leebradley: Fixed. I'd initially done this for testing the console output. Unfortunately I can't really find a good way to capture the output using log macros `info!` and whatnot, but it's probably not a big deal.

cli/src/action/database/mod.rs
leebradley: Done!

cli/src/action/database/sqlite.rs
```diff
@@ -70,48 +67,20 @@ pub fn sqlite_migrations(connection_string: String) -> Result<(), CliError> {
...
 
-/// Returns the path to the default sqlite database
+/// Creates and returns the path to the default sqlite database
 ///
-/// If `SPLINTER_STATE_DIR` is set, returns `SPLINTER_STATE_DIR/splinter_state.db`.
-/// If `SPLINTER_HOME` is set, returns `SPLINTER_HOME/data/splinter_state.db`.
-/// Otherwise, returns `/var/lib/splinter/splinter_state.db`
+/// Gets the splinter default state path, creating it if it does not exist. Creates a db file with
+/// the name splinter_state.db.
 pub fn get_default_database() -> Result<String, CliError> {
```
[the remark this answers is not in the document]
leebradley: Done. I split this out into another commit.

cli/src/main.rs
```diff
@@ -948,7 +948,25 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
                                 .help("Database connection URI"),
                         ),
                 ),
-        )
+        );
+
+        app = app.subcommand(
+            SubCommand::with_name("upgrade")
+                .about("Upgrade splinter")
```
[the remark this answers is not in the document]
leebradley: Updated to now say "Upgrade splinter by importing YAML state files to a database"

cli/src/main.rs
leebradley: Done

cli/src/action/database/mod.rs
leebradley: Fixed!

cli/src/main.rs
```diff
@@ -951,6 +951,27 @@ fn run<I: IntoIterator<Item = T>, T: Into<OsString> + Clone>(args: I) -> Result<
...
 
+    #[cfg(feature = "upgrade")]
+    {
+        app = app.subcommand(
+            SubCommand::with_name("upgrade")
+                .about("Upgrade splinter by importing YAML state files to a database")
+                .arg(
+                    Arg::with_name("state_dir")
+                        .value_name("state-dir")
+                        .short("S")
```
agunde406, line 962: Can you also add `.long("state-dir")` here and `.long("connect")` below

cli/src/action/database/upgrade.rs
```diff
@@ -0,0 +1,493 @@
...
+            result.nodes, result.proposals
+        );
+
+        Ok(())
+    }
+}
+
+impl Action for ImportFromYamlAction {
+    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+        let (state_dir, database_uri) = match arg_matches {
```
agunde406, line 193: If I provide a state-dir but not a database uri (connect value) I expect that database to be in the same state dir as what I provided. This does not happen now, instead it puts the database in the default location.

agunde406: It would also be useful to log the state directory being used and what database the state is being moved to for debugging purposes. (with absolute paths for files)

agunde406 (changes requested):

agunde406:
Also we need to write a man page for this command but that can go in another PR

cli/src/action/database/upgrade.rs
leebradley: Fixed!
- Code now looks for the database in the state directory if sqlite feature is enabled
- `splinter upgrade` now shows what directory it's using for the state and what database uri it's using (with absolute paths)

cli/src/main.rs
leebradley: Done

cli/src/action/database/upgrade.rs
```diff
@@ -0,0 +1,492 @@
...
+        info!("Backing up state files... ");
+        let new_circuits_location = circuits_location.with_extension("yaml.old");
+        let new_proposals_location = proposals_location.with_extension("yaml.old");
+        fs::rename(circuits_location, new_circuits_location)
+            .map_err(|e| CliError::ActionError(format!("could not move circuits file: {}", e)))?;
+        fs::rename(proposals_location, new_proposals_location)
+            .map_err(|e| CliError::ActionError(format!("could not move proposals file: {}", e)))?;
+
+        info!(
+            "Successfully imported {} node(s) and {} proposal(s)",
```
agunde406, line 171: Should nodes be circuits (in the log and in the import results)?

leebradley: Yep! Fixed

agunde406 (approved):

vaporos (approved):

PR #1366, opened 2021-05-27
Add `oauth/users` endpoint to list OAuth users

shannynalayna:
This adds a rest endpoint to the `OAuthResourceProvider`, `oauth/users` behind the `oauth-user-list` feature, which lists OAuth users.

To test, run `cargo test --features experimental test_list_oauth_users`

### DIFF
--- a/libsplinter/src/biome/oauth/store/mod.rs
+++ b/libsplinter/src/biome/oauth/store/mod.rs
@@ -431,3 +431,10 @@ impl Iterator for OAuthUserIter {
         (self.inner.len(), Some(self.inner.len()))
     }
 }
+
+#[cfg(feature = "oauth-user-list")]
+impl ExactSizeIterator for OAuthUserIter {
+    fn len(&self) -> usize {
+        self.inner.len()
+    }
+}
--- a/libsplinter/src/protocol/mod.rs
+++ b/libsplinter/src/protocol/mod.rs
@@ -73,6 +73,8 @@ pub(crate) const OAUTH_CALLBACK_MIN: u32 = 1;
 pub(crate) const OAUTH_LOGIN_MIN: u32 = 1;
 #[cfg(all(feature = "oauth", feature = "rest-api-actix"))]
 pub(crate) const OAUTH_LOGOUT_MIN: u32 = 1;
+#[cfg(all(feature = "oauth-user-list", feature = "rest-api-actix"))]
+pub(crate) const OAUTH_USER_READ_PROTOCOL_MIN: u32 = 1;
 
 #[cfg(feature = "registry")]
 pub const REGISTRY_PROTOCOL_VERSION: u32 = 1;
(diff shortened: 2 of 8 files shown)


splinterd/api/static/openapi.yaml
```diff
@@ -2534,6 +2592,18 @@ components:
...
+    OAuthUser:
+      type: object
+      properties:
+        subject:
+          type: string
+          description: "Oauth provider's unique identifier for the user"
+          example: "129367690558975733497"
+        user_id:
+          type: string
+          description: "Internal unique identifier for the user"
```
agunde406, line 2604: Instead of "Internal" should this say `Biome`?

shannynalayna: This terminology is used throughout the Biome endpoints, which return the equivalent Biome user_id. But, perhaps it would make sense in this instance to add `Biome` to the description since the ID isn't as obviously related to Biome in this endpoint. updating ...

shannynalayna: Updated to say `Biome's internal unique identifier for the user`

libsplinter/src/biome/oauth/store/mod.rs
```diff
@@ -417,6 +417,10 @@ impl OAuthUserIter {
             inner: Box::new(users.into_iter()),
         }
     }
+
+    pub fn total(&self) -> usize {
```
peterschwarz, line 421: Perhaps this iterator should just implement `ExactSizeIterator`, like it's inner value.

libsplinter/src/protocol/mod.rs
```diff
@@ -73,6 +73,8 @@ pub(crate) const OAUTH_CALLBACK_MIN: u32 = 1;
 pub(crate) const OAUTH_LOGIN_MIN: u32 = 1;
 #[cfg(all(feature = "oauth", feature = "rest-api-actix"))]
 pub(crate) const OAUTH_LOGOUT_MIN: u32 = 1;
+#[cfg(all(feature = "oauth", feature = "oauth-user-list"))]
```
peterschwarz, line 76: this should include `feature = "rest-api-actix"`

peterschwarz (changes requested):

libsplinter/src/protocol/mod.rs
shannynalayna: Updated this guard to use the `oauth-user-list` and `rest-api-actix` features, as the `oauth-user-list` feature is dependent on `oauth` so this was unnecessary.

libsplinter/src/biome/oauth/store/mod.rs
shannynalayna: Updated this commit to implementing the `ExactSizeIterator` trait for the `OAuthUserIter` struct.

peterschwarz (approved):

agunde406 (approved):

PR #1370, opened 2021-06-01
Deprecate splinterd yaml state in favor of db

leebradley:
Add a feature "deprecate-yaml" that deprecates yaml state in splinterd
in favor of using sqlite or postgresql. If splinterd is run with yaml
state, it will abort and point the user to the "splinter upgrade"
command, which imports the yaml state into the specified database.


splinterd/src/daemon.rs
```diff
@@ -159,40 +160,78 @@ impl SplinterDaemon {
...
+        #[cfg(feature = "deprecate-yaml")]
+        {
+            let circuits_location = Path::new(&self.state_dir).join("circuits.yaml");
+            let proposals_location = Path::new(&self.state_dir).join("circuit_proposals.yaml");
+
+            let circuits_location_exists = circuits_location.exists();
+            let proposals_location_exists = circuits_location.exists();
+
+            if circuits_location_exists || proposals_location_exists {
+                error!("splinterd found yaml state files. run the `splinter upgrade` command to import your yaml state into a database");
```
vaporos, line 172: Suggest changing to "Found yaml state files; run the `splinter upgrade` command to import your yaml state into into a database"

splinterd/src/daemon.rs
```diff
@@ -159,40 +160,78 @@ impl SplinterDaemon {
...
+                }
+
+                if proposals_location_exists {
+                    error!(
+                        "yaml proposals state file: {}",
+                        proposals_location.to_string_lossy()
+                    );
+                }
+
+                panic!("aborting");
```
vaporos, line 188: Do not panic here, instead exit with error code 1.

vaporos: Correction, return StarError here.

splinterd/src/daemon.rs
```diff
@@ -159,40 +160,78 @@ impl SplinterDaemon {
...
-                            })?
-                            .to_string();
-
-                        Box::new(
-                            YamlAdminServiceStore::new(circuits_location, proposals_location)
-                                .map_err(|err| {
+                        #[cfg(feature = "deprecate-yaml")]
+                        {
+                            error!("splinterd no longer supports yaml state");
+                            panic!("aborting");
```
vaporos, line 201: Return StartError here.

Or consider removing this storage option completely.

splinterd/src/daemon.rs
leebradley: I've changed this to return StartError which propagates up to the main() function and exits with an error code 1. My concern with doing a process exit here was just making sure that the resources were dropped, but I wasn't sure how to weigh that against the error output the user would see. Let me know if I should change it to something else.

splinterd/src/daemon.rs
leebradley: Ah yeah, I think removing the option entirely is probably the cleaner solution

splinterd/src/daemon.rs
leebradley: @vaporos I've also changed this code to attempt to convert the directory into an absolute path, similar to what @agunde406  suggested for the upgrade command PR

vaporos (approved):

splinterd/src/daemon.rs
```diff
@@ -163,11 +166,45 @@ impl SplinterDaemon {
...
+                    error!(
+                        "Found outdated proposals state file: {}",
+                        fs::canonicalize(&proposals_location)
+                            .unwrap_or(proposals_location)
+                            .to_string_lossy()
+                    );
+                }
+
+                return Err(StartError::StorageError(
+                    format!("Run the `splinter upgrade` command to update outdated state files to a Splinter {}.{} database", env!("CARGO_PKG_VERSION_MAJOR", "unknown"), env!("CARGO_PKG_VERSION_MINOR", "unknown")),
```
agunde406, line 197: Please wrap this line at 100 characters

agunde406 (changes requested):

splinterd/src/daemon.rs
leebradley: Fixed!

leebradley:
@rbuysse pointed out that our toml configuration needs to be updated to no longer reference yaml state. I've removed references to storage option in the example configuration files since it's scheduled for deprecation and it's not really configurable.

peterschwarz:
The config issue is addressed in #1383

peterschwarz (approved):

agunde406 (approved):

PR #1372, opened 2021-06-02
Change logging library to log4rs

Caleb-Hill:
No change to functionality from flexilogger just a new facade for
logging. This is the first step to allowing user configuration of
logging levels and targets.

### DIFF
--- a/splinterd/Cargo.toml
+++ b/splinterd/Cargo.toml
@@ -38,9 +38,9 @@ clap = "2.32"
 crossbeam-channel = "0.3"
 ctrlc = "3.0"
 cylinder = "0.2.2"
-flexi_logger = "0.14"
 health = { path = "../services/health", optional = true }
 log = "0.4"
+log4rs = "1"
 openssl = { version = "0.10", optional = true }
 protobuf = "2.23"
 rand = "0.7"
--- a/splinterd/src/main.rs
+++ b/splinterd/src/main.rs
@@ -27,8 +27,9 @@ mod transport;
 
 #[cfg(feature = "challenge-authorization")]
 use cylinder::{load_key_from_path, secp256k1::Secp256k1Context, Context, Signer};
-use flexi_logger::{style, DeferredNow, LogSpecBuilder, Logger};
-use log::Record;
+use log4rs::config::{Appender, Logger, Root};
+use log4rs::encode::pattern::PatternEncoder;
+
 use rand::{thread_rng, Rng};
 #[cfg(feature = "metrics")]
 use splinter::metrics::influx::InfluxRecorder;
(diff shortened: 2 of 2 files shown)


splinterd/src/main.rs
```diff
@@ -485,25 +467,27 @@ fn main() {
...
-    log_spec_builder.module("h2", log::LevelFilter::Warn);
-
-    Logger::with(log_spec_builder.build())
-        .format(log_format)
-        .log_target(flexi_logger::LogTarget::StdOut)
-        .start()
-        .expect("Failed to create logger");
+    let config = config.logger(Logger::builder().build("h2", log::LevelFilter::Warn));
+    let log_config = config.build(Root::builder().appender("stdout").build(clap_log_level));
+    if let Ok(lc) = log_config {
```
peterschwarz, line 488: What should happen in the error case?

Caleb-Hill: If this commit actually could have an error case we would probably want it to either panic or print a message and continue.  As written it shouldn't ever throw any errors because its not a dynamic configuration. There's basically one compile time option and 4 well established and tested runtime options. I can add a panic/expect if you'd prefer that instead though.

peterschwarz: Perhaps an else block with an `unreachable!()` call and a comment explaining why

Caleb-Hill: That seems reasonable, Ill add push changes in a minute

peterschwarz (approved):
Tested - it doesn't do color output, but I'm not sure that this is a blocker

splinterd/src/main.rs
```diff
@@ -485,25 +467,31 @@ fn main() {
 
     let matches = app.get_matches();
 
-    let log_level = match matches.occurrences_of("verbose") {
+    let clap_log_level = match matches.occurrences_of("verbose") {
```
vaporos, line 470: This change seems unnecessary

splinterd/Cargo.toml
```diff
@@ -38,9 +38,9 @@ clap = "2.32"
 crossbeam-channel = "0.3"
 ctrlc = "3.0"
 cylinder = "0.2.2"
-flexi_logger = "0.14"
 health = { path = "../services/health", optional = true }
 log = "0.4"
+log4rs = "1.0"
```
vaporos, line 43: Use "1" here as the minor revision has no meaning. This differentiates when it is added with purpose to specify a minimum, as in the cylinder case.

Caleb-Hill: fixed

splinterd/src/main.rs
Caleb-Hill: fixed, left from when this was a larger more comprehensive change

peterschwarz (approved):

PR #1373, opened 2021-06-02
Update RefMap to have a generic key

agunde406:
The RefMap is used to keep track of peer's to know when
the connection should be dropped. With the addition of
challenge authorization the peer id is going to be updated
to be an Enum instead of a String, as such the reference map
needs to be updated to take keys that are other types.

### DIFF
--- a/libsplinter/src/collections/ref_map.rs
+++ b/libsplinter/src/collections/ref_map.rs
@@ -18,18 +18,20 @@
 //! internal reference count associated with the given `ref_id`. When `remove_ref` is called, the
 //! reference count is decremented. If a reference count reaches zero, then the item is removed.
 
+use std::borrow::Borrow;
 use std::collections::HashMap;
+use std::hash::Hash;
 
 use super::error::RefMapRemoveError;
 
 /// A map that will keep track of the number of times an ID has been added, and only remove the ID
 /// once the reference count is 0.
-pub struct RefMap {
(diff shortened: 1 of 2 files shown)


vaporos (dismissed):

Caleb-Hill (approved):

libsplinter/src/collections/ref_map.rs
```diff
@@ -54,21 +55,20 @@ impl RefMap {
...
-                )))
+                return Err(RefMapRemoveError(
+                    "Trying to remove a reference that does not exist".into(),
+                ))
             }
         };
 
         if ref_count == 1 {
-            Ok(Some(ref_id.into()))
+            Ok(Some(ref_id.clone()))
```
peterschwarz, line 69: These `clone` and `to_string` calls might be avoidable if you follow a similar pattern to how structures like the standard library's HashMap.  See [HashMap::get](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get) as an example.

agunde406: Updated. Clone is now no longer required and a str can be passed to remove_ref now. Because of how the remove_ref returns the K (not the value) and updates the value if not 0, I needed to switch to using `remove_entry` instead of `remove` to have access to the key.

peterschwarz (approved):

vaporos (approved):

PR #1379, opened 2021-06-04
Add authorization type and node public key to routing table

agunde406:
This information will be used when Challenge Authorization
is implemented to know whether to route over the node id or
a public key.

### DIFF
--- a/libsplinter/src/circuit/routing/mod.rs
+++ b/libsplinter/src/circuit/routing/mod.rs
@@ -186,6 +186,7 @@ pub struct Circuit {
     circuit_id: String,
     roster: Vec<Service>,
     members: Vec<String>,
+    authorization_type: AuthorizationType,
 }
 
 impl Circuit {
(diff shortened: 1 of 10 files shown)


libsplinter/src/circuit/routing/mod.rs
```diff
@@ -220,11 +220,21 @@ impl Circuit {
...
+        public_key: Vec<u8>,
+    },
+}
+
 /// The routing table representation of a node
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CircuitNode {
     node_id: String,
     endpoints: Vec<String>,
+    authorization_type: AuthorizationType,
```
vaporos, line 237: Conceptually this is part of the identifier and it might be clearer in the code if we crated an identifier struct/trait such that all comparisons/lookups happen against the identifier (having multiple parts).

libsplinter/src/circuit/routing/mod.rs
```diff
@@ -234,8 +257,19 @@ impl CircuitNode {
     ///
     /// * `node_id` -  The unique ID for the circuit
     /// * `endpoints` -  A list of endpoints the node can be reached at
-    pub fn new(node_id: String, endpoints: Vec<String>) -> Self {
-        CircuitNode { node_id, endpoints }
+    pub fn new(
+        node_id: String,
+        endpoints: Vec<String>,
+        #[cfg(feature = "challenge-authorization")] public_key: Option<Vec<u8>>,
```
peterschwarz, line 263: Do we want public keys to be represented as Cylinder `PublicKey`s?

agunde406: The plan is to add a Splinter specific type to represent a Public key in these structs before challenge-authorization is stabilized. This struct will only have to/from bytes methods, any other operations would require converting to a Cylinder Public key.

peterschwarz (approved):

vaporos (approved):

PR #1392, opened 2021-06-16
Remove linebreak from splinter upgrade output

leebradley:

### DIFF
--- a/splinterd/src/daemon.rs
+++ b/splinterd/src/daemon.rs
@@ -194,7 +194,7 @@ impl SplinterDaemon {
                 }
 
                 return Err(StartError::StorageError(format!(
-                    "Run the `splinter upgrade` command to update outdated state files to
+                    "Run the `splinter upgrade` command to update outdated state files to \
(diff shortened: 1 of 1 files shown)


splinterd/src/daemon.rs
```diff
@@ -194,7 +194,7 @@ impl SplinterDaemon {
                 }
 
                 return Err(StartError::StorageError(format!(
-                    "Run the `splinter upgrade` command to update outdated state files to
+                    "Run the `splinter upgrade` command to update outdated state files to\
```
agunde406, line 197: this needs a space after to: to\ => to \

Otherwise you get this output `files toa Splinter`

agunde406 (changes requested):

splinterd/src/daemon.rs
leebradley: Fixed!

agunde406 (approved):

vaporos (approved):

PR #1397, opened 2021-06-17
Update example config

rbuysse:


splinterd/packaging/splinterd.toml.example
```diff
@@ -13,44 +13,133 @@
 # limitations under the License.
 
 # Config file format version
-version = "1"
+#version = "1"
```
vaporos, line 16: This line (only this line) should be left uncommented since its metadata about the format of the file.

splinterd/packaging/splinterd.toml.example
```diff
@@ -13,44 +13,133 @@
...
+# Specifies the storage directory for Splinter state. This option overrides the
+# SPLINTER_STATE_DIR environment variable, if set.
+#state_dir = "/var/lib/splinter"
+
+# Specifies the database used for Splinter state.
+#database = "splinter_state.db"
 
 # Friendly identifier for this node. Must be unique on the network
-node_id = "acme-node-000"
+#node_id = ""
```
vaporos, line 29: Should we add a note that this must match the value in the file in /var/lib/splinter?

vaporos (changes requested):

vaporos (approved):

peterschwarz (approved):

PR #1418, opened 2021-06-28
Stabilize upgrade

leebradley:


cli/Cargo.toml
```diff
@@ -63,6 +63,7 @@ default = [
 
 stable = [
     "default",
+    "upgrade",
```
agunde406, line 66: This features should probably be removed all together or moved into default.

vaporos, line 66: This should go into default.

agunde406 (changes requested):

vaporos (approved):

agunde406 (approved):

peterschwarz (approved):

shannynalayna (approved):

PR #1428, opened 2021-07-07
Update the authorization state machine to match the new design

agunde406:
Update the state machine to match the design described in the
challenge authorization design doc:

https://www.splinter.dev/community/planning/challenge_authorization.html

Note this commit is only for trust v1 and the related changes to keep
trust v0 working.

Challenge authorization will be added in a future commit.

### DIFF
--- a/libsplinter/src/network/auth/state_machine/trust_v1.rs
+++ b/libsplinter/src/network/auth/state_machine/trust_v1.rs
@@ -21,8 +21,8 @@ use super::{AuthorizationAction, AuthorizationActionError, AuthorizationState, I
 #[derive(PartialEq, Debug, Clone)]
 pub(crate) enum TrustAuthorizationState {
     TrustConnecting,
-    Identified(String),
-    Authorized(String),
+    Identified(Identity),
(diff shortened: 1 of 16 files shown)


libsplinter/src/network/auth/state_machine/trust_v1.rs
```diff
@@ -38,7 +38,7 @@ impl fmt::Display for TrustAuthorizationState {
 /// The state transitions that can be applied on a connection during authorization.
 #[derive(PartialEq, Debug)]
 pub(crate) enum TrustAuthorizationAction {
-    TrustIdentifying(Identity),
+    TrustIdentifying(String),
```
peterschwarz: Why doesn't this stay as an Identity?

agunde406: Fixed, that was an accident.

Caleb-Hill (approved):

peterschwarz (approved):

vaporos (approved):

PR #1438, opened 2021-07-16
Fix identity comparison in Peer Manger to decide between inbound/outboud connections

agunde406:
If both peers are trying to reconnect at the same time it is possible for two connections to be created for the peer. If this has happens, one connection must be closed but both sides must close the same connection. Previously this was done by comparing the identity of this connection with the local node ID. This however is no longer sufficient with the addition of challenge authorization. This comparison must now be between the connection's identity and the required local authorization used when requesting the connection originally. 

To make this possibly, unidentified peers that are added by endpoint only must now include the required local authorization type so the correct identity is used  in case of future comparison. 

The local authorization used default to challenge auth if it is configured. The key will either be set to splinterd.priv or the only key configured if there is only 1. If there is more than 1 key and splinterd is not one of them a new option has been added to splinterd CLI `--peering-key`. This option takes a name of key that is in the splinterd config directory and will use that for the peering key. 

The `--peer` option also has been updated to take the format `tcp+trust://ipaddr:port/` if the authorization used must be trust. Otherwise default to challenge if supported.

### DIFF
--- a/libsplinter/src/admin/service/shared.rs
+++ b/libsplinter/src/admin/service/shared.rs
@@ -1932,12 +1932,16 @@ impl AdminServiceShared {
             return Ok(());
         }
 
-        let peer_node_pair = self.token_to_peer.get(peer_id).ok_or_else(|| {
-            AdminSharedError::ServiceProtocolError(format!(
-                "Missing service information for peer token : {}",
-                peer_id
-            ))
-        })?;
+        let peer_node_pair = match self.token_to_peer.get(peer_id) {
+            Some(peer_node_pair) => peer_node_pair,
+            None => {
+                warn!(
+                    "Ignoring connection; missing service information for peer token: {}",
+                    peer_id
+                );
+                return Ok(());
+            }
+        };
 
         // We have already received a service protocol request, don't sent another request
         if self
--- a/libsplinter/src/peer/mod.rs
+++ b/libsplinter/src/peer/mod.rs
@@ -92,6 +92,8 @@ pub(crate) enum PeerManagerRequest {
     },
     AddUnidentified {
         endpoint: String,
+        #[cfg(feature = "challenge-authorization")]
+        local_authorization: PeerAuthorizationToken,
         sender: Sender<Result<EndpointPeerRef, PeerUnknownAddError>>,
     },
     RemovePeer {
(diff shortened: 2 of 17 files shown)


splinterd/man/splinterd.1.md
```diff
@@ -185,12 +185,19 @@ OPTIONS
 `--peers PEER-URL` `[,...]`
 : Specifies one or more Splinter nodes that `splinterd` will automatically
   connect to when it starts. The *PEER-URL* argument must specify another node's
-  network endpoint, using the format `protocol_prefix://ip:port`.
+  network endpoint, using the format `protocol_prefix://ip:port` or
```
peterschwarz, line 188: This should probably document the default authorization being challenge.

agunde406: Added

peterschwarz (approved):

peterschwarz (dismissed):

peterschwarz (dismissed):

libsplinter/src/peer/mod.rs
```diff
@@ -213,7 +213,8 @@ impl PeerManager {
         retry_interval: u64,
         max_retry_attempts: u64,
         strict_ref_counts: bool,
-        identity: String,
+        // identity is not used if challenge-authorization is enabled
```
vaporos, line 216: Should this value be logged in the debug!() statement below?

splinterd/src/main.rs
```diff
@@ -267,6 +333,16 @@ fn main() {
                 .multiple(true)
                 .alias("peer"),
         )
+        .arg(
+            Arg::with_name("peering_key")
+                .long("peering-key")
+                .long_help(
+                    "The name of the key to use for challenge authorization with specified peers. \
```
vaporos, line 343: Argument help should ideally fit on a single line.

libsplinter/src/admin/service/shared.rs
```diff
@@ -1932,12 +1932,16 @@ impl AdminServiceShared {
...
-            AdminSharedError::ServiceProtocolError(format!(
-                "Missing service information for peer token : {}",
-                peer_id
-            ))
-        })?;
+        let peer_node_pair = match self.token_to_peer.get(peer_id) {
+            Some(peer_node_pair) => peer_node_pair,
+            None => {
+                warn!(
+                    "Ignoring connection, Missing service information for peer token {}",
```
vaporos, line 1939: Maybe "Ignoring connection; missing service information for peer token: {}"

libsplinter/src/peer/mod.rs
agunde406: I added a commit to add it to the log

splinterd/src/main.rs
agunde406: Shortened the help line

libsplinter/src/admin/service/shared.rs
agunde406: replaced with suggested

peterschwarz (dismissed):

vaporos:
LGTM but needs conflict resolution

peterschwarz (approved):

vaporos (approved):

PR #1439, opened 2021-07-16
Fix scabbard batch result send/receive bug

isabeltomb:
Update batch result receiving process to loop until getting a `None` result from the scheduler before shutting down. This change ensures the batch result is received.

Previously, when submitting transactions to scabbard, the error: "Unable to send batch result; receiver must have dropped" would occur randomly and somewhat frequently. This change fixes the issue by waiting for the scheduler to indicate it is finished by sending `None` before shutting down.

### Testing:
1. start two splinter nodes that build locally from this branch
2. use the scabbard cli to upload the smallbank smart contract
3. use the transact cli to start a workload
4. watch the logs to ensure there are no errors being reported by scabbard that say 
`Unable to send batch result; receiver must have dropped`

### DIFF
--- a/services/scabbard/libscabbard/src/service/state.rs
+++ b/services/scabbard/libscabbard/src/service/state.rs
@@ -252,12 +252,25 @@ impl ScabbardState {
         self.executor
             .execute(scheduler.take_task_iterator()?, scheduler.new_notifier()?)?;
 
+        let mut recv_result: Option<BatchExecutionResult> = None;
+
         // Get the results and shutdown the scheduler
-        let recv_result = result_rx.recv_timeout(Duration::from_secs(EXECUTION_TIMEOUT));
+        // after receiving the batch result wait until the receiver gets a `None` response
+        // from the scheduler before shutting down
+        loop {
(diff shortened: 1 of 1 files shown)


peterschwarz (dismissed):

services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -252,12 +252,25 @@ impl ScabbardState {
         self.executor
             .execute(scheduler.take_task_iterator()?, scheduler.new_notifier()?)?;
 
+        let mut recv_result: Option<BatchExecutionResult> = None;
+
         // Get the results and shutdown the scheduler
-        let recv_result = result_rx.recv_timeout(Duration::from_secs(EXECUTION_TIMEOUT));
+        // after recieving the batch result wait until the receiver gets a `None` response
```
agunde406, line 258: recieving -> receiving

agunde406 (changes requested):

services/scabbard/libscabbard/src/service/state.rs
isabeltomb: Fixed

agunde406 (approved):

peterschwarz (approved):

PR #1441, opened 2021-07-20
Add splinterd db version detection and requirement

leebradley:
Add a startup check in splinterd that checks if the database that the
daemon uses is out of date. If so, splinterd aborts and informs the user
that they must run the `splinter database migrate` cli command.

### DIFF
--- a/libsplinter/src/store/mod.rs
+++ b/libsplinter/src/store/mod.rs
@@ -31,6 +31,10 @@ pub use self::sqlite::ForeignKeyCustomizer;
 use diesel::r2d2::{ConnectionManager, Pool};
 
 use crate::error::{InternalError, InvalidArgumentError};
+#[cfg(feature = "postgres")]
+use crate::migrations::any_pending_postgres_migrations;
+#[cfg(feature = "sqlite")]
+use crate::migrations::any_pending_sqlite_migrations;
 
 /// An abstract factory for creating Splinter stores backed by the same storage
 pub trait StoreFactory {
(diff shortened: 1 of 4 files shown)


peterschwarz (approved):

libsplinter/src/store/mod.rs
```diff
@@ -91,6 +95,12 @@ pub fn create_store_factory(
                     "Failed to build connection pool".to_string(),
                 )
             })?;
+            let conn = pool
+                .get()
+                .map_err(|err| InternalError::from_source(Box::new(err)))?;
+            if !any_pending_postgres_migrations(&conn)? {
+                return Err(InternalError::with_message(String::from("The splinter database is out of date. You must run `splinter migrate` before running splinterd")));
```
agunde406, line 102: Please wrap these error messages at 100 characters

agunde406 (changes requested):
If you try to start splinterd with an empty database(the file exists but migration have not be run at all) you get the following error  

```
[2021-07-20 15:32:09.827] T[main] ERROR [splinterd] Failed to start daemon, unable to start the Splinter daemon: unable to set up storage: Failed to initialize store factory: no such table: __diesel_schema_migrations
```

Would it be possible to handle this error and provide a simular error telling the user to run migrations

libsplinter/src/store/mod.rs
agunde406: the command is also `splinter database migrate`

leebradley: Cool! Made requested changes:
* Simplified `any_pending_postgres_migrations` logic to translate an error fetching the current version as a blank slate (I hadn't expected the API would behave that way, so it's good you caught it)
* Wrapped error messages to 100 characters
* Corrected cli command to say `splinter database migrate`

libsplinter/src/store/mod.rs
```diff
@@ -91,6 +95,15 @@ pub fn create_store_factory(
                     "Failed to build connection pool".to_string(),
                 )
             })?;
+            let conn = pool
+                .get()
+                .map_err(|err| InternalError::from_source(Box::new(err)))?;
+            if !any_pending_postgres_migrations(&conn)? {
+                return Err(InternalError::with_message(String::from(
+                    "The splinter database is out of date. You must \
```
peterschwarz, line 103: If this test does fail, it implies that the database has not been initialized.  Same fix, but slightly different context.

leebradley: Updated message to say:
> This version of splinter requires migrations that are not yet applied to the database. Run `splinter database migrate` to apply migrations before running splinterd

peterschwarz (approved):

agunde406 (approved):

PR #1443, opened 2021-07-20
Add Splinter specific PublicKey struct

agunde406:
This struct will be used for internal splinter structs
so we don't need to use the Cylinder's public key as
a part of the splinter API but also is more specific then
a Vec.

### DIFF
--- a/libsplinter/src/network/auth/handlers/builder.rs
+++ b/libsplinter/src/network/auth/handlers/builder.rs
@@ -223,7 +223,9 @@ impl AuthorizationDispatchBuilder {
             let signers_to_use = match &self.local_authorization {
                 Some(ConnectionAuthorizationType::Challenge { public_key }) => {
                     let signer = signers.iter().find(|signer| match signer.public_key() {
-                        Ok(signer_public_key) => signer_public_key.as_slice() == public_key,
+                        Ok(signer_public_key) => {
+                            signer_public_key.as_slice() == public_key.as_slice()
+                        }
                         Err(_) => false,
                     });
 
--- a/libsplinter/src/peer/token.rs
+++ b/libsplinter/src/peer/token.rs
@@ -60,15 +60,6 @@ impl PeerAuthorizationToken {
         }
     }
 
-    #[cfg(feature = "challenge-authorization")]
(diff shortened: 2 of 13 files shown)


Caleb-Hill (approved):

libsplinter/src/peer/token.rs
```diff
@@ -65,7 +67,9 @@ impl PeerAuthorizationToken {
     pub fn has_public_key(&self, public_keys: &[Vec<u8>]) -> bool {
         match self {
             PeerAuthorizationToken::Trust { .. } => false,
-            PeerAuthorizationToken::Challenge { public_key } => public_keys.contains(&public_key),
+            PeerAuthorizationToken::Challenge { public_key } => {
+                public_keys.contains(&public_key.as_slice().to_vec())
```
peterschwarz, line 71: Why does this need to be converted to a vec?  There might be a way to do this without an allocation.

libsplinter/src/network/auth/handlers/builder.rs
```diff
@@ -223,7 +223,9 @@ impl AuthorizationDispatchBuilder {
             let signers_to_use = match &self.local_authorization {
                 Some(ConnectionAuthorizationType::Challenge { public_key }) => {
                     let signer = signers.iter().find(|signer| match signer.public_key() {
-                        Ok(signer_public_key) => signer_public_key.as_slice() == public_key,
+                        Ok(signer_public_key) => {
+                            signer_public_key.as_slice() == public_key.as_slice()
```
peterschwarz, line 227: Since `PublicKey` implements `PartialEq`, can we just do the comparison on the public keys?

agunde406: This is comparing splinter::public_key::PublicKey against a cylinder::PublicKey

libsplinter/src/peer/token.rs
agunde406: This method is actually not used anywhere so I added a commit to remove it

peterschwarz (approved):

isabeltomb (approved):

PR #1446, opened 2021-07-22
Node id update command

Caleb-Hill:
Adds node_id import functionality to the cli update command.

### DIFF
--- a/cli/src/action/database/upgrade/mod.rs
+++ b/cli/src/action/database/upgrade/mod.rs
@@ -39,13 +39,6 @@ impl Action for UpgradeAction {
     fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
         let state_dir = get_state_dir(arg_matches)?;
         let database_uri = get_database_uri(arg_matches)?;
-        info!("Upgrading splinterd state");
-        info!(
-            "Source yaml state directory: {}",
-            state_dir.to_string_lossy()
-        );
-        info!("Destination database uri: {}", database_uri);
-        info!("Loading YAML datastore... ");
         let store_factory = splinter::store::create_store_factory(database_uri).map_err(|err| {
             CliError::ActionError(format!("failed to initialized store factory: {}", err))
         })?;
@@ -55,6 +48,14 @@ impl Action for UpgradeAction {
             let db_store = store_factory.get_node_id_store();
             node_id::migrate_node_id_to_db(state_dir.clone(), db_store)?;
         }
+        info!("Upgrading splinterd state");
+        info!(
+            "Source yaml state directory: {}",
+            state_dir.to_string_lossy()
+        );
+        let database_uri = get_database_uri(arg_matches)?;
+        info!("Destination database uri: {}", database_uri);
+        info!("Loading YAML datastore... ");
         let db_store = store_factory.get_admin_service_store();
         yaml::import_yaml_state_to_database(state_dir, &*db_store)?;
         Ok(())
--- a/cli/src/action/database/upgrade/mod.rs
+++ b/cli/src/action/database/upgrade/mod.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Provides database upgrade functionality
+
 mod yaml;
 
 use std::path::PathBuf;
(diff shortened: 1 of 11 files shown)


libsplinter/src/node_id/store/mod.rs
```diff
@@ -16,6 +16,7 @@
 pub mod diesel;
 /// NodeIdStore error types
```
peterschwarz, line 17: Not part of this PR, but this should be module doc in the error module, not a comment here.

peterschwarz (changes requested):

libsplinter/src/node_id/store/file/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+//! NodeIdStore that supports the 0.4 node_id file format.
+
+use std::fs;
+use std::path::PathBuf;
+
+use super::NodeIdStore;
+use super::NodeIdStoreError;
+
+/// The 0.4 node_id file is depricated, this exists to help migrate the node_id.
```
peterschwarz, line 23: depricate -> deprecated

But more specifically, change it to "soft-deprecated", as Rust used to describe `Error::description()` until it was deprecated via the metadata - since we're not adding that metadata here yet.

I would also change this sentence to be part of the doc detail and the summary line should just be "A NodeIdStore backed by a file".

cli/src/action/database/upgrade/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+use super::Action;
+
+/// The overarching Action possibly containing multiple upgrade actions
+pub struct UpgradeAction;
+
+impl Action for UpgradeAction {
+    fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
+        let state_dir = get_state_dir(arg_matches)?;
+        let database_uri = get_database_uri(arg_matches)?;
+        let connection_uri = database_uri.parse().map_err(|_| {
```
peterschwarz, line 41: Why not return a `ConnectionUri` from the `get_database_uri` function - it's already parsed as such to check connectivity.

cli/src/action/database/upgrade/mod.rs
```diff
@@ -44,6 +46,11 @@ impl Action for UpgradeAction {
                 CliError::ActionError(format!("failed to initialized store factory: {}", err))
             })?;
 
+        #[cfg(feature = "node-id-upgrade")]
```
agunde406, line 49: Can you also update the man page for the `splinter upgrade` command

cli/src/action/database/upgrade/mod.rs
Caleb-Hill: Updated to return a ConnectionUri

libsplinter/src/node_id/store/file/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+//! A NodeIdStore backed by a file.
+
+use std::fs;
+use std::path::PathBuf;
+
+use super::NodeIdStore;
+use super::NodeIdStoreError;
+
+/// The 0.4 node_id file is soft-deprecated, this exists to help migrate the node_id.
```
peterschwarz, line 23: The summary line should be the same as the module doc. This line is more of a detail description.

Caleb-Hill: Updated the line to match the module doc

peterschwarz (approved):

cli/src/action/database/upgrade/node_id.rs
```diff
@@ -0,0 +1,89 @@
...
+    match (from.get_node_id(), to.get_node_id()) {
+        (Ok(Some(id)), Ok(None)) => to
+            .set_node_id(id)
+            .map_err(|e| CliError::ActionError(format!("{}", e)))
+            .map(|_| WarningEmited::No),
+        (Ok(Some(_)), Ok(Some(_))) => Err(CliError::ActionError(
+            "destination store already has node_id set".to_string(),
+        )),
+        (Ok(None), _) => {
+            warn!("node_id file is empty, skipping import");
```
agunde406, line 44: Make this capital (and the the other warnings) to match the other logging that happens during the subcommand

agunde406: I think we should have all of the skipping logs start like "Skipping ...."

cli/src/action/database/upgrade/yaml.rs
```diff
@@ -101,9 +101,8 @@ pub fn import_yaml_state_to_database<P: Into<PathBuf>>(
     let proposals_location = state_dir.join(PROPOSALS_FILE);
 
     if !(circuits_location.exists() || proposals_location.exists()) {
-        return Err(CliError::ActionError(
-            "no yaml state files found to import".to_string(),
-        ));
+        warn!("no yaml state files found to import");
```
agunde406, line 104: Maybe changed this to  "Skipping YAML circuit state import, no yaml state files found"

cli/src/action/database/upgrade/yaml.rs
```diff
@@ -336,6 +336,7 @@ proposals:
     }
 
     #[test]
+    #[ignore]
```
agunde406, line 339: I think this test should just be updated to verify that it return Ok(())

agunde406 (dismissed):

peterschwarz (approved):

shannynalayna (approved):

PR #1449, opened 2021-07-29
Node id block startup if file exists

Caleb-Hill:
As node_id moves into the database eventually we want splinterd not to start with the old node_id storage file in place. This makes splinterd check for the file and exit with an error message if it exists.

### DIFF
--- a/splinterd/src/main.rs
+++ b/splinterd/src/main.rs
@@ -195,8 +195,9 @@ fn load_signer_keys(
             .path();
 
         if path.extension() == Some(OsStr::new("priv")) {
-            let private_key = load_key_from_path(&path)
-                .map_err(|err| UserError::KeyError(InternalError::from_source(Box::new(err))))?;
+            let private_key = load_key_from_path(&path).map_err(|err| {
+                UserError::InternalError(InternalError::from_source(Box::new(err)))
+            })?;
             let signing_key = Secp256k1Context::new().new_signer(private_key);
 
             if path.file_stem() == Some(OsStr::new(peering_key)) {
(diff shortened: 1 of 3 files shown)


peterschwarz (changes requested):
Tested.  Works as expected when upgrading from the yaml file to a DB stored node-id. However, it does not have the existing behaviour of starting up and generating a node id if one is missing and not provided on the command line.

peterschwarz (approved):
Tested

peterschwarz (approved):

splinterd/src/main.rs
```diff
@@ -97,65 +102,116 @@ fn create_config(_toml_path: Option<&str>, _matches: ArgMatches) -> Result<Confi
...
+                }
+                None => {
+                    // If the config object does not have a node_id, continue with the node_id read
+                    // from the file.
+                    Ok(file_node_id)
+                }
+            }
+        } else {
+            // If node file does not exist, need to create and save a node_id file.
+            // Check if the config obejct has a node_id, otherwise generate a random one.
```
isabeltomb, line 187: s/obejct/object

splinterd/src/main.rs
```diff
@@ -97,65 +102,116 @@ fn create_config(_toml_path: Option<&str>, _matches: ArgMatches) -> Result<Confi
...
+                .database()
+                .parse()
+                .map_err(|_| UserError::InvalidArgument("db_connection".to_string()))?;
+            let store = create_store_factory(database_uri)?.get_node_id_store();
+            let db_node_id = store.get_node_id();
+            let config_node_id = config.node_id();
+            let save_new_node_id = |node_id| -> Result<(), UserError> {
+                store
+                    .set_node_id(node_id)
+                    .map_err(|err| UserError::from(InternalError::from_source(Box::new(err))))
```
isabeltomb, line 124: I think "node-file-block" needs to be added to the feature guard for `use splinter::error::InternalError;` on line 44, attempting to compile with just the `node-file-block` feature returns the error: "failed to resolve: use of undeclared type `InternalError`"

isabeltomb (commented):
Tested

isabeltomb (approved):

peterschwarz (approved):

PR #1459, opened 2021-08-03
Add a --skip option to splinter keygen

agunde406:
If the key already exists, the command will log that it is
skipping and exit without error.

This also updates the create_key_pair function to not return
the public key as bytes as this was never used.

### DIFF
--- a/cli/src/action/keygen.rs
+++ b/cli/src/action/keygen.rs
@@ -101,6 +101,7 @@ impl Action for KeyGenAction {
             private_key_path,
             public_key_path,
             args.is_present("force"),
+            args.is_present("skip"),
             true,
         )?;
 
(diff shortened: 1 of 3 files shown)


peterschwarz (dismissed):

cli/src/action/keygen.rs
```diff
@@ -116,9 +117,18 @@ pub fn create_key_pair(
...
+            info!(
+                "Skipping, key already exists: {}",
+                private_key_path.display()
+            );
+            return Ok(());
+        }
+
         if private_key_path.exists() {
             return Err(CliError::EnvironmentError(format!(
                 "File already exists: {:?}",
```
leebradley, line 134: So if a user specifies `--skip` (for which the documentation says `Skip generating the files if they exist.`) the keygen command would still abort with the error `File already exists: ...` if only one of the keypair files exists.

Aborting makes sense because the files are a pair, and `--skip` should only skip if both exist. I just think the message in combination with `--skip` becomes contextually confusing. Maybe there should be a custom error that explains this when `skip_create` is true, but the files exist?

leebradley (changes requested):

cli/src/action/keygen.rs
agunde406: Updated to check each case and return errors in the case of one file missing.

peterschwarz (approved):

leebradley (approved):

leebradley (approved):

Caleb-Hill (approved):

shannynalayna (approved):

PR #1467, opened 2021-08-12
Trim line breaks off node_id's from file

Caleb-Hill:
The node_id file is supposed to support having a line break/white space after
the node_id string. This change just trims the node_id string of any
tailing white space.


peterschwarz (approved):

leebradley (approved):
Tested locally. Appears to fix the issue!

PR #1482, opened 2021-08-23
Add specificity to pending batches metric

leebradley:
This change fixes an issue where pending batches from multiple scabbard
circuits would override one another, by adding a tag that clears up
which specific circuit the pending batches list originates from.

### DIFF
--- a/libsplinter/src/metrics/mod.rs
+++ b/libsplinter/src/metrics/mod.rs
@@ -19,6 +19,7 @@ pub mod influx;
 #[macro_export]
 macro_rules! counter {
     ($t:tt, $v:expr) => {};
+    ($t:tt, $v:expr, $($key:expr => $value:expr)*) => {};
 }
 
 // no-op `gauge` macro for when the `metrics` feature is not enabled
@@ -26,6 +27,7 @@ macro_rules! counter {
 #[macro_export]
 macro_rules! gauge {
     ($t:tt, $v:expr) => {};
+    ($t:tt, $v:expr, $($key:expr => $value:expr)*) => {};
 }
 
 // no-op `histogram` macro for when the `metrics` feature is not enabled
@@ -33,4 +35,5 @@ macro_rules! gauge {
 #[macro_export]
 macro_rules! histogram {
     ($t:tt, $v:expr) => {};
+    ($t:tt, $v:expr, $($key:expr => $value:expr)*) => {};
 }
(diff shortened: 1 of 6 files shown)


services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -87,14 +91,18 @@ impl ScabbardShared {
         .expect("String -> PeerId -> String conversion should not fail");
 
         // initialize pending_batches metric
-        gauge!("splinter.scabbard.pending_batches", 0);
+        gauge!("splinter.scabbard.pending_batches", 0,
+                "circuit_id" => circuit_id.clone()
```
peterschwarz, line 95: This should actually be `"service" => format!("{}::{}", circuit_id, service_id)`

libsplinter/src/metrics/mod.rs
```diff
@@ -19,18 +19,21 @@ pub mod influx;
 #[macro_export]
 macro_rules! counter {
     ($t:tt, $v:expr) => {};
+    ($t:tt, $v:expr, $key:expr => $value:expr) => {};
```
peterschwarz, line 22: Shouldn't this be repeated?

leebradley: Fixed!

peterschwarz (dismissed):

agunde406 (changes requested):
Tested, one small nitpick

services/scabbard/libscabbard/src/service/shared.rs
```diff
@@ -125,12 +133,17 @@ impl ScabbardShared {
         self.accepting_batches
     }
 
-    pub fn add_batch_to_queue(&mut self, batch: BatchPair) -> Result<(), ScabbardError> {
-        self.batch_queue.push_back(batch);
+    fn update_pending_batches(&self, _batches: i64) {
```
agunde406, line 136: Please add a comment about why this requires batches to be prefixes with underscore.

leebradley: Updated with an explanation:
```rust
    /// Updates pending batches metrics gauge
    ///
    /// # Arguments
    ///
    /// * `_batches` - The number of pending batches for this service. It is prefixed with an
    /// underscore due to rust recognizing the metrics macro noop when the metrics feature is
    /// disabled
    fn update_pending_batches(&self, _batches: i64) {
        gauge!(
            "splinter.scabbard.pending_batches",
            _batches,
            "service" => format!("{}::{}", self.circuit_id, self.service_id)
        );
    } 
```

agunde406 (approved):

peterschwarz (approved):

PR #1488, opened 2021-08-25
Allow endpoints to belong to multiple tokens that are PeerAuthorizationToken::Challenge 

agunde406:
Before challenge authorization was implemented, their was a requirement that endpoints to peer id had a 1:1 relationship (same with connections to endpoints), however with challenge authorization, the different peer ids must have its own connection. This means that as long as two peer id's are not different trust ids, the peers can share the same endpoint(s). 

This PR relaxes the requirement for unique endpoints in the ConnectionManager, PeerMananger, and PeerMap. 

This PR also updates the protocol agreement in the admin service to be based on the PeerAuthorizationToken instead of on the admin service id, this ensures that all connections are correctly waited for and verified. 

An integration tests is added that created two circuit, with different keys for both nodes in each circuit. The original test that has two circuits where one node's keys stays the same is still ignored and requires more bug fixes that will be coming soon.

### DIFF
--- a/libsplinter/src/network/connection_manager/builder.rs
+++ b/libsplinter/src/network/connection_manager/builder.rs
@@ -240,9 +240,13 @@ fn handle_request<T: ConnectionMatrixLifeCycle, U: ConnectionMatrixSender>(
             authorizer,
             subscribers,
         ),
-        CmRequest::RemoveConnection { endpoint, sender } => {
+        CmRequest::RemoveConnection {
+            endpoint,
+            connection_id,
+            sender,
+        } => {
             let response = state
-                .remove_connection(&endpoint)
+                .remove_connection(&endpoint, &connection_id)
                 .map(|meta_opt| meta_opt.map(|meta| meta.endpoint().to_owned()));
 
             if sender.send(response).is_err() {
@@ -254,7 +258,7 @@ fn handle_request<T: ConnectionMatrixLifeCycle, U: ConnectionMatrixSender>(
                 .send(Ok(state
                     .connection_metadata()
                     .iter()
-                    .map(|(key, _)| key.to_string())
+                    .map(|(_, metadata)| metadata.endpoint().to_string())
                     .collect()))
                 .is_err()
             {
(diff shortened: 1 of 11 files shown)


libsplinter/src/network/connection_manager/builder.rs
```diff
@@ -333,42 +337,42 @@ fn send_heartbeats<T: ConnectionMatrixLifeCycle, U: ConnectionMatrixSender>(
...
                 if reconnecting {
                     if last_connection_attempt.elapsed().as_secs() > retry_frequency {
-                        reconnections.push(endpoint.to_string());
+                        reconnections.push(metadata.clone());
                     }
                 } else {
-                    trace!("Sending heartbeat to {}", endpoint);
-                    if let Err(err) = matrix_sender
-                        .send(metadata.connection_id.clone(), heartbeat_message.clone())
+                    trace!("Sending heartbeat to {}", metadata.endpoint());
```
peterschwarz, line 343: Do we send a heartbeat to all connections, even if there is a duplicate endpoint?  If so, we should include the connection id as part of any connection logging here (e.g. `{endpoint} ({connection_id})`).  

As for my question, I suppose one of the connections could be closed remotely, but not the other, so we would have to detect that.

libsplinter/src/peer/mod.rs
```diff
@@ -1540,6 +1534,37 @@ fn log_connect_request_err(
...
+    if matches!(peer_id, PeerAuthorizationToken::Challenge { .. }) {
+        return false;
+    }
+
+    for endpoint in endpoints {
+        if let Some(peers) = peer_map.get_peer_from_endpoint(endpoint) {
+            for peer_meta in peers {
+                // allow irrefutable_let_patterns because it is required if challenge-authorization
+                // feature is enabled
+                #[allow(irrefutable_let_patterns)]
```
peterschwarz, line 1555: could you just use the `matches!` macro like above?

libsplinter/src/network/connection_manager/mod.rs
```diff
@@ -205,9 +206,9 @@ pub struct Connector {
 }
 
 impl Connector {
-    /// Request a connection to the given endpoint.
+    /// Request a connection to the given endpoint with a provided connection id.
```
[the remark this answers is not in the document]
agunde406: Fixed

libsplinter/src/peer/mod.rs
agunde406: Yes, that does work, updated

vaporos (approved):

peterschwarz (approved):

PR #1494, opened 2021-08-27
Explicitly set gameroom splinterd REST API to http

agunde406:
This allows gameroom to work when experimental features is enabled,
which turns on HTTPS support and changes the default type to
https if http:// is not included in the endpoint.


leebradley (approved):
Works!

rbuysse (approved):

PR #1496, opened 2021-08-31
Test auth support actix 1

Caleb-Hill:
Add authorization support to the test framework

### DIFF
--- a/splinterd/src/node/builder/mod.rs
+++ b/splinterd/src/node/builder/mod.rs
@@ -21,8 +21,11 @@ pub(super) mod scabbard;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use cylinder::{secp256k1::Secp256k1Context, Context, Signer, VerifierFactory};
+use cylinder::{secp256k1::Secp256k1Context, Context, Signer, Verifier, VerifierFactory};
 use rand::{thread_rng, Rng};
+use splinter::biome::credentials::rest_api::{
+    BiomeCredentialsRestResourceProvider, BiomeCredentialsRestResourceProviderBuilder,
+};
 use splinter::error::InternalError;
 use splinter::public_key::PublicKey;
 use splinter::rest_api::actix_web_1::RestApiBuilder as RestApiBuilder1;
@@ -59,6 +62,8 @@ pub struct NodeBuilder {
     node_id: Option<String>,
     enable_biome: bool,
     signers: Option<Vec<Box<dyn Signer>>>,
+    biome_auth: Option<BiomeCredentialsRestResourceProvider>,
+    cylinder_auth: Option<Box<dyn Verifier>>,
 }
 
 impl Default for NodeBuilder {
@@ -79,6 +84,8 @@ impl NodeBuilder {
             node_id: None,
             enable_biome: false,
             signers: None,
+            biome_auth: None,
+            cylinder_auth: None,
         }
     }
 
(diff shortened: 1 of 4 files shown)


splinterd/tests/framework/network.rs
```diff
@@ -87,6 +87,13 @@ impl Network {
             auth_types: vec![AuthType::Biome, AuthType::Cylinder],
         }
     }
+    pub fn new_with_auth(auth_types: Vec<AuthType>) -> Self {
```
peterschwarz, line 90: There could be a method that takes `&mut self` to set these values.

peterschwarz (approved):

splinterd/tests/framework/network.rs
```diff
@@ -16,10 +16,14 @@
...
 use std::collections::HashMap;
 use std::fs::File;
+use std::iter::Iterator;
 
 use cylinder::{secp256k1::Secp256k1Context, Context};
+use splinter::biome::credentials::rest_api::BiomeCredentialsRestResourceProviderBuilder;
+use splinter::biome::{MemoryCredentialsStore, MemoryKeyStore, MemoryRefreshTokenStore};
 use splinter::error::{InternalError, InvalidArgumentError};
 use splinter::registry::Node as RegistryNode;
+use splinter::rest_api::AuthConfig;
```
vaporos, line 26: Avoid importing actix_web_1 directly from the rest_api module; the re-export was added for backward compatibility.

splinterd/src/node/builder/mod.rs
```diff
@@ -185,6 +188,11 @@ impl NodeBuilder {
         self
     }
 
+    pub fn with_auth_configs(mut self, auth_configs: Vec<AuthConfig>) -> Self {
```
vaporos, line 191: This is awkward here for two reasons: 1) it exposes actix web 1 code into the testing APIs, potentially causing a lot of test rewriting when we remove actix web 1 support in 0.7 in a couple months; 2) it is decoupled from the variant type; this is only valid for actix web 1, but the API allows setting actix web 3 variant and an actix web 1 auth config.

splinterd/tests/framework/network.rs
```diff
@@ -89,6 +89,10 @@ impl Network {
         }
     }
 
+    pub fn with_auth(&mut self, auth_types: Vec<AuthType>) {
```
vaporos, line 92: This should probably be called with_auth_types.

vaporos (changes requested):

splinterd/tests/framework/network.rs
```diff
@@ -41,15 +44,49 @@ pub enum NetworkNode {
 
 impl Network {
     pub fn new() -> Network {
+        let biome_auth = {
+            #[cfg(feature = "biome-credentials")]
```
peterschwarz, line 48: I think you can just use `cfg!(feature = "biome-credentials")` for this

splinterd/tests/framework/network.rs
```diff
@@ -91,9 +127,37 @@ impl Network {
                 .with_admin_signer(admin_signer)
                 .with_signers(signers)
                 .with_external_registries(self.external_registries.clone())
-                .with_biome_enabled()
-                .build()?
-                .run()?;
+                .with_biome_enabled();
+            let builder = match (self.biome_auth, self.cylinder_auth) {
```
peterschwarz, line 131: Why the match statement? Couldn't you just do

```rust
if self.biome_auth {
   ...
   builder = builder.with_biome_auth(...);
}

if self.cylinder_auth {
    builder = builder.with_cylinder_auth(...);
}
```

Seems clearer to me, given there's no relationship between the two auth methods.

splinterd/tests/framework/network.rs
Caleb-Hill: No longer import anything from actix_web_1

peterschwarz (dismissed):

splinterd/src/node/runnable/mod.rs
```diff
@@ -41,6 +42,30 @@ pub(super) enum RunnableNodeRestApiVariant {
     ActixWeb3(RunnableRestApi),
 }
 
+impl RunnableNodeRestApiVariant {
+    #[cfg(feature = "biome-credentials")]
+    pub fn with_biome_auth(self, provider: BiomeCredentialsRestResourceProvider) -> Self {
```
isabeltomb, line 47: the `auth_configs` value that these methods set in the `RestApiBuilder` will always get over written lower down in this file, line 117-118 `let (rest_api_shutdown_handle, rest_api_join_handle) = rest_api.with_auth_configs(auth_configs)` so I think you'll probably just need to update that chunk to do `append_auth` as well rather than `with_auth_configs`

isabeltomb (approved):

splinterd/src/node/runnable/mod.rs
```diff
@@ -85,12 +110,12 @@ impl RunnableNode {
                     // Build the `BiomeResourceProvider` to allow the node to access `Biome` endpoints
                     let mut biome_resource_provider =
                         BiomeResourceProvider::new(&*admin_subsystem.store_factory)?;
-                    auth_configs.append(&mut biome_resource_provider.auth_configs);
+                    rest_api =
+                        rest_api.append_auth_configs(&mut biome_resource_provider.auth_configs);
```
vaporos, line 114: This should perhaps move to the builder so that we can remove the mut from 'RunnableNodeRestApiVariant::ActixWeb1(mut rest_api)'

vaporos (changes requested):

vaporos (approved):

splinterd/tests/framework/network.rs
Caleb-Hill: removed

splinterd/tests/framework/network.rs
Caleb-Hill: removed

splinterd/tests/framework/network.rs
Caleb-Hill: fixed

peterschwarz (approved):

isabeltomb (dismissed):

peterschwarz (approved):

isabeltomb (approved):

vaporos (approved):

PR #1497, opened 2021-09-01
Scabbard receipt store integration

isabeltomb:
This PR adds the new receipt store trait implemented in Sawtooth-lib to scabbard. The new receipt store is added to scabbard along side the existing receipt store while it is still experimental. The new receipt store is behind the experimental feature "diesel-receipt-store".

The new receipt store trait is implemented for both PostgreSQL and SQLite backends, scabbard will use the same PostgreSQL or SQLite instance that is being used by Splinter for the receipt store.

To test run splinter database migrate with experimental features turned on you should see either of the following messages:
```
Running migrations against PostgreSQL database for receipt store: postgres://postgres:<pg-password>@localhost:5432/splinter
Successfully applied PostgreSQL migrations
```
or
```
Running migrations against SQLite database for receipt store: /data/splinter_state.db
Successfully applied SQLite migrations
```
then start up a splinter node with experimental features on, look at the transaction receipts database tables start a second splinter node, create a circuit and follow the steps to upload a smart contract and grant necessary permissions. 

The smart contract setup steps should create 4 transaction receipts which can be viewed in the transaction_receipt table of the database being used with the SQL statement `SELECT * FROM 'transaction_receipt';`
Starting a workload will submit additional transactions and add more transaction receipts to the tables.


splinterd/src/daemon.rs
```diff
@@ -453,6 +455,15 @@ impl SplinterDaemon {
             }
         }
 
+        #[cfg(feature = "scabbard-receipt-store")]
+        run_receipt_store_migrations(&self.db_url.to_string()).map_err(|err| {
```
peterschwarz, line 459: This should be run with the `splinter database migrate` command, not here in the daemon

services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -162,6 +164,45 @@ impl ScabbardState {
         // initialize committed_batches metric
         counter!("splinter.scabbard.committed_batches", 0);
 
+        #[cfg(feature = "diesel-receipt-store")]
+        let receipt_store: Arc<RwLock<dyn ReceiptStore>> =
```
peterschwarz, line 168: Should this be an alternative to the existing lmdb files?

services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -666,18 +671,25 @@ enum EventQuery {
 /// An iterator that wraps the `TransactionReceiptStore` and returns `StateChangeEvent`s using an
 /// in-memory cache.
 pub struct Events {
+    // transaction_receipt_store is not used if diesel-receipt-store is enabled
+    #[allow(unused_variables)]
     transaction_receipt_store: Arc<RwLock<TransactionReceiptStore>>,
+    #[cfg(feature = "diesel-receipt-store")]
+    receipt_store: Arc<RwLock<dyn ReceiptStore>>,
```
peterschwarz, line 678: These are different enough that I would actually suggest two different implementations of the struct, one for when the feature is enabled and one for `not`.

peterschwarz (changes requested):

splinterd/src/daemon.rs
isabeltomb: Moved this functionality to be run with `splinter database migrate` if "scabbard-receipt-store" is enabled

services/scabbard/libscabbard/src/service/state.rs
isabeltomb: Modified ScabbardState to only create the lmdb receipt store if "diesel-receipt-store" is not enabled

services/scabbard/libscabbard/src/service/state.rs
isabeltomb: Updated

peterschwarz (approved):

agunde406 (approved):
Tested locally.

PR #1499, opened 2021-09-02
Scabbard Commit Hash Store trait and LMDB Implementation

peterschwarz:
This change adds the `CommitLogStore` trait to libscabbard and implements a "Legacy" LMDB version that uses the `index_get/put` against an existing DB. The legacy implementation can properly detect if they LMDB instance supplied is has the index pre-configured and return an invalid state error, if not.

### DIFF
--- a/services/scabbard/libscabbard/Cargo.toml
+++ b/services/scabbard/libscabbard/Cargo.toml
@@ -63,6 +63,7 @@ stable = [
   "client-reqwest",
   "default",
   "events",
+  "lmdb",
   "rest-api",
   "rest-api-actix",
 ]
@@ -87,6 +88,7 @@ commit-store = []
 diesel-receipt-store = ["diesel"]
 events = ["splinter/events"]
 factory-builder = []
+lmdb = []
 rest-api = ["futures", "splinter/rest-api"]
 rest-api-actix = ["actix-web", "splinter/rest-api-actix"]
 service-arg-validation = ["splinter/service-arg-validation"]
--- a/services/scabbard/libscabbard/Cargo.toml
+++ b/services/scabbard/libscabbard/Cargo.toml
@@ -72,6 +72,7 @@ experimental = [
   "stable",
   # The following features are experimental:
   "back-pressure",
+  "commit-store",
   "diesel-receipt-store",
   "factory-builder",
   "metrics",
@@ -82,6 +83,7 @@ authorization = ["splinter/authorization"]
 back-pressure = []
 client = []
 client-reqwest = ["client", "reqwest"]
+commit-store = []
 diesel-receipt-store = ["diesel"]
 events = ["splinter/events"]
 factory-builder = []
--- a/services/scabbard/libscabbard/src/store/error.rs
+++ b/services/scabbard/libscabbard/src/store/error.rs
@@ -1,0 +1,76 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


services/scabbard/libscabbard/src/store/lmdb/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    fn set_current_commit_hash(
+        &self,
+        _service: &ServiceId,
+        commit_hash: &str,
+    ) -> Result<(), CommitHashStoreError> {
+        let current_root_bytes = hex::parse_hex(commit_hash).map_err(|e| {
+            InvalidArgumentError::new(
+                "commit_hash".into(),
+                format!("The commit hash provided is is invalid: {}", e),
```
agunde406, line 73: is is -> is

agunde406 (changes requested):

services/scabbard/libscabbard/src/store/lmdb/mod.rs
peterschwarz: fixed

services/scabbard/libscabbard/Cargo.toml
```diff
@@ -65,6 +65,7 @@ stable = [
   "events",
   "rest-api",
   "rest-api-actix",
+  "lmdb",
```
agunde406, line 68: This should be in alphabetical order

agunde406 (changes requested):

agunde406 (approved):

agunde406 (approved):

services/scabbard/libscabbard/src/store/lmdb/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+impl LmdbCommitHashStore {
+    /// Constructs a new commit log store around an LMDB database instance.
+    pub fn new(db: LmdbDatabase) -> Self {
+        Self { db }
+    }
+}
+
+impl CommitHashStore for LmdbCommitHashStore {
```
vaporos, line 42: Because this implementation ignores ServiceId, it does not fully implement the intent of the trait. This would require higher-level code to include conditionals based on implementation.

If we keep this approach, maybe we should add a service_id argument to new() and then make sure the service id passed on these calls matches it.

Another approach would be to let this implementation handle more than a single lmbd database instance. If this store were filesystem-aware and knew where the lmdb database files resided, this implementation could instantiate LmdbDatabase when it needed it. This could also be split out into a factory pattern if we wanted something else to answer "return the LmdbDatabase for this service id".

services/scabbard/libscabbard/src/store/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+
+mod error;
+
+pub use error::CommitHashStoreError;
+
+/// Canonical Service Identifier
+///
+/// This identifier includes both the circuit and the service ID's.  This ensures that an entry for
+/// an individual scabbard service will be scoped by the circuit.
+pub struct ServiceId<'a> {
```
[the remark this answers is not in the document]
peterschwarz: Removed

services/scabbard/libscabbard/src/store/lmdb/mod.rs
peterschwarz: I have simplified the trait remove the use the ServiceId argument.  Instances will be localized to a specific circuit/service ID pair

services/scabbard/libscabbard/src/store/error.rs
```diff
@@ -0,0 +1,76 @@
...
+//! Error types and logic for CommitHashStores.
+
+use std::error::Error;
+use std::fmt::Display;
+
+use splinter::error::{
+    InternalError, InvalidArgumentError, InvalidStateError, ResourceTemporarilyUnavailableError,
+};
+
+/// Error type for the [CommitHashStore](super::CommitLogStore) trait.
```
agunde406, line 24: There are several places in the code and commit message that mentioned CommitLogStore, should they all be changed to CommitHashStore?

agunde406 (changes requested):

services/scabbard/libscabbard/src/store/error.rs
peterschwarz: Fixed

agunde406 (approved):

vaporos (approved):

PR #1506, opened 2021-09-10
Remove clones from InfluxRecorder

peterschwarz:
This change removes clones from the influx recorder, reducing the clones of entire strings. While small, metrics will happen frequently as more metrics are added, so its performance impact should be as minimal as possible.


agunde406 (approved):
Tested

isabeltomb (approved):

PR #1510, opened 2021-09-14
Challenge authorization stabilization updates 

agunde406:


libsplinter/src/network/auth/handlers/mod.rs
```diff
@@ -106,9 +106,9 @@ impl Handler for AuthorizationErrorHandler {
             AuthorizationError::AuthorizationRejected(err_msg) => {
                 match self.auth_manager.next_remote_state(
```
peterschwarz, line 107: Should this also be renamed?

agunde406: ah yes, missed those will update

peterschwarz (approved):

vaporos (approved):

PR #1519, opened 2021-09-19
Scabbard storage config

peterschwarz:
This PR modifies the scabbard factory to create the required storage configuration (either the stores themselves or the exact lmdb files) at service-create time.

* It makes the ScabbardFactoryBuilder the only way to construct the factory (this is needed due to the changes in possibilities for storage configuration).
* It re-arranges many of the sawtooth feature dependencies, as several experimental features were being included with a stable feature.
* Adds two features for the specific db types in order to make use of the updated receipt store.

### DIFF
--- a/services/scabbard/libscabbard/Cargo.toml
+++ b/services/scabbard/libscabbard/Cargo.toml
@@ -75,7 +75,6 @@ experimental = [
   # The following features are experimental:
   "commit-store",
   "diesel-receipt-store",
-  "factory-builder",
   "https",
   "metrics",
   "postgres",
@@ -88,7 +87,6 @@ client-reqwest = ["client", "reqwest"]
 commit-store = []
 diesel-receipt-store = ["diesel", "sawtooth/transaction-receipt-store-lmdb"]
 events = ["splinter/events"]
-factory-builder = []
 https = []
 lmdb = []
 postgres = ["diesel/postgres", "diesel_migrations", "sawtooth/postgres"]
--- a/services/scabbard/libscabbard/Cargo.toml
+++ b/services/scabbard/libscabbard/Cargo.toml
@@ -45,7 +45,7 @@ transact = { version = "0.3", features = ["sawtooth-compat"] }
 [dependencies.sawtooth]
 version = "0.6.6"
 default-features = false
-features = ["diesel", "lmdb-store", "postgres", "receipt-store", "sqlite", "transaction-receipt-store"]
+features = ["lmdb-store", "receipt-store", "transaction-receipt-store"]
 
 [dev-dependencies]
 tempdir = "0.3"
@@ -86,13 +86,13 @@ authorization = ["splinter/authorization"]
 client = []
 client-reqwest = ["client", "reqwest"]
 commit-store = []
-diesel-receipt-store = ["diesel"]
+diesel-receipt-store = ["diesel", "sawtooth/transaction-receipt-store-lmdb"]
 events = ["splinter/events"]
 factory-builder = []
 https = []
 lmdb = []
-postgres = ["diesel/postgres", "diesel_migrations"]
+postgres = ["diesel/postgres", "diesel_migrations", "sawtooth/postgres"]
 rest-api = ["futures", "splinter/rest-api"]
 rest-api-actix = ["actix-web", "splinter/rest-api-actix"]
 service-arg-validation = ["splinter/service-arg-validation"]
-sqlite = ["diesel/sqlite", "diesel_migrations"]
+sqlite = ["diesel/sqlite", "diesel_migrations", "sawtooth/sqlite"]
--- a/services/scabbard/libscabbard/src/service/factory.rs
+++ b/services/scabbard/libscabbard/src/service/factory.rs
@@ -140,28 +140,6 @@ pub struct ScabbardFactory {
     signature_verifier_factory: Arc<Mutex<Box<dyn VerifierFactory>>>,
 }
 
-impl ScabbardFactory {
-    pub fn new(
-        state_db_dir: Option<String>,
-        state_db_size: Option<usize>,
-        receipt_db_dir: Option<String>,
-        receipt_db_size: Option<usize>,
-        #[cfg(feature = "diesel-receipt-store")] receipt_db_url: String,
-        signature_verifier_factory: Arc<Mutex<Box<dyn VerifierFactory>>>,
-    ) -> Self {
-        ScabbardFactory {
-            service_types: vec![SERVICE_TYPE.into()],
-            state_db_dir: state_db_dir.unwrap_or_else(|| DEFAULT_STATE_DB_DIR.into()),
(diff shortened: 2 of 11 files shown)


services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -34,7 +34,92 @@ const DEFAULT_STATE_DB_SIZE: usize = 1 << 30; // 1024 ** 3
...
+        any(feature = "postgres", feature = "sqlite")
+    ))]
+    fn connection_uri(connection_uri: &str) -> Self {
+        let connection_uri =
+            if cfg!(feature = "postgres") && connection_uri.starts_with("postgres://") {
+                #[cfg(feature = "postgres")]
+                {
+                    ConnectionUri::Postgres(connection_uri.into())
+                }
+                // This cannot be reached do to the if test failing without this feature.
```
agunde406, line 95: The comment is confusing, please reword it

services/scabbard/libscabbard/src/service/state.rs
```diff
@@ -77,35 +80,24 @@ pub struct ScabbardState {
...
     ) -> Result<Self, ScabbardStateError> {
         // Initialize the database
         let mut indexes = INDEXES.to_vec();
         indexes.push(CURRENT_STATE_ROOT_INDEX);
-        let state_db_file = state_db_path.to_path_buf();
-        let receipt_db_file = receipt_db_path.to_path_buf();
         let state_db_path = state_db_path.with_extension("lmdb");
-        #[cfg(not(feature = "diesel-receipt-store"))]
-        let receipt_db_path = receipt_db_path.with_extension("lmdb");
+        eprintln!("state db path: {:?}", state_db_path);
```
agunde406, line 100: Remove eprintln

agunde406 (changes requested):

services/scabbard/libscabbard/Cargo.toml
```diff
@@ -86,13 +86,13 @@ authorization = ["splinter/authorization"]
...
 client-reqwest = ["client", "reqwest"]
 commit-store = []
-diesel-receipt-store = ["diesel"]
+diesel-receipt-store = ["diesel", "sawtooth/transaction-receipt-store-lmdb"]
 events = ["splinter/events"]
 factory-builder = []
 https = []
 lmdb = []
-postgres = ["diesel/postgres", "diesel_migrations"]
+postgres = ["diesel/postgres", "diesel_migrations", "sawtooth/diesel", "sawtooth/postgres"]
```
vaporos, line 92: We should be able to simplify this such that we can add "sawtooth/postgres" but not "sawtooth/diesel" since "sawtooth/postgres" should pull in diesel.

services/scabbard/libscabbard/src/service/factory.rs
peterschwarz: Repaced this with an Unknown type when sqlite is not enabled, which simplifies conversion with a  match statement

services/scabbard/libscabbard/src/service/state.rs
peterschwarz: Removed

services/scabbard/libscabbard/Cargo.toml
peterschwarz: Yeah, removed.

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -47,7 +47,6 @@ pub struct ScabbardFactoryBuilder {
     signature_verifier_factory: Option<Arc<Mutex<Box<dyn VerifierFactory>>>>,
 }
 
-#[cfg(feature = "factory-builder")]
```
vaporos, line 50: Doesn't this have to match the cfg on the struct itself?

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -34,7 +34,88 @@ const DEFAULT_STATE_DB_SIZE: usize = 1 << 30; // 1024 ** 3
 const DEFAULT_RECEIPT_DB_DIR: &str = "/var/lib/splinter";
 const DEFAULT_RECEIPT_DB_SIZE: usize = 1 << 30; // 1024 ** 3
 
-#[cfg(feature = "factory-builder")]
```
vaporos, line 37: This technically should be in the previous commit that removed factory-builder. (See other "this should match the struct comment".)

services/scabbard/libscabbard/src/service/factory.rs
peterschwarz: Looks like I missed that one

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -34,7 +34,88 @@ const DEFAULT_STATE_DB_SIZE: usize = 1 << 30; // 1024 ** 3
...
+#[cfg(feature = "diesel-receipt-store")]
+type ScabbardReceiptStore = Arc<RwLock<dyn ReceiptStore>>;
+
+#[cfg(feature = "diesel-receipt-store")]
+pub enum ConnectionUri {
+    #[cfg(feature = "postgres")]
+    Postgres(Box<str>),
+    #[cfg(feature = "sqlite")]
+    Sqlite(Box<str>),
+    #[cfg(not(feature = "sqlite"))]
```
vaporos, line 63: Should this be #[cfg(not(all(feature = "sqlite", feature = "postgres"))]?

vaporos: Nevermind, I see the approach now is about the default.

services/scabbard/libscabbard/src/service/factory.rs
peterschwarz: Yep, missed that one

agunde406 (approved):
Tested

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -491,15 +719,35 @@ mod tests {
...
-            Some(1024 * 1024),
-            Some("/tmp".into()),
-            Some(1024 * 1024),
-            #[cfg(feature = "diesel-receipt-store")]
-            ":memory:".to_string(),
-            Arc::new(Mutex::new(Box::new(Secp256k1Context::new()))),
-        )
+        #[cfg(not(all(feature = "diesel-receipt-store", feature = "sqlite")))]
+        let receipt_store_factory_config = ScabbardFactoryStorageConfig::Lmdb {
+            db_dir: "/tmp".into(),
```
vaporos, line 724: Instead of using /tmp here, we should use TempDir to create a directory for us - https://docs.rs/tempdir/0.3.7/tempdir/

peterschwarz: I will add a commit, as this was using the same config as the original test.

vaporos (approved):

agunde406 (approved):

PR #1521, opened 2021-09-20
Node file block clean up

Caleb-Hill:


splinterd/src/node_id.rs
```diff
@@ -0,0 +1,64 @@
+use rand::Rng;
```
agunde406, line 1: Missing copyright

splinterd/src/daemon.rs
```diff
@@ -246,10 +247,17 @@ impl SplinterDaemon {
         let admin_service_verifier = secp256k1_context.new_verifier();
         let auth_config_verifier = secp256k1_context.new_verifier();
         let signing_context = Arc::new(Mutex::new(secp256k1_context));
+        let node_id: String = crate::node_id::get_node_id(
```
agunde406, line 250: Add a use statement above for get_node_id instead of useing the whole crate::node_id::get_node_id

agunde406 (changes requested):

splinterd/src/daemon.rs
```diff
@@ -1272,9 +1283,12 @@ impl SplinterDaemonBuilder {
             CreateError::MissingRequiredField("Missing field: heartbeat".to_string())
         })?;
 
-        let node_id = self.node_id.ok_or_else(|| {
-            CreateError::MissingRequiredField("Missing field: node_id".to_string())
-        })?;
+        let node_id = self
```
agunde406, line 1286: This should remain an option here and not be generated if None, that happens in your get_node_id() function that is called in SplinterDaemon.start(). Not this means that it should not be returning an error here.

splinterd/src/daemon.rs
```diff
@@ -1318,9 +1318,12 @@ impl SplinterDaemonBuilder {
             CreateError::MissingRequiredField("Missing field: initial_peers".to_string())
         })?;
 
-        let display_name = self.display_name.ok_or_else(|| {
-            CreateError::MissingRequiredField("Missing field: display_name".to_string())
-        })?;
+        let display_name = self
```
agunde406, line 1321: This should also remain an option and use the node id that is created in SplinterDaemon.start() if display name is not set.

agunde406 (changes requested):

agunde406 (approved):
Tested

peterschwarz (approved):

PR #1524, opened 2021-09-20
Add check for io errors opening sqlite database

Caleb-Hill:
Sqlite connection pool initialization fails clumsily if the database
file can't be opened. This commit will force an early return with both
the filename and Error in the message.

### DIFF
--- a/cli/src/action/database/sqlite.rs
+++ b/cli/src/action/database/sqlite.rs
@@ -30,10 +30,11 @@ use super::SplinterEnvironment;
 use crate::error::CliError;
 
 const DEFAULT_SQLITE: &str = "splinter_state.db";
+const MEMORY: &str = ":memory:";
 
 /// Run the sqlite migrations against the provided connection string
 pub fn sqlite_migrations(connection_string: String) -> Result<(), CliError> {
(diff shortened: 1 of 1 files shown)


cli/src/action/database/sqlite.rs
```diff
@@ -33,6 +33,16 @@ const DEFAULT_SQLITE: &str = "splinter_state.db";
 
 /// Run the sqlite migrations against the provided connection string
 pub fn sqlite_migrations(connection_string: String) -> Result<(), CliError> {
+    if let Err(err) = OpenOptions::new()
```
vaporos, line 36: Should exclude the special connection string ":memory:" from this check.

vaporos (changes requested):

agunde406 (approved):
Tested

peterschwarz (changes requested):

peterschwarz (approved):

cli/src/action/database/sqlite.rs
Caleb-Hill: It is now excluded

vaporos (approved):

PR #1529, opened 2021-09-24
Enfroce protocol version is valid for proposal during vote

agunde406:
It is possible that a proposal could have been stored in state
and when a vote has been received the agreed upon protocol
version has changed. This is possible due to node restarts/upgrades.

If the agreed upon protocol no longer supports the proposal,
the vote is rejected. The proposal is also rejected. 

To support rejecting the proposal , the peers are moved into a temporary vec,
that includes an Instant for when they are added to the list.
After 10 seconds, the peers can be dropped. This is triggered by
either propose_change or create_proposal, both functions being
called by consensus. This ensures that the peers are not
dropped until the proposal has been handled. This also enables
the integration test that test rejecting a proposal.

### DIFF
--- a/libsplinter/src/admin/service/shared.rs
+++ b/libsplinter/src/admin/service/shared.rs
@@ -2892,14 +2892,40 @@ impl AdminServiceShared {
     }
 
     fn validate_circuit_vote(
-        &self,
+        &mut self,
         proposal_vote: &CircuitProposalVote,
         signer_public_key: &[u8],
         circuit_proposal: &StoreProposal,
         node_id: &str,
         protocol: u32,
     ) -> Result<(), AdminSharedError> {
-        self.validate_protocol_vote(circuit_proposal, protocol)?;
+        if let Err(err) = self.validate_protocol_vote(circuit_proposal, protocol) {
+            warn!(
+                "Removing proposal for {}, no longer valid for agreed protocol version: {}",
+                circuit_proposal.circuit_id(),
+                protocol
+            );
+
+            self.remove_proposal(circuit_proposal.circuit_id())?;
+            self.peers_to_be_removed.push((
+                Instant::now(),
+                circuit_proposal
+                    .circuit()
+                    .list_tokens(
+                        #[cfg(feature = "challenge-authorization")]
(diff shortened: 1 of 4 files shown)


peterschwarz (approved):

vaporos:
This feels like we are confusing two different things - the protocol version of the communication and the version of the data. It feels like proposals should have the version number of the circuit which will result (data protocol version); which would presumably not be invalidated by the communication protocol's version.

libsplinter/src/admin/service/shared.rs
```diff
@@ -319,6 +325,23 @@ impl AdminServiceShared {
         }
     }
 
+    /// Remove the peers who have been held onto for the default holding time. In some cases
+    /// peers should not be removed when a proposal is removed because it can cause messages to be
+    /// dropped. Instead, the will be dropped after 10 seconds or when the cleanlup_help_peer_refs
```
vaporos, line 330: Typo - cleanlup

libsplinter/src/admin/service/shared.rs
```diff
@@ -319,6 +325,23 @@ impl AdminServiceShared {
         }
     }
 
+    /// Remove the peers who have been held onto for the default holding time. In some cases
```
vaporos, line 328: This feels like a workaround to a more serious peer ref handling bug.

libsplinter/src/admin/service/shared.rs
```diff
@@ -2892,14 +2892,40 @@ impl AdminServiceShared {
...
-        self.validate_protocol_vote(circuit_proposal, protocol)?;
+        if let Err(err) = self.validate_protocol_vote(circuit_proposal, protocol) {
+            warn!(
+                "Removing proposal for {}, no longer valid for agreed protocol version: {}",
+                circuit_proposal.circuit_id(),
+                protocol
+            );
+
+            self.remove_proposal(circuit_proposal.circuit_id())?;
+            self.peers_to_be_removed.push((
```
vaporos, line 2910: Why remove these here automatically? I don't think this would be expected behavior from a user perspective.

PR #1538, opened 2021-09-30
Sqlite migrations check error type

isabeltomb:
The `sqlite_migrations` method has check that will return an error if
the file being used for the splinter state database does not exist.
However if the file does not exist it will be created later by sqlite
and therefore no error should be returned in this situation. This
commit adds a match statement to this check that will ensure an error
is not returned if the error type is `std::io::ErrorKind::NotFound`.


Caleb-Hill (approved):
Tested

peterschwarz (approved):
Tested

PR #1540, opened 2021-09-30
scabbard db upgrade

peterschwarz:
To test:

Create a scabbard circuit between two nodes.

Apply one transaction against the circuit:

```
$ scabbard cr create smallbank --owners $(cat ~/.cylinder/keys/$USER.pub) --service-id $CIRCUIT_ID::a000 --url $SPLINTER_URL
```

Run 

```
$ splinter upgrade -vv
...
Upgraded scabbard service CIRCUIT_ID::a000
```
where `CIRCUIT_ID` is the circuit created above.

This can be verified by using the sqlite client:

```
$ sqlite3 -header $SPLINTER_HOME/data/splinter_state.db
> select * from scabbard_commit_log;
circuit_id|service_id|commit_hash
CIRCUIT_ID|a000|859ac1f96d790dd660cfee2e788eeb4b8f46c12a5ff49504ad88d05fff412cfa
```

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -51,7 +51,7 @@ features = ["admin-service", "registry"]
 
 [dependencies.scabbard]
 path = "../services/scabbard/libscabbard"
-features = ["commit-store", "postgres", "sqlite"]
+features = ["commit-store", "postgres", "sqlite", "lmdb"]
 optional = true
 
 [dependencies.transact]
--- a/cli/src/action/database/upgrade/scabbard.rs
+++ b/cli/src/action/database/upgrade/scabbard.rs
@@ -1,0 +1,216 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Provides scabbard state upgrade functionality
+
+use std::path::Path;
+
+use diesel::r2d2::{ConnectionManager, Pool};
+use scabbard::store::{
+    diesel::DieselCommitHashStore,
+    transact::{factory::LmdbDatabaseFactory, TransactCommitHashStore},
+    CommitHashStore,
+};
+use splinter::{
+    admin::store::{diesel::DieselAdminServiceStore, AdminServiceStore},
+    error::InternalError,
+    node_id::store::{diesel::DieselNodeIdStore, NodeIdStore},
+    store::ConnectionUri,
+};
+
+use super::error::UpgradeError;
+
+/// Migrate all of the service state's current commit hashes to the [`CommitHashStore`].
+pub fn upgrade_scabbard_commit_hash_state(
+    state_dir: &Path,
+    database_uri: &ConnectionUri,
+) -> Result<(), UpgradeError> {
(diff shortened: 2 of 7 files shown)


cli/src/action/database/upgrade/scabbard.rs
```diff
@@ -0,0 +1,231 @@
...
+
+struct UpgradeStores {
+    admin_service_store: Box<dyn AdminServiceStore>,
+    commit_hash_store_factory: CommitHashStoreFactory,
+    node_id_store: Box<dyn NodeIdStore>,
+}
+
+impl UpgradeStores {
+    fn new(database_uri: &ConnectionUri) -> Result<Self, UpgradeError> {
+        match database_uri {
```
vaporos, line 111: This match would fail to compile if postgres, sqlite, and memory were all disabled. We should make sure at a minimum that a reasonable compilation error is given.

cli/src/action/database/upgrade/scabbard.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+            #[cfg(feature = "memory")]
+            _ => Err(UpgradeError::InvalidState(InvalidStateError::with_message(
+                format!("Unsupported database: {}", database_uri),
+            ))),
+        }
+    }
+}
+
+enum CommitHashStoreFactory {
```
vaporos, line 171: This should be implemented in a way which provides the benefit of abstraction/encapsulation. Currently, having this factory enum provides no opportunity for the code using it to not be postgres/sqlite-aware. If it were a trait on two different implementations, then only the creation of the factory would need to be postgres/sqlite-aware.

vaporos: I guess maybe some of the benefits of the abstraction are here, and the biggest issues would be a) the abstraction can be circumvented and the pool accessed directly (thus making the caller postgres/sqlite-aware); b) with this pattern, the implementations are fixed (you could not write another of these factory implementations like you can an abstract factory), which is perhaps ok here but wouldn't be in other places.

cli/Cargo.toml
```diff
@@ -51,6 +51,7 @@ features = ["admin-service", "registry"]
 
 [dependencies.scabbard]
 path = "../services/scabbard/libscabbard"
+default-features = false
```
vaporos, line 54: Is this necessary (or should we adjust libscabbard's default feature list instead)?

peterschwarz: This does ensure that we don't pull in anything that might be added to default in the future.

cli/src/action/database/upgrade/scabbard.rs
peterschwarz: Should we configure it such that the upgrade command is only available if at least sqlite or postgres are enabled?

cli/Cargo.toml
vaporos: Not sure what we would be optimizing for; if size, then I think the compiler is good enough it will exclude code that isn't used anyway. From that perspective, the only reason for ever turning off default features is if default features pulled in a lot of additional crates to compile.

peterschwarz: Dropped this commit

cli/src/action/database/upgrade/scabbard.rs
peterschwarz: I've replaced this with a similar pattern to libsplinter.

vaporos (approved):

cli/src/action/database/upgrade/scabbard.rs
vaporos: I'm not sure how we would implement that cleanly. 

The default branch - "Unsupported database:" -- that would seem to be a reasonable error from the CLI if we have the databases disabled. But that's only on if memory is on, which I assume is an artifact of it being an enum. I wonder if this enum should be declared non-exhaustive.

Maybe something to handle in a different PR though, or not at all, since this doesn't actually break the build and we are compiling with --no-default-features. You would have to be trying to recreate the feature conditions. Though if we remove node-id-upgrade feature then that may expose it more directly.

agunde406 (approved):

PR #1547, opened 2021-10-04
Update Sawtooth-lib dependency

isabeltomb:
- Update all sawtooth dependencies minimum required version 0.6.7
- Update the scabbard "diesel-receipt-store" feature's sawtooth/transaction-receipt-store-lmdb dependency to sawtooth/lmdb to match the feature name change in sawtooth-lib
- Update all instances of `DieselReceiptStore::new(pool)` to include the added `service_id` argument. The service ID, formatted "circuit_id::service_id" is used for this arg in scabbard


services/scabbard/libscabbard/Cargo.toml
```diff
@@ -84,7 +84,7 @@ experimental = [
 authorization = ["splinter/authorization"]
 client = []
 client-reqwest = ["client", "reqwest"]
-diesel-receipt-store = ["diesel", "sawtooth/transaction-receipt-store-lmdb"]
+diesel-receipt-store = ["diesel", "sawtooth/lmdb"]
```
peterschwarz, line 87: Should this go under the lmdb feature?

isabeltomb: Renamed "diesel-receipt-store" to "receipt-store" because we put both the diesel backed and lmdb backed receipt store impls behind the feature

cli/Cargo.toml
```diff
@@ -101,13 +101,15 @@ https-certs = []
 postgres = [
     "diesel/postgres",
     "splinter/postgres",
+    "sawtooth/postgres",
```
peterschwarz, line 104: It also should get removed from the sawtooth dependency definition

vaporos (approved):

PR #1550, opened 2021-10-04
Negative REST API security tests

isabeltomb:
- Adds the ability to add authorization handlers to the testing framework
- Adds a memory backed implementation of `RoleBasedAuthorizationStore` to be used in tests
- Adds a test that attempts to access all REST API endpoints provided by the testing framework that require permissions and checks that a 401 is returned by each when no permissions are configured.


splinterd/src/node/running/mod.rs
```diff
@@ -85,11 +85,17 @@ impl Node {
         self.network_subsystem.network_endpoints()
     }
 
-    pub fn admin_service_client(self: &Node) -> Box<dyn AdminServiceClient> {
-        Box::new(ReqwestAdminServiceClient::new(
-            format!("http://localhost:{}", self.rest_api_port),
-            "foo".to_string(),
-        ))
+    pub fn admin_service_client(self: &Node, auth: Option<String>) -> Box<dyn AdminServiceClient> {
```
vaporos, line 88: This method feels like the wrong place to inject the auth string (the size of the patch as evidence).

PR #1559, opened 2021-10-06
Updates for `service-arg-validation` in scabbard

agunde406:


services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -351,6 +351,15 @@ impl ServiceArgValidator for ScabbardArgValidator {
             )
         })?;
 
+        for service in peer_services {
+            if service.is_empty() {
+                return Err(InvalidArgumentError::new(
+                    "peer_services".into(),
+                    "service cannot be empty".into(),
```
peterschwarz, line 358: Using singular "service" here.  Suggest "must provide at least one service ID"

services/scabbard/libscabbard/src/service/factory.rs
```diff
@@ -363,6 +372,13 @@ impl ServiceArgValidator for ScabbardArgValidator {
         })?;
 
         for key in admin_keys {
+            if key.is_empty() {
+                return Err(InvalidArgumentError::new(
+                    "admin_keys".into(),
+                    "key cannot be empty".into(),
```
peterschwarz, line 378: Similar singular usage here

services/scabbard/libscabbard/src/service/factory.rs
agunde406: updated

Caleb-Hill (approved):

PR #1568, opened 2021-10-13
Stabilize libscabbard "receipt-store"

isabeltomb:
Stabilize the "receipt-store" feature in libscabbard by removing it

### DIFF
--- a/services/scabbard/libscabbard/src/service/rest_api/actix/state.rs
+++ b/services/scabbard/libscabbard/src/service/rest_api/actix/state.rs
@@ -95,6 +95,7 @@ pub fn make_get_state_with_prefix_endpoint() -> ServiceEndpoint {
     }
 }
 
+#[cfg(feature = "sqlite")]
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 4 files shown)


services/scabbard/libscabbard/Cargo.toml
```diff
@@ -45,7 +45,7 @@ transact = { version = "0.3.10", features = ["sawtooth-compat"] }
 [dependencies.sawtooth]
 version = "0.6.7"
 default-features = false
-features = ["lmdb-store", "receipt-store", "transaction-receipt-store"]
+features = ["lmdb", "transaction-receipt-store"]
```
vaporos, line 48: This change - "lmdb-store" to "lmdb" looks unintentional.

services/scabbard/libscabbard/src/service/factory/factory_database.rs
```diff
@@ -44,9 +44,6 @@ use crate::store::{
 
 const DEFAULT_LMDB_DIR: &str = "/var/lib/splinter";
 
-#[cfg(feature = "receipt-store")]
-type ScabbardReceiptStore = Arc<RwLock<dyn ReceiptStore>>;
```
vaporos, line 48: This change looks unrelated to removing the feature.

vaporos (changes requested):

services/scabbard/libscabbard/src/service/factory/factory_database.rs
isabeltomb: I removed it because the type was originally created so that separate args weren't needed for the two different receipt stores. It was also removed from `factory_static_configuration.rs` but you're right, in this file it isn't directly related so I changed this back

services/scabbard/libscabbard/Cargo.toml
vaporos: Ok, makes sense.

services/scabbard/libscabbard/src/service/factory/factory_static_configuration.rs
```diff
@@ -457,95 +432,65 @@ impl ServiceFactory for ScabbardFactory {
...
-                feature = "receipt-store",
-                not(any(feature = "postgres", feature = "sqlite"))
-            ))]
-            _ => {
-                return Err(FactoryCreateError::Internal(
-                    "receipt-store was enabled, but without any databases".into(),
-                ))
-            }
-        };
+        let (receipt_store, receipt_purge): (Arc<RwLock<dyn ReceiptStore>>, _) =
```
vaporos, line 435: This also looks unrelated, removing the use of ScabbardReceiptStore. Should at minimum be a separate commit.

agunde406 (approved):

peterschwarz (approved):

PR #1577, opened 2021-10-15
Stabilize scabbard-receipt-store in CLI

isabeltomb:
Stabilize the "scabbard-receipt-store" feature in splinter CLI by removing it


peterschwarz (approved):
I think the logging messages that this guards should be removed in a followup PR, as those migrations were pushed down into the `scabbard::migrations` functions.

PR #1579, opened 2021-10-15
Update authorization protocol and admin message handler to use PublicKey

agunde406:
This is a part of on going work to replace the use of Strings, Vec<Strings>, and Vec<u8> to represent public keys with the splinter PublicKey struct internally.


vaporos (dismissed):

libsplinter/src/network/auth/authorization/challenge/handlers.rs
```diff
@@ -323,7 +323,7 @@ impl Handler for AuthChallengeSubmitRequestHandler {
                 .verify(
                     &self.nonce,
                     &Signature::new(request.signature.to_vec()),
-                    &PublicKey::new(request.public_key.to_vec()),
+                    &PublicKey::new(request.public_key.clone().into_bytes()),
```
peterschwarz, line 326: What’s being cloned here?

libsplinter/src/network/auth/authorization/challenge/handlers.rs
```diff
@@ -1013,10 +1013,12 @@ mod tests {
...
                 submit_requests: vec![SubmitRequest {
-                    public_key: public_key::PublicKey::from_bytes(other_signer
-                        .public_key()
-                        .expect("Unable to get public key")
-                        .into_bytes()),
+                    public_key: public_key::PublicKey::from_bytes(
+                        other_signer
+                            .public_key()
+                            .expect("Unable to get public key")
+                            .into_bytes(),
```
peterschwarz, line 1020: Should we have a `From` implementation?

libsplinter/src/network/auth/authorization/challenge/handlers.rs
agunde406: Its the splinter::public_key::PublicKey stored in the request. into_bytes consumes the public key so it needs to be cloned to convert it to cylinder::PublicKey. Could be changed to `&PublicKey::new(request.public_key.as_slice().to_vec()),`  instead

PR #1584, opened 2021-10-18
Check log file is writeable

Caleb-Hill:
Files are often unwritable, this checks to make sure the log files exist and are writable by the service before they are used.

### DIFF
--- a/splinterd/src/main.rs
+++ b/splinterd/src/main.rs
@@ -28,6 +28,11 @@ pub mod node_id;
 mod routes;
 mod transport;
 
+#[cfg(feature = "log-config")]
+use std::fs::OpenOptions;
+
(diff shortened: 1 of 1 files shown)


rbuysse (dismissed):

rbuysse (approved):

splinterd/src/main.rs
```diff
@@ -665,6 +670,39 @@ fn start_daemon(matches: ArgMatches, _log_handle: Handle) -> Result<(), UserErro
     #[cfg(feature = "log-config")]
     {
         let appenders = if let Some(appenders) = config.appenders() {
+            let check_file_readability = |path: &Path| {
```
peterschwarz, line 673: Perhaps we could move all of the logging setup into its own module in a future PR.

PR #1591, opened 2021-10-21
Break apart scabbard-migration features

agunde406:
The scabbard migrations were only being run if the scabbard-migrations
feature was enabled which was experimental. This means the transaction
receipt migrations were not being run on stable.

This commit fixes the issues by moving pulling in scabbard/postgres
and scabbard/sqlite with the associated postgres and sqlite features.

And scabbard/database-support and scabbard/lmdb are now pulled in
by scabbard-migrations.

The migrations from scabbard will always be run.


rbuysse:
Running packages built from this commit on the LR networks seems to fix a regression I was seeing.

PR #1602, opened 2021-10-25
Splinter keygen group option

Caleb-Hill:
This adds the `--group` flag to the splinter keygen command.

### DIFF
--- a/cli/src/action/keygen.rs
+++ b/cli/src/action/keygen.rs
@@ -27,7 +27,7 @@ use std::os::unix::fs::MetadataExt;
 use clap::ArgMatches;
 use cylinder::{secp256k1::Secp256k1Context, Context};
 use cylinder::{PrivateKey, PublicKey};
-use users::get_group_by_name;
+use users::{get_group_by_gid, get_group_by_name};
 
 use crate::error::CliError;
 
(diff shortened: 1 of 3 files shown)


cli/src/action/keygen.rs
```diff
@@ -65,6 +65,18 @@ impl FromStr for GroupOptions {
 impl Action for KeyGenAction {
     fn run<'a>(&mut self, arg_matches: Option<&ArgMatches<'a>>) -> Result<(), CliError> {
         let args = arg_matches.ok_or(CliError::RequiresArgs)?;
+        let group: Option<GroupOptions> = args
+            .value_of("group")
+            .map(|s| Some(GroupOptions::from_str(s)))
```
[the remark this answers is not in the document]
Caleb-Hill: No reason, fixed

cli/src/action/keygen.rs
```diff
@@ -38,6 +39,29 @@ const DEFAULT_SYSTEM_KEY_NAME: &str = "splinterd";
...
+    Auto,
+    Named(String),
+    GroupID(u32),
+}
+
+impl FromStr for GroupOptions {
+    type Err = CliError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "auto" => Ok(GroupOptions::Auto),
```
peterschwarz, line 59: If this is only for parsing the clap args, you should just parse this value there.  See the recent change with ScabbardState

peterschwarz (changes requested):

cli/src/action/keygen.rs
Caleb-Hill: Fixed

cli/src/action/keygen.rs
```diff
@@ -148,7 +149,7 @@ fn write_keys(
     public_key_path: PathBuf,
     force_create: bool,
     skip_create: bool,
-    change_permissions: bool,
+    group: Option<GroupOptions>,
```
peterschwarz: As this previously was always passing true, should this still be an option?  If no option is provided on the command line, the behaviour of this command will change. Given that `Auto` is the previous behaviour, perhaps this should not be optional and the arg should be or-elsed to that.

peterschwarz: After some discussion, this bahaviour change is intentional.

peterschwarz (approved):

vaporos (approved):

PR #1614, opened 2021-10-27
Detect incorrect state daemon

peterschwarz:
This PR changes the ScabbardFactory builder to fail if LMDB state is not enabled, but there are any existing LMDB files in the state directory.

To test:

1. Start a pair of nodes, with at least one node providing the argument --scabbard-state lmdb
2. Create a circuit and execute a single transaction
3. Stop the node with the above --scabbard-state argument
4. Restart the node without that argument

You should see an error like:

```
[2021-10-28 10:10:27.114] T[main] ERROR [splinterd] Failed to start daemon, unable to start the Splinter daemon: there has been a user error: LMDB database files exist, but LMDB storage is not enabled
```


isabeltomb (approved):
Tested

PR #1622, opened 2021-10-29
Remove `ScabbardReceiptStore` types

isabeltomb:
Remove unnecessary `ScabbardReceiptStore` types. This type was initially
created so that the receipt store implementation used by scabbard could
change depending on whether or not the "receipt-store" experimental
feature was enabled. The "receipt-store" feature has been stabilized
and removed so the `ScabbardReceiptStore` type is no longer needed.

### DIFF
--- a/services/scabbard/libscabbard/src/service/factory/mod.rs
+++ b/services/scabbard/libscabbard/src/service/factory/mod.rs
@@ -735,14 +735,14 @@ impl ScabbardFactory {
             self.create_sql_merkle_state_purge_handle(circuit_id, &service_id),
         );
 
-        let (receipt_store, commit_hash_store): (Arc<RwLock<dyn ReceiptStore>>, Box<dyn CommitHashStore>) =
+        let (receipt_store, commit_hash_store): (Arc<dyn ReceiptStore>, Box<dyn CommitHashStore>) =
             match &self.store_factory_config {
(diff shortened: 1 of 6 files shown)


agunde406 (dismissed):

services/scabbard/libscabbard/src/service/factory/factory_database.rs
```diff
@@ -424,33 +422,35 @@ impl ServiceFactory for ScabbardFactory {
...
-                    ))),
-                    Box::new(DieselCommitHashStore::new(
-                        pool.clone(),
-                        circuit_id,
-                        &service_id,
-                    )),
-                ),
-            };
+        let (receipt_store, commit_hash_store): (
+            Arc<RwLock<dyn ReceiptStore>>,
```
peterschwarz, line 426: Since you’re removing the type, might as well remove the arc-mutex combo here as it’s not necessary here (in another commit)

vaporos (approved):

services/scabbard/libscabbard/src/service/factory/factory_database.rs
isabeltomb: Added a commit to remove the RwLock

peterschwarz (approved):

agunde406 (approved):

PR #1660, opened 2021-11-16
Update several dependencies in the splinter repository

vaporos:
Updates log, gag, dirs, flexi-logger, and reqwest.


agunde406 (approved):

vaporos:
Removed flexi-logger commit from this PR as we investigate "[flexi_logger][ERRCODE::Time] flexi_logger has to work with UTC rather than with local time, caused by The system's UTC offset could not be determined"

Caleb-Hill (approved):

agunde406 (approved):

peterschwarz (approved):

PR #1677, opened 2021-11-29
Correct check for SQL merkle trees existence

agunde406:
On creation there is always a "default" tree, this should
not be considered existing SQL Merkle Radix trees if LMDB is
enabled.


vaporos (approved):

Caleb-Hill (approved):

peterschwarz (approved):

peterschwarz:
Tested

PR #1683, opened 2021-12-02
Clippy fixes

Caleb-Hill:


agunde406 (approved):

libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -90,13 +90,12 @@ impl RestResourceProvider for AdminService {
...
 #[derive(Clone)]
 pub struct CircuitResourceProvider {
-    node_id: String,
     store: Box<dyn AdminServiceStore>,
 }
 
 impl CircuitResourceProvider {
-    pub fn new(node_id: String, store: Box<dyn AdminServiceStore>) -> Self {
-        Self { node_id, store }
+    pub fn new(store: Box<dyn AdminServiceStore>) -> Self {
```
vaporos, line 97: Note this change is pub and should not be backported.

libsplinter/src/network/auth/mod.rs
```diff
@@ -66,6 +66,7 @@ pub(crate) struct ManagedAuthorizationState {
     accepting_state: AuthorizationAcceptingState,
 
     // Tracks whether the local node has completed authorization with the remote node
+    #[cfg(any(feature = "trust-authorization", feature = "challenge-authorization"))]
```
vaporos, line 69: Typo in commit - "feaures"

vaporos (changes requested):
The commit which breaks compatibility with the current public API should explicitly state it breaks compatibility in the commit message.

libsplinter/src/admin/rest_api/mod.rs
Caleb-Hill: Updated the commit message to say that

libsplinter/src/network/auth/mod.rs
Caleb-Hill: Fixed

agunde406 (approved):

vaporos (approved):

PR #1686, opened 2021-12-06
0.4 Fix clippy lint for 1.57 Rust

agunde406:


libsplinter/src/admin/rest_api/mod.rs
```diff
@@ -72,6 +72,8 @@ impl RestResourceProvider for AdminService {
 ///
 /// * `rest-api-actix`
 #[derive(Clone)]
+// Allow dead code here as this is a part of the public API but it may never be used
+#[allow(dead_code)]
```
vaporos, line 76: See the PR for 0-6 branch, which removes node_id from the struct and adds and underscore the the node_id argument on new(). That might be a better approach since it doesn't introduce any allow statements.

cli/src/action/registry/mod.rs
```diff
@@ -65,7 +65,7 @@ impl Action for RegistryGenerateAction {
         let keys = args
             .values_of("key_files")
             .ok_or_else(|| CliError::ActionError("One or more key files must be specified".into()))?
-            .map(|key_file| read_private_key(key_file))
+            .map(read_private_key)
```
vaporos, line 68: Typo in commit message - redudant_closure

libsplinter/src/admin/rest_api/mod.rs
agunde406: 0-4 branch does not have the "admin" feature which blocks pulling in all of the admin service. Because of this both the node_id and store field will come up as unused in different configurations. For example when the health service is built.

cli/src/action/registry/mod.rs
agunde406: fixed

vaporos (approved):

rbuysse (approved):

PR #1689, opened 2021-12-08
Fix 401 in tests

isabeltomb:
This PR fixes a bug that was found in the two REST API endpoint tests. The logs for the tests showed "The server didn't upgrade: 401 Unauthorized" when the test attempted to send a request to the `/ws/admin/register/{}` endpoint. These two tests use an actual authorization handler instead of the mock handler so when any request is sent to the REST API it must have a valid auth header. The admin service event client for the node was being created with "foo" as the auth string, causing any requests it made to fail.

- Update the testing framework so that you can set the auth header that will be used by the admin service even client when submitting requests. If no auth header is set "foo" will still be used by default.
- Update the REST API endpoint tests to pass in an auth string created with the admin signer's JWT when creating the test network.


peterschwarz (dismissed):

splinterd/src/node/builder/mod.rs
```diff
@@ -237,6 +239,11 @@ impl NodeBuilder {
         }
     }
 
+    pub fn with_auth_string(mut self, auth: Option<String>) -> Self {
```
peterschwarz, line 242: Perhaps, though this could be named "with_client_auth"

isabeltomb: Updated

agunde406 (approved):

peterschwarz (approved):

PR #1691, opened 2021-12-09
Log default adjustment

Caleb-Hill:
Redo the -v verbosity flag logic and add `scabbard` and `cylinder` default loggers.

### DIFF
--- a/splinterd/src/config/default.rs
+++ b/splinterd/src/config/default.rs
@@ -127,14 +127,14 @@ impl PartialConfigBuilder for DefaultPartialConfigBuilder {
                 "splinter".to_string(),
                 UnnamedLoggerConfig {
                     appenders: None,
-                    level: Some(log::Level::Info),
+                    level: Some(log::Level::Trace),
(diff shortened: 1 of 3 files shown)


vaporos (approved):

splinterd/src/config/default.rs
```diff
@@ -137,6 +137,27 @@ impl PartialConfigBuilder for DefaultPartialConfigBuilder {
                     level: Some(log::Level::Trace),
                 },
             ),
+            (
+                "sawtooth".to_string(),
```
agunde406, line 141: I think scabbard needs to be included in this commit

agunde406 (changes requested):

splinterd/src/config/default.rs
Caleb-Hill: Added a scabbard logger.

vaporos (approved):

agunde406 (approved):

PR #1704, opened 2021-12-13
Remove experimental use of Actix Web 3

vaporos:
This removes actix web 3 use (which is experimental and unfinished) with
the goal of speeding up compile times, particularly for backports in CI.

When testing on M1X, the overall time difference is not significant:

```
Before:

real	8m51.654s
user	28m39.814s
sys	2m3.126s

After:

real	8m29.726s
user	25m42.805s
sys	1m52.968s
```

However, this may still save time in CI runs.


vaporos:
Could not demonstrate that this will significantly speed up builds in practice.

PR #1718, opened 2021-12-15
Enclose http statuses in quotes in openapi.yaml

Caleb-Hill:
The openapi standard states that all http statuses should be enclosed in
quotes for cross compatibility between yaml and json representations.
This commit encloses all status codes in the file.

For reference:
https://spec.openapis.org/oas/latest.html#responses-object

Section 4.8.16.2 contains the relevant bits of the standard.

### DIFF
--- a/splinterd/api/static/openapi.yaml
+++ b/splinterd/api/static/openapi.yaml
@@ -35,15 +35,15 @@ paths:
         - $ref: "#/components/parameters/auth"
         - $ref: "#/components/parameters/protocol_version"
       responses:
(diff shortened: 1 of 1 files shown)


vaporos (commented):
In the example - https://spec.openapis.org/oas/latest.html#responses-object - it uses single quotes for yaml quoting.

Caleb-Hill:
> 

Fixed

vaporos (approved):

agunde406 (approved):

PR #1721, opened 2021-12-16
Update oauth2 dependency

isabeltomb:
Update the oauth2 dependency to 4 from 3. Update
`set_redirect_url` to `set_redirect_uri` based on changes in oauth2.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -54,7 +54,7 @@ log = "0.4"
 metrics = {version = "0.17", features = ["std"], optional = true}
 mio = { version = "0.6", default-features = false }
 mio-extras = "2"
-oauth2 = { version = "3.0", optional = true }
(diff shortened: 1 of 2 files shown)


libsplinter/Cargo.toml
```diff
@@ -54,7 +54,7 @@ log = "0.4"
 metrics = {version = "0.17", features = ["std"], optional = true}
 mio = { version = "0.6", default-features = false }
 mio-extras = "2"
-oauth2 = { version = "3.0", optional = true }
+oauth2 = { version = "4.1.0", optional = true }
```
vaporos, line 57: Set this to "4" unless the intent is to explicitly say that it will only work with 4.1 or later; in that case, set it to "4.1".

vaporos (changes requested):

vaporos (approved):

agunde406 (approved):

PR #1729, opened 2021-12-16
Use a single SplinterProtocolVersion value across libsplinter

vaporos:
This replaces crate::rest_api::*_PROTOCOL_VERSION with a single
SPLINTER_PROTOCOL_VERSION. This is consistent with the original
intended design and the future direction of the entire Splinter REST API
being versioned as a whole (as opposed to versioned per-component).

### DIFF
--- a/libsplinter/src/admin/client/event/ws/actix_web_client.rs
+++ b/libsplinter/src/admin/client/event/ws/actix_web_client.rs
@@ -29,7 +29,7 @@ use crate::events::{
     Igniter, ParseBytes, ParseError, Reactor, WebSocketClient, WebSocketError, WsResponse,
 };
 use crate::hex;
-use crate::rest_api::ADMIN_PROTOCOL_VERSION;
+use crate::rest_api::SPLINTER_PROTOCOL_VERSION;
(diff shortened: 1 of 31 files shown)


peterschwarz (approved):

libsplinter/src/protocol/mod.rs
```diff
@@ -21,7 +21,7 @@ pub mod service;
 
 // Peer authorization protocol versions
 #[cfg(any(feature = "trust-authorization", feature = "challenge-authorization"))]
-pub const PEER_AUTHORIZATION_PROTOCOL_VERSION: u32 = 1;
+pub const PEER_SPLINTER_PROTOCOL_VERSION: u32 = 1;
```
agunde406, line 24: This changes should be split into a different commit.

agunde406 (changes requested):

libsplinter/src/protocol/mod.rs
vaporos: This was unintentional. Removed from the commit.

agunde406 (approved):

isabeltomb (approved):

PR #1747, opened 2021-12-22
Move scabbard's service implementation to a v2 directory

vaporos:


Caleb-Hill (approved):

peterschwarz (approved):

agunde406 (approved):

vaporos:
The eventual organization of the service files has changed, so this specific PR no longer makes sense as-is. A similar PR in the future may make sense, taking into account the current structure.

PR #1749, opened 2021-12-22
Only label PRs when they're opened

rbuysse:
When new commits are pushed to a pull request, the synchronize action
is triggered which re-runs all GHA workflows. By limiting the activity
type to "opened", this action will not re-apply labels to PRs if they're
updated.


rbuysse:
Supporting GHA documentation: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

agunde406 (dismissed):

PR #1763, opened 2022-01-06
Add sqlite migration for renamed role tables

Caleb-Hill:
sqlite does not always update the parent table references on table
renames. When the `roles` table was updated to be named `rbac_roles` the
foreign key constraint on the `role_permission' table was not updated
making it impossible to insert any new role:permission relations because
the foreign key constraint would fail.

This commit adds a database migration to update the foreign key
constraint on the `rbac_role_permissions` table to reference the
`rbac_roles` table instead of the `roles` table.

It also adds an undo sql migration that breaks it again.


peterschwarz (approved):

vaporos (approved):

Caleb-Hill:
backport for 0-6 here https://github.com/Cargill/splinter/pull/1764

PR #1765, opened 2022-01-07
Modify splinter user list command

isabeltomb:
Modify the splinter user list command to be able to list just biome or
just oauth users. Previously, in the case that either the /biome/users or
/oauth/users endpoint wasn't available this CLI command would fail and
return an error saying the endpoint doesn't exist. This was an issue
when attempting to list users while only biome auth was configured
because the /oauth/users endpoint doesn't exist, causing the command
to return an error.

An error is still logged in case one or both of the auth types are
configured and the appropriate endpoint fails to return the list of
users.


agunde406 (approved):

cli/src/action/user/mod.rs
```diff
@@ -52,22 +52,46 @@ fn display_splinter_users(
...
-    let biome_oauth_users = client
-        .list_oauth_users()?
-        .data
-        .into_iter()
-        .map(ClientSplinterUser::from);
+    let biome_users = match client.list_biome_users() {
+        Ok(users) => Some(users.into_iter().map(ClientSplinterUser::from)),
+        Err(e) => {
+            info!(
+                "Ignore if biome credentials are not enabled: Biome users not listed: {}",
```
vaporos, line 59: This error message will be confusing if the CLI is just pointed at the wrong HTTP port, as the user is unlikely to understand what we mean by "enabled" here. Suggest simplifying this to Just The Facts: "Unable to retrieve Biome users: {}".

If both of these fail, should that still be an error?

isabeltomb: Updated both log messages. If both fail the match statement on line 76
`let users = match (biome_users, biome_oauth_users) {...}`  will catch it and return an error.

vaporos (approved):

agunde406 (approved):

PR #1777, opened 2022-01-10
Remove check excluding full filepath templates

Caleb-Hill:
Checking through the configured template directories for the passed in
filename is good when the user is using those directories. But when the
user passes in the full filepath to the template file it will
incorrectly reject the template path.

This commit changes to the code to always try to load the filepath as
given and then maps the errors to the previously returned errors.

### DIFF
--- a/cli/src/template.rs
+++ b/cli/src/template.rs
@@ -15,7 +15,7 @@
 //! Data structure and implementation of the circuit template representation for the CLI.
 
 use std::collections::HashMap;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use splinter::circuit::template::{
     CircuitCreateTemplate, CircuitTemplateError, CircuitTemplateManager, RuleArgument,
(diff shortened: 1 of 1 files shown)


cli/src/template.rs
```diff
@@ -91,13 +91,20 @@ impl CircuitTemplate {
...
-        if !possible_values.iter().any(|(stem, _)| stem == name) {
-            return Err(CliError::ActionError(format!(
-                "Template with name {} was not found. Available templates: {:?}",
-                name, possible_values
-            )));
-        }
-        let template = manager.load(name)?;
+        let template = manager.load(name).map_err(|err| {
+            let named_path = Path::new(name);
+            if named_path.has_root() {
```
agunde406, line 96: Template load actually supports both passing in with the root or without https://github.com/Cargill/splinter/blob/main/libsplinter/src/circuit/template/mod.rs#L183 , so instead this should check the following https://github.com/Cargill/splinter/blob/main/libsplinter/src/circuit/template/mod.rs#L180

agunde406 (changes requested):

cli/src/template.rs
Caleb-Hill: Fixed

agunde406 (approved):

isabeltomb (approved):

PR #1779, opened 2022-01-10
Backport: Add quotations around user input in error

Caleb-Hill:
The `first_service` field needs to be a 4 char base 62 string. If it
isn't an error is returned that says:

"Field first_service is invalid (abcde)"

This commit adds "" marks around the user sourced `first_service`
string to better show what constitutes the wrong string.

Like so:

"Field first_service is invalid ("abcde")"


Caleb-Hill:
https://github.com/Cargill/splinter/pull/1778

agunde406 (approved):

isabeltomb (approved):

PR #1815, opened 2022-02-10
Return batch statuses even when some timeout

leebradley:
This change modifies the Splinter /batch_statuses endpoint so that it
will return batch statuses, even if some of them time out. Batch
statuses that time out are returned as "Unknown".

Previously, if a single batch status were to timeout, the endpoint would
return none at all. The endpoint result would be a non-parsable string
error containing the problem batch ids.

This change should give users of this endpoint maximal information.
Users can use the successful ids, and they can determine programatically
which ones timed out.

Also updated several tests to validate the new expected responses.

### DIFF
--- a/services/scabbard/libscabbard/src/client/reqwest/mod.rs
+++ b/services/scabbard/libscabbard/src/client/reqwest/mod.rs
@@ -319,13 +319,13 @@ fn wait_for_batches(
                 )));
             }
         } else {
-            let any_invalid_batches = batch_infos
+            let any_invalid_or_unknown = batch_infos
                 .iter()
-                .any(|info| matches!(info.status, BatchStatus::Invalid(_)));
+                .any(|info| matches!(info.status, BatchStatus::Invalid(_) | BatchStatus::Unknown));
 
-            if any_invalid_batches {
+            if any_invalid_or_unknown {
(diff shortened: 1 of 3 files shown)


vaporos (approved):

services/scabbard/libscabbard/src/client/reqwest/mod.rs
```diff
@@ -321,11 +321,11 @@ fn wait_for_batches(
         } else {
             let any_invalid_batches = batch_infos
                 .iter()
-                .any(|info| matches!(info.status, BatchStatus::Invalid(_)));
+                .any(|info| matches!(info.status, BatchStatus::Invalid(_) | BatchStatus::Unknown));
```
peterschwarz, line 324: The variable `any_invalid_batches` should match this new state - `any_invalid_or_unknown` as a suggestion.

peterschwarz (changes requested):

services/scabbard/libscabbard/src/client/reqwest/mod.rs
leebradley: Good point! Updated

peterschwarz (dismissed):

peterschwarz (approved):

vaporos (approved):

PR #1821, opened 2022-02-16
Backport 0-6: Add error data description

leebradley:
The error_data field in the openapi spec currently defines error_data as
an array of integers, which while technically accurate is misleading.
This change clarifies that the error_data field is actually an arbitrary
array of bytes, and specifically calls out the allowed ranges for these
values.


leebradley:
This is a backport of #1807

isabeltomb (approved):

PR #1827, opened 2022-02-24
Backport 0-6: Update `splinter circuit show` CLI command

isabeltomb:
Backport of #1826

### DIFF
--- a/cli/src/action/circuit/api.rs
+++ b/cli/src/action/circuit/api.rs
@@ -275,6 +275,12 @@ impl fmt::Display for CircuitSlice {
             if let Some(public_key) = &member.public_key {
                 display_string += &format!("        Public Key: {}\n", public_key);
             }
+
+            display_string += "        Endpoints:\n";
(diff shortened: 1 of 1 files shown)


agunde406 (approved):

cli/src/action/circuit/api.rs
```diff
@@ -275,6 +275,12 @@ impl fmt::Display for CircuitSlice {
             if let Some(public_key) = &member.public_key {
                 display_string += &format!("        Public Key: {}\n", public_key);
             }
+
+            display_string += &"        Endpoints:\n".to_string();
```
peterschwarz, line 279: This will result in lint when merged. Change this to 

```rust
display_string += "        Endpoints:\n";
```

peterschwarz (changes requested):
This will have lint

cli/src/action/circuit/api.rs
isabeltomb: Updated

agunde406 (approved):

PR #1832, opened 2022-02-25
Replace tempdir library with tempfile

leebradley:
Replace tempdir library with tempfile:3.3 because tempdir is no longer
maintained. See advisory: https://rustsec.org/advisories/RUSTSEC-2018-0017

Remove tempdir from libscabbard, as it is unused.

Move the library to dev-dependencies in splinterd, as it is only ever used
for testing purposes.

### DIFF
--- a/cli/src/action/database/upgrade/node_id.rs
+++ b/cli/src/action/database/upgrade/node_id.rs
@@ -96,7 +96,7 @@ mod tests {
 
     use splinter::error::InternalError;
     use splinter::node_id::store::{error::NodeIdStoreError, NodeIdStore};
-    use tempdir::TempDir;
+    use tempfile::Builder;
 
     use super::*;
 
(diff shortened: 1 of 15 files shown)


peterschwarz (approved):

cli/src/action/database/upgrade/node_id.rs
```diff
@@ -201,7 +201,7 @@ mod tests {
     #[test]
     // Test reading from the node_id file works as intended.
     fn test_migrate_to_db_with_file() {
-        let directory = TempDir::new("test").expect("could not create temp directory");
+        let directory = TempDir::new().expect("could not create temp directory");
```
vaporos, line 204: Keeping the prefix here could be done with this library:

```
let directory = Builder::new().prefix("test").tempdir().expect(...);
```

vaporos (changes requested):

peterschwarz (approved):

cli/src/action/database/upgrade/node_id.rs
leebradley: Updated!

vaporos (approved):

PR #1853, opened 2022-03-15
Add state pruning after each commit

peterschwarz:
This change prunes the previous state root hash from the tree. As scabbard consensus has finality, this state snapshot is no longer needed.


agunde406 (approved):

peterschwarz:
This PR should be closed until such time that the issues with state pruning and SQL-backed merkle state are resolved in transact.

PR #1855, opened 2022-03-15
Add retry to sending messages in PeerInterconnect

agunde406:
This commit adds retry to sending messages to a connection by
updating the pending incoming messages loop to handle booth
incoming and outgoing messages.

Now if the PeerInterconnect receives a message that should be
sent to a Peer, but the message fails to send because the
connection has disconnected, the message will be sent to the
pending thread, where it will be retried up to 3 times, with
10 seconds in between each try.

The connection could go away because of race condition where,
duplicate connections are rectified but one side removes
the old connection before the other finishes authorization.
This change would allow for the message to be sent over the
new connection successfully.

Also fixes periodic test failures in integration test
test_2_party_circuit_duplicate_connection

### DIFF
--- a/libsplinter/src/peer/error.rs
+++ b/libsplinter/src/peer/error.rs
@@ -185,25 +185,6 @@ impl fmt::Display for PeerUpdateError {
     }
 }
 
-/// Errors that could be raised by `PeerInterconnect`
-#[derive(Debug, PartialEq)]
-pub enum PeerInterconnectError {
-    /// `PeerInterconnect` start up failed
-    StartUpError(String),
-}
-
-impl error::Error for PeerInterconnectError {}
-
-impl fmt::Display for PeerInterconnectError {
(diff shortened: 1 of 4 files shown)


libsplinter/src/peer/interconnect.rs
```diff
@@ -357,6 +366,7 @@ where
             dispatched_sender,
             recv_join_handle,
             send_join_handle,
+
```
peterschwarz, line 369: Spurious whitespace

peterschwarz (changes requested):
It would be nice to break out the new code into a new file - it's a considerable addition.

libsplinter/src/peer/interconnect.rs
agunde406: Fixed, and split out the pending loop into its own file.

peterschwarz (approved):

vaporos (approved):

PR #1858, opened 2022-03-16
Add Timer for new service API 

agunde406:
The Timer is in charge of checking the configured
`TimerFilters` for pending work.

On start up the Timer takes a list of `TimerFilters`
and their associated `TimerHandlerFactory`. The factories
must return a handler that can handle messages of type `Vec<u8>`.

On wake up, the Timer will check each `TimerFilter` for pending work.
For each `FullyQualifiedServiceId` returned from the filter,
the associated `TimerHandlerFactory` will be used to get a
new `TimerHandler` that will run in a threadpool.

### DIFF
--- a/libsplinter/src/runtime/service/timer/thread.rs
+++ b/libsplinter/src/runtime/service/timer/thread.rs
@@ -1,0 +1,261 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you mcay not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The internal thread used by the Timer
+
+use std::sync::mpsc::{Receiver, Sender};
+use std::thread;
+
+use crate::error::InternalError;
+use crate::service::{FullyQualifiedServiceId, MessageSenderFactory, ServiceType};
+use crate::threading::{
+    lifecycle::ShutdownHandle,
+    pool::{JobExecutor, ThreadPool, ThreadPoolBuilder},
+};
+
+use super::message::TimerMessage;
+use super::FilterCollection;
+
+const DEFAULT_POOL_SIZE: usize = 8;
+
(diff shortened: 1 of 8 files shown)


libsplinter/src/runtime/service/timer/thread.rs
```diff
@@ -0,0 +1,243 @@
...
+use super::FilterCollection;
+
+const DEFAULT_POOL_SIZE: usize = 8;
+
+pub struct TimerThread {
+    join_handle: thread::JoinHandle<()>,
+}
+
+impl TimerThread {
+    pub fn start(
```
vaporos, line 34: The inner thread doesn't follow the two-phase threading::ShutdownHandle approach when handling the thread_pool.

libsplinter/src/service/timer_filter.rs
```diff
@@ -14,9 +14,9 @@
 
 use crate::error::InternalError;
 
-use super::FullyQualifiedServiceId;
+use super::{FullyQualifiedServiceId, Typed};
 
-pub trait TimerFilter {
+pub trait TimerFilter: Typed {
```
vaporos, line 19: We should discuss whether TimerFilter should be support more than a single service type. For example, if we have "scabbard::v3" and "scabbard::v4", should we require different TimerFilter instances. This is also going to be the same for MessageHandler, etc.

vaporos: Also Typed should return a struct like ServiceType - should discuss whether that change should go in before or after this.

libsplinter/src/runtime/service/timer/thread.rs
agunde406: Updated

peterschwarz (dismissed):

peterschwarz (dismissed):

libsplinter/src/runtime/service/timer/thread.rs
```diff
@@ -0,0 +1,261 @@
...
+        } else {
+            warn!(
+                "Received a wake up for service id that doesn't have \
+                pending work: {}",
+                id
+            )
+        }
+    } else {
+        for service_id in service_ids.into_iter() {
+            let new_handle_factory = handler_factory.clone_box();
```
Caleb-Hill, line 230: Not a blocker but we should choose one of clone_box or clone_boxed and stick with it

Caleb-Hill (approved):

Caleb-Hill (approved):

libsplinter/src/runtime/service/timer/thread.rs
agunde406: Story added for the cleanup

peterschwarz (approved):

PR #1859, opened 2022-03-16
Implement LifecycleExecutor

agunde406:
The LifecycleExecutor will check the database for pending
services that need to be prepared, finalized, retired or purged.

It will check the database on a configured interval or can be
woken up by an alarm. When it finds a pending service, it will
get the commands from the Lifecycle implementation for the service
type and add its own command to update the Lifecycle store. It
then passes all commands to an executor.

The commands should all be very quick so the LifecycleExecutor
only has one thread.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -130,6 +130,7 @@ experimental = [
     "service-id",
     "service-lifecycle",
     "service-lifecycle-executor",
+    "service-lifecycle-store",
     "service-message-converter",
     "service-message-handler",
     "service-message-handler-factory",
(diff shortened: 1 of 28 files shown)


libsplinter/Cargo.toml
```diff
@@ -129,6 +129,8 @@ experimental = [
     "service-arguments-converter",
     "service-id",
     "service-lifecycle",
+    "service-lifecycle-executor",
```
peterschwarz, line 132: This should probably be split into a separate commit

libsplinter/src/runtime/service/lifecycle_executor/store/diesel/mod.rs
```diff
@@ -175,6 +179,90 @@ impl DieselLifecycleStore<diesel::sqlite::SqliteConnection> {
...
+    C: diesel::Connection<TransactionManager = AnsiTransactionManager> + 'static,
+    C::Backend: diesel::backend::UsesAnsiSavepointSyntax,
+{
+    pub fn new(connection: &'a C) -> Self {
+        DieselConnectionLifecycleStore { connection }
+    }
+}
+
+#[cfg(feature = "postgres")]
+impl<'a> LifecycleStore for DieselConnectionLifecycleStore<'a, diesel::pg::PgConnection> {
```
peterschwarz, line 201: This seems like it should be in its own commit as well (it's not mentioned in the commit message)

Overall the commit message doesn't cover the purpose of these changes.

peterschwarz (changes requested):

libsplinter/Cargo.toml
agunde406: split

peterschwarz (approved):

isabeltomb (approved):

PR #1861, opened 2022-03-18
Service type struct

peterschwarz:


agunde406 (approved):
`Typed` should be removed in a follow up PR

vaporos (approved):

PR #1862, opened 2022-03-18
Service message dispatch

peterschwarz:


agunde406 (changes requested):

agunde406 (approved):

libsplinter/src/service/dispatch/task_job_executor.rs
```diff
@@ -0,0 +1,138 @@
...
+    ) -> Result<(), InternalError> {
+        let factory = message_handler_factory.clone_boxed();
+        let sender_factory = sender_factory.clone_boxed();
+
+        self.job_executor.execute(move || {
+            let mut handler = factory.new_handler();
+            let sender = match sender_factory.new_message_sender(&to_service) {
+                Ok(sender) => sender,
+                Err(err) => {
+                    error!("Unable to create new message sender: {}", err);
```
vaporos, line 126: We lose the information about the message being sent; the message itself is just bytes but we could log from_service/to_service. Same with the error!() below.

vaporos:
The pieces here which are not part of the API of writing a service should go into runtime::service::.

libsplinter/src/service/dispatch/task_job_executor.rs
peterschwarz: Updated

peterschwarz:
> The pieces here which are not part of the API of writing a service should go into runtime::service::.

Moved all of the new code to `runtime::service::dispatch`

agunde406 (dismissed):

libsplinter/src/circuit/handlers/direct_message.rs
```diff
@@ -20,12 +20,24 @@ use crate::protos::circuit::{
     CircuitDirectMessage, CircuitError, CircuitError_Error, CircuitMessageType,
 };
 
+#[cfg(feature = "service-message-handler-dispatch")]
+use crate::error::InternalError;
+#[cfg(feature = "service-message-handler-dispatch")]
+use crate::{
+    runtime::service::ServiceDispatcher,
+    service::{
+        CircuitId as Service2CircuitId, FullyQualifiedServiceId, ServiceId as Service2ServiceId,
```
vaporos, line 29: Remove the "2" here.

peterschwarz: Replaced `Service2-` with `Routing-` in both cases.

agunde406 (dismissed):

vaporos (approved):

libsplinter/src/circuit/handlers/direct_message.rs
peterschwarz: Updated to remove `Service2-` and add `Routing-` to the `circuit::routing::ServiceId` usage (via its own commit).

agunde406 (approved):

PR #1864, opened 2022-03-21
Rename Service::peer_id to ::local_peer_id

peterschwarz:
This change renames circuit::routing::Service::peer_id to ::local_peer_id as it is a more accurate name for its purpose (matching its rustdoc comment, for example).


vaporos (changes requested):
This changes a public API - circuit::routing::Service::peer_id() etc., so minimally a those functions should be retained as wrappers to the new names.

peterschwarz:
I've restored the original functions and marked them as deprecated.

vaporos (dismissed):

agunde406 (approved):

vaporos (approved):

PR #1867, opened 2022-03-23
Rest api common error and paging modules

Caleb-Hill:
This PR adds some of the base level utility modules to common as well as dependencies on libsplinter and serde


rest_api/common/src/paging.rs
```diff
@@ -0,0 +1,168 @@
+// Copyright 2018-2022 Cargill Incorporated
```
vaporos, line 1: This should be versioned as v1 - see Grid sdk's rest_api::resources::paging::v1 module for layout.

rest_api/common/src/paging.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    #[test]
+    fn test_50limit_150offset_paging_response() {
+        // Create paging response from limit of 50, offset of 150, and total of 1000
+        let test_paging_response = get_response_paging_info(Some(50), Some(150), TEST_LINK, 1000);
+        let generated_paging_response = create_test_paging_response(150, 50, 200, 100, 950);
+        assert_eq!(test_paging_response, generated_paging_response);
+    }
+
+    fn create_test_paging_response(
```
vaporos, line 142: Feels like this could be PagingBuilder code and Paging fields non-pub so that the integrity of the struct is always valid.

rest_api/common/src/paging.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub offset: usize,
+    pub limit: usize,
+    pub total: usize,
+    pub first: String,
+    pub prev: String,
+    pub next: String,
+    pub last: String,
+}
+
+pub fn get_response_paging_info(
```
vaporos, line 32: Feels like this should be PagingBuilder.

rest_api/common/src/error.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
```
vaporos, line 2: Typo in this commit message - "ment"

Caleb-Hill: fixed

rest_api/common/src/paging.rs
Caleb-Hill: Added v1 module to paging

rest_api/common/src/paging.rs
Caleb-Hill: Added builder

rest_api/common/src/paging.rs
Caleb-Hill: Fixed

rest_api/common/src/paging/v1/builder.rs
```diff
@@ -0,0 +1,138 @@
...
+            limit: self.limit,
+        }
+    }
+}
+
+impl PagingBuilder<Yes, Yes> {
+    pub fn build(self) -> Paging {
+        let limit = self.limit.unwrap_or(DEFAULT_LIMIT);
+        let offset = self.offset.unwrap_or(DEFAULT_OFFSET);
+        let link = self.link.unwrap();
```
vaporos, line 92: Remove this unwrap (not allowed in splinter libraries).

vaporos (changes requested):

rest_api/common/src/paging/v1/builder.rs
```diff
@@ -0,0 +1,138 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::marker::PhantomData;
+
+use super::{Paging, DEFAULT_LIMIT, DEFAULT_OFFSET};
+
+pub struct PagingBuilder<HasLink, HasQueryCount> {
```
vaporos, line 19: If we are going to use generics to enforce correctness, then the types should not be Yes/No but rather be containers with the values themselves, to avoid needing to unwrap during build().

rest_api/common/src/paging/v1/builder.rs
Caleb-Hill: fixed

peterschwarz (approved):

vaporos (approved):

PR #1868, opened 2022-03-23
Add AdminService integration with the new Lifecycle Executor API

agunde406:
The admin service will now take a list of lifecycle
dispatches and call the service update method on each
one. If the current implementation does not support
the service type, it should return Ok.

The ServiceOrchestrator will implement the LifecycleDispatch trait
directly. A new SyncLifecycleInterface is added to support the new
async service lifecycle processe for the new Service API.

With this update a lock around the orchestrator is is removed in favor
of the lock around AdminShared.

### DIFF
--- a/libsplinter/src/admin/lifecycle/orchestrator.rs
+++ b/libsplinter/src/admin/lifecycle/orchestrator.rs
@@ -1,0 +1,168 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+
+use crate::error::InternalError;
+use crate::orchestrator::{ServiceDefinition, ServiceOrchestrator};
+
+use super::LifecycleDispatch;
+
+impl LifecycleDispatch for ServiceOrchestrator {
+    fn add_service(
+        &self,
+        circuit_id: &str,
+        service_id: &str,
+        service_type: &str,
+        args: Vec<(String, String)>,
+    ) -> Result<(), InternalError> {
+        if !self
+            .supported_service_types()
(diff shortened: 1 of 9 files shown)


Caleb-Hill (approved):

isabeltomb (approved):

libsplinter/src/admin/lifecycle/orchestrator.rs
```diff
@@ -0,0 +1,168 @@
...
+            .contains(&service_type.to_string())
+        {
+            trace!(
+                "Ignoring call to add service, service type not supported: {}",
+                service_type
+            );
+            return Ok(());
+        }
+
+        debug!(
```
peterschwarz, line 149: Does this double any logging in the existing `add_stopped_service` method?

libsplinter/src/admin/lifecycle/orchestrator.rs
```diff
@@ -0,0 +1,168 @@
...
+            service_id: service_id.to_string(),
+            service_type: service_type.to_string(),
+        };
+
+        let mut arg_map = HashMap::new();
+        for (key, value) in args {
+            arg_map.insert(key, value);
+        }
+
+        self.add_stopped_service(service_defination, arg_map)
```
peterschwarz, line 165: These shadow the same name - might be better to do
```rust
ServiceOrchestrator::add_stopped_service(&self, service_definition, arg_map)
```
instead, to make it obvious what is being called (and avoid any possible future accidental infinite recursions if the args change).

libsplinter/src/admin/lifecycle/orchestrator.rs
agunde406: Nope, there is no logging that I was able to find.

libsplinter/src/admin/lifecycle/orchestrator.rs
agunde406: Updated all methods to use this pattern

libsplinter/src/admin/lifecycle/sync.rs
```diff
@@ -0,0 +1,614 @@
...
+        let instant = time::Instant::now();
+        while service.status() == &LifecycleStatus::New {
+            if instant.elapsed() > self.time_to_wait {
+                return Err(InternalError::with_message(format!(
+                    "Service {} was not finalized in time",
+                    service_id
+                )));
+            }
+
+            thread::sleep(std::time::Duration::from_secs(TIME_BETWEEN_DATABASE_CHECK));
```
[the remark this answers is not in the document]
agunde406: updated

peterschwarz (approved):

isabeltomb (approved):

PR #1874, opened 2022-03-30
Update splinter-dev publishing to use OIDC for AWS creds

rbuysse:
The developers of the `configure-aws-credentials` action now recommend
using OIDC instead of an IAM user with credentials stored in GitHub secrets.

https://github.com/aws-actions/configure-aws-credentials


rbuysse:
This branch was pushed to upstream to trigger an example run with new credential changes: https://github.com/Cargill/splinter/actions/runs/2066937449. The AWS role is scoped to only allow assuming roles from `Cargill/*` repos.

Example failure here: https://github.com/rbuysse/splinter/runs/5759764056#step:2:5

peterschwarz (approved):

PR #1876, opened 2022-04-01
Replace the use of whitelist with allow_list in CORS

agunde406:

### DIFF
--- a/splinterd/man/splinterd.1.md
+++ b/splinterd/man/splinterd.1.md
@@ -269,14 +269,14 @@ OPTIONS
 : Specifies the path and file name for the REST API key.
   (Default: `/etc/splinter/certs/rest_api.key`.)
 
-`--whitelist WHITELIST` `[,...]`
+`--allow-list ALLOW_LIST` `[,...]`
 : Lists one or more trusted domains for cross-origin resource sharing (CORS).
   This option allows the specified domains to access restricted web resources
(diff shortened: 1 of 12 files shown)


splinterd/src/main.rs
```diff
@@ -386,13 +386,14 @@ fn main() {
 
     #[cfg(feature = "rest-api-cors")]
     let app = app.arg(
-        Arg::with_name("whitelist")
```
peterschwarz, line 389: This should probably be an alias for allow_list and hidden in the doc output, for backwards compatibility

agunde406: An alias is already added on line 392

peterschwarz (approved):

splinterd/man/splinterd.1.md
```diff
@@ -269,14 +269,14 @@ OPTIONS
 : Specifies the path and file name for the REST API key.
   (Default: `/etc/splinter/certs/rest_api.key`.)
 
-`--whitelist WHITELIST` `[,...]`
+`--allow_list ALLOW_LIST` `[,...]`
```
vaporos, line 272: Should be --allow-list, not --allow_list

agunde406: fixed

peterschwarz (approved):

vaporos (approved):

PR #1886, opened 2022-04-07
Permission map updates

Caleb-Hill:
This PR adds a Method enum to act as a lowest common denominator for actix-web-1::Method and any future framework Method types. In future PRs the actix-web-4::Method will use this struct.


vaporos (approved):

libsplinter/src/rest_api/auth/authorization/permission_map.rs
```diff
@@ -56,6 +62,39 @@ impl PermissionMap {
...
+    Patch,
+    Delete,
+    Head,
+    Options,
+    Connect,
+    Trace,
+    Extension(String),
+}
+
+impl From<&Actix1Method> for Method {
```
peterschwarz, line 79: Will this move to the actix-1 crate (in a future PR)?

peterschwarz (approved):

libsplinter/src/rest_api/auth/authorization/permission_map.rs
Caleb-Hill: Yeah, it probably won't look exactly like this because of the crate boundaries but the essence yes

PR #1891, opened 2022-04-08
Add scabbard_store module for scabbard v3

isabeltomb:
- Move the `CommitHashStore` and related modules to a `commit_hash` submodule in the store module
- Add a scabbard store module with the structs that will be used in the scabbard store. Most structs and enums are guarded with `#[allow(dead_code)]` to stop clippy lint because they are not yet being used - this will be removed in the future
- Add the beginning of the `ScabbardStore` trait


services/scabbard/libscabbard/src/store/scabbard_store/commit.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CommitEntry {
+    service_id: FullyQualifiedServiceId,
+    epoch: i64,
+    value: String,
+    decision: Option<ConsensusDecision>,
+}
+
+#[allow(dead_code)]
```
agunde406, line 28: We will have to be careful to make sure none of these are left behind once the structs are used

agunde406 (approved):

peterschwarz (approved):

PR #1905, opened 2022-04-14
Common auth exports

Caleb-Hill:
Changes related to being able to implement an auth service in actix_web_4 without breaking all of libsplinter and splinterd.

### DIFF
--- a/rest_api/common/Cargo.toml
+++ b/rest_api/common/Cargo.toml
@@ -40,3 +40,5 @@ experimental = [
     "stable",
     # The following features are experimental:
 ]
+
+authorization = ["splinter/authorization"]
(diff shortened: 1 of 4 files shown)


rest_api/common/src/auth/mod.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2018-2022 Cargill Incorporated
```
vaporos, line 1: The commit message mentions re-exports but there are none within this commit.

rest_api/common/Cargo.toml
```diff
@@ -40,3 +40,5 @@ experimental = [
     "stable",
     # The following features are experimental:
 ]
+
+authorization = ["splinter/authorization"]
```
vaporos, line 44: This should be added to experimental.

PR #1906, opened 2022-04-15
Add `Action` and `Message` structs/enums to scabbard store

isabeltomb:
Add Action and Message sub-modules to the scabbard_store module which contain structs used for storing consensus related information in the scabbard store.

These structs and enums are currently behind `#[allow(dead_code)]` to avoid clippy lint, this will be removed in the future when they are put to use.

### DIFF
--- a/services/scabbard/libscabbard/src/store/scabbard_store/action.rs
+++ b/services/scabbard/libscabbard/src/store/scabbard_store/action.rs
@@ -1,0 +1,21 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::store::scabbard_store::two_phase::action::ConsensusAction;
+
+#[allow(dead_code)]
+#[derive(Debug)]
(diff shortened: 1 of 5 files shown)


agunde406 (approved):

services/scabbard/libscabbard/src/store/scabbard_store/action.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum ScabbardConsensusAction {
+    Scabbard2pcConsensusAction(ConsensusAction),
+}
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum ConsensusAction {
```
peterschwarz, line 29: We should bump these into two_phase (or some such) submodule.

isabeltomb: Updated

agunde406 (approved):

services/scabbard/libscabbard/src/store/scabbard_store/two_phase/action.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::time::SystemTime;
+
+use splinter::service::ServiceId;
+
+use super::message::Scabbard2pcMessage;
+use crate::store::scabbard_store::context::ScabbardContext;
+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub enum ScabbardConsensusAction {
```
peterschwarz, line 24: This enum, which will include more than just 2pc, should be at the top level, but the 2pc enums should be in a two_phase module.

vaporos (approved):

services/scabbard/libscabbard/src/store/scabbard_store/two_phase/action.rs
isabeltomb: fixed

peterschwarz (approved):

agunde406 (approved):

PR #1910, opened 2022-04-20
Update ProposalStore to follow clone_boxed pattern

Caleb-Hill:
Having a trait require `Clone` introduces a `Sized` trait requirement
which makes it difficult wrap the trait object in any sort of pointer
type. This commit updates the `ProposalStore` trait by removing the
`Clone` trait bound, including a `clone_boxed` method and updating all
current implementations and usages of the trait to use a boxed trait
instead of generics.


libsplinter/src/admin/rest_api/actix/proposals.rs
```diff
@@ -44,20 +44,20 @@ pub fn make_list_proposals_resource<PS: ProposalStore + 'static>(proposal_store:
     #[cfg(feature = "authorization")]
     {
         resource.add_method(Method::Get, CIRCUIT_READ_PERMISSION, move |r, _| {
-            list_proposals(r, web::Data::new(proposal_store.clone()))
+            list_proposals(r, web::Data::new(proposal_store.clone_boxed()))
```
agunde406, line 47: If you add the following to the trait file, you can still just call `clone()`

```
impl Clone for Box<dyn ProposalStore> {
    fn clone(&self) -> Box<dyn ProposalStore> {
        self.clone_boxed()
    }
}
```

agunde406: Since you added that you should put the clone_boxed back to just clone() as well

agunde406 (approved):

PR #1929, opened 2022-04-28
Update Dockerfile to copy more than Cargo.tomls

Caleb-Hill:
The Dockerfile-installed-focal, and tests/Dockerfile copy just the
Cargo.tomls for the rest_api crates. Which is fine as long as no other
crates depend on those rest_api crates. As soon as splinterd pulls in
the actix_web_1 or 4 crates cargo won't be able to find those local
dependencies.

This commit updates the dockerfiles to copy the entire actix_web_1,
actix_web_4, and common rest_api crates into the build directory.


agunde406 (dismissed):

agunde406:
No longer required.

PR #1947, opened 2022-05-04
Add ConsensusActionRunner for Scabbard v3

agunde406:


services/scabbard/libscabbard/src/service/v3/consensus/consensus_action_runner/notify_observer/command.rs
```diff
@@ -0,0 +1,201 @@
...
+        notification: ConsensusActionNotification,
+        service_id: &FullyQualifiedServiceId,
+        epoch: u64,
+    ) -> Result<Vec<Box<dyn StoreCommand<Context = C>>>, InternalError> {
+        let mut commands: Vec<Box<dyn StoreCommand<Context = C>>> = Vec::new();
+        match notification {
+            // Generates a new value to agree on and creates commands to add a commit entry to
+            // track the value and an new event to start agreement on that value
+            ConsensusActionNotification::RequestForStart() => {
+                let s: String = rand::thread_rng()
```
peterschwarz, line 83: Instead of adding a new dep for this temporary use - we could just use the string value of system time millis

peterschwarz (changes requested):

services/scabbard/libscabbard/src/service/v3/consensus/consensus_action_runner/notify_observer/command.rs
agunde406: good idea, ill update that

agunde406: updated

peterschwarz (approved):

isabeltomb (approved):

PR #1948, opened 2022-05-05
Combine scabbard store consensus participant and coordinator

isabeltomb:
This PR combines the participant and coordinator tables for consensus contexts and actions in the diesel backed `ScabbardStore` implementation.

### DIFF
--- a/services/scabbard/libscabbard/src/service/v3/timer_filter.rs
+++ b/services/scabbard/libscabbard/src/service/v3/timer_filter.rs
@@ -174,7 +174,10 @@ mod tests {
             .with_alarm(SystemTime::now())
             .with_coordinator(fqsi.clone().service_id())
             .with_epoch(1)
-            .with_participant_processes(vec![peer_service1.clone()])
+            .with_participants(vec![Participant {
+                process: peer_service1.clone(),
+                vote: None,
+            }])
             .with_state(Scabbard2pcState::WaitingForVoteRequest)
             .with_this_process(fqsi2.clone().service_id())
             .build()
@@ -219,7 +222,10 @@ mod tests {
             .with_alarm(updated_alarm)
             .with_coordinator(fqsi.clone().service_id())
             .with_epoch(1)
-            .with_participant_processes(vec![peer_service1.clone()])
+            .with_participants(vec![Participant {
+                process: peer_service1.clone(),
+                vote: None,
+            }])
             .with_state(Scabbard2pcState::WaitingForVoteRequest)
             .with_this_process(fqsi2.clone().service_id())
             .build()
--- a/services/scabbard/libscabbard/src/store/mod.rs
+++ b/services/scabbard/libscabbard/src/store/mod.rs
@@ -19,7 +19,7 @@ mod command;
 mod commit_hash;
 #[cfg(any(feature = "postgres", feature = "sqlite"))]
 pub(crate) mod pool;
-#[cfg(feature = "scabbardv3")]
+#[cfg(feature = "scabbardv3-store")]
 mod scabbard_store;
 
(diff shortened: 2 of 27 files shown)


services/scabbard/libscabbard/src/service/v3/timer_filter.rs
```diff
@@ -219,7 +222,10 @@ mod tests {
             .with_alarm(updated_alarm)
             .with_coordinator(fqsi.clone().service_id())
             .with_epoch(1)
-            .with_participant_processes(vec![peer_service1.clone()])
+            .with_participants(vec![Participant {
+                process: peer_service1.clone(),
+                vote: None,
```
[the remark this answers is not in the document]
isabeltomb: I was thinking that we would differentiate with checking if context.coordinator() == context.this_process()

services/scabbard/libscabbard/src/store/mod.rs
```diff
@@ -19,7 +19,7 @@ mod command;
 mod commit_hash;
 #[cfg(any(feature = "postgres", feature = "sqlite"))]
 pub(crate) mod pool;
-#[cfg(feature = "scabbardv3")]
+#[cfg(feature = "scabbardv3-store")]
```
agunde406, line 22: This feature should also guard the pub use statements below

services/scabbard/libscabbard/src/service/v3/timer_filter.rs
```diff
@@ -174,7 +174,10 @@ mod tests {
             .with_alarm(SystemTime::now())
             .with_coordinator(fqsi.clone().service_id())
             .with_epoch(1)
-            .with_participant_processes(vec![peer_service1.clone()])
+            .with_participants(vec![Participant {
```
agunde406, line 177: Type in commit message: ParticiantContext -> ParticipantContext

services/scabbard/libscabbard/src/store/scabbard_store/diesel/models.rs
```diff
@@ -283,6 +279,51 @@ impl
...
+                    .ok_or_else(|| {
+                        InternalError::with_message(
+                        "Failed to convert to ScabbardContext, context has state 'voted' but vote
+                        is unset"
+                        .to_string(),
+                    )
+                    })?
+                    .ok_or_else(|| {
+                        InternalError::with_message(
+                            "Failed to convert to ScabbardContext, context has state 'voted' but an
```
agunde406, line 301: if the lines need to be wrapped they should have a backslash at the end of the first line, otherwise an unintended new line is added to the string (same for several other error messages)

agunde406: looks like this has been fixed in other parts of this commit but not in the two "Failed to convert to ScabbardContext" errors in this file

agunde406 (changes requested):

services/scabbard/libscabbard/src/store/scabbard_store/diesel/models.rs
isabeltomb: Thought I got all of those, should be fixed now

services/scabbard/libscabbard/src/service/v3/timer_filter.rs
isabeltomb: Fixed

services/scabbard/libscabbard/src/store/mod.rs
isabeltomb: Fixed

agunde406 (approved):

peterschwarz (approved):

PR #1950, opened 2022-05-05
Move biome endpoints

Caleb-Hill:
Moves biome key-management endpoints to the actix-web-1 crate

### DIFF
--- a/rest_api/actix_web_1/Cargo.toml
+++ b/rest_api/actix_web_1/Cargo.toml
@@ -45,6 +45,8 @@ stable = [
     # The following features are stable:
     "admin-service",
     "authorization",
+    "biome",
+    "biome-key-management",
     "rest-api",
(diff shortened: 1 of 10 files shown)


rest_api/actix_web_1/Cargo.toml
```diff
@@ -45,6 +45,8 @@ stable = [
     # The following features are stable:
     "admin-service",
     "authorization",
+    "biome",
```
agunde406, line 48: Typo in commit message: libsplinterd -> libsplinter

Caleb-Hill:
This only moves one endpoint and thats because the rest of the biome endpoints are tightly tied to the authorization code. So there will be a PR moving all the auth  endpoints including the rest of the biome endpoints

agunde406 (approved):

PR #1958, opened 2022-05-11
Move registry endpoints

Caleb-Hill:
Moves registry endpoints to actix_web_1 crate.


libsplinter/src/registry/mod.rs
```diff
@@ -28,9 +28,9 @@
 pub mod client;
 #[cfg(feature = "diesel")]
 mod diesel;
-mod error;
 #[cfg(feature = "rest-api-actix-web-1")]
 mod rest_api;
+pub mod error;
```
peterschwarz, line 31: I think we're tending towards re-exporting the items that should be public, vs making the module public

Caleb-Hill: Made module private, relevant types are re-exported bellow.

PR #1964, opened 2022-05-13
Scabbard store updates

isabeltomb:
- Fix a doc comment
- Fix module visibility
- Replace IdentifiedConsensusAction and IdentifiedConsensusEvent with a generic Identified<T> struct
- Remove epoch from the context table's primary key - there should only be one context per `service_id` when a new epoch starts the existing context should be updated rather than adding a new context

### DIFF
--- a/services/scabbard/libscabbard/src/service/v3/consensus/consensus_runner/mod.rs
+++ b/services/scabbard/libscabbard/src/service/v3/consensus/consensus_runner/mod.rs
@@ -26,9 +26,9 @@ use splinter::error::InternalError;
 use splinter::service::FullyQualifiedServiceId;
 use splinter::store::command::StoreCommandExecutor;
 
-use crate::store::action::ConsensusAction;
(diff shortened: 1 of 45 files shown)


services/scabbard/libscabbard/src/service/v3/consensus/consensus_runner/mod.rs
```diff
@@ -109,7 +109,7 @@ where
                 }
             };
 
-            let (event_id, event) = event.deconstruct();
+            let (event_id, event) = (event.id, event.record);
```
peterschwarz, line 112: We could still keep the deconstruct method.

isabeltomb: Added a deconstruct method for Identified<T>

peterschwarz (approved):

agunde406 (approved):

PR #1974, opened 2022-05-19
Move orchestrator rest api code

Caleb-Hill:
This PR updates the ServiceOrchestrator code to split out the resource providing parts and from the actual orchestration of services parts. Then it moves those resource providers to the actix-web-1 crate.


rest_api/actix_web_1/Cargo.toml
```diff
@@ -25,10 +25,10 @@ description = """\
 [dependencies]
 actix-web = { version = "1", optional = true }
 futures = { version = "0.1", optional = true }
-log = { version = "0.4", optional = true }
```
agunde406, line 28: These features should stay optional, and instead get pulled in if "service" or "admin-service" is pulled in.

agunde406 (changes requested):

agunde406 (approved):
Tested

vaporos (approved):

peterschwarz (approved):

PR #1977, opened 2022-05-23
Add store and operation to ConstraintViolationType

agunde406:
Previously, the ConstraintViolationType depended on the
source error to provide helpful information about the
violation. This resulted in errors that were too cryptic
to be helpful.

This commit adds from_source_with_violation_type_with_store
which adds the store name and operation that will be prefixed
in front of the source error:

Store operation: source error

### DIFF
--- a/libsplinter/src/error/constraint_violation.rs
+++ b/libsplinter/src/error/constraint_violation.rs
@@ -55,6 +55,8 @@ impl fmt::Display for ConstraintViolationType {
 pub struct ConstraintViolationError {
     violation_type: ConstraintViolationType,
     source: Option<Box<dyn error::Error>>,
+    store: Option<String>,
+    operation: Option<String>,
 }
 
 impl ConstraintViolationError {
@@ -77,6 +79,8 @@ impl ConstraintViolationError {
         Self {
             violation_type,
             source: None,
+            store: None,
+            operation: None,
         }
     }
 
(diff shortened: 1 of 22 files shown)


libsplinter/src/error/constraint_violation.rs
```diff
@@ -104,6 +108,43 @@ impl ConstraintViolationError {
...
+    /// let constraint_violation_error =
+    ///     ConstraintViolationError::from_source_with_violation_type_with_store(
+    ///          ConstraintViolationType::Unique,
+    ///          Box::new(db_err),
+    ///         "DbStore".to_string(),
+    ///         "db_operation".to_string()
+    /// );
+    /// assert_eq!(format!("{}", constraint_violation_error), "DbStore db_operation: db error");
+    /// ```
+    pub fn from_source_with_violation_type_with_store(
```
vaporos, line 137: Should this be "from_source_with_violation_type_and_store"? I thought we had an example of using "and" but I can't find it.

agunde406: updated

vaporos (dismissed):

services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/add_commit_entry.rs
```diff
@@ -42,7 +44,10 @@ impl<'a> AddCommitEntryOperation for ScabbardStoreOperations<'a, SqliteConnectio
             scabbard_service::table
                 .filter(scabbard_service::service_id.eq(format!("{}", commit_entry.service_id())))
                 .first::<ScabbardServiceModel>(self.conn)
-                .optional()?
+                .optional()
+                .map_err(|err| {
+                    ScabbardStoreError::from_source_with_operation(err, OPERATION_NAME.to_string())
```
peterschwarz, line 49: The usage here makes me wonder if this should just be a static string

peterschwarz: This is open to discussion, but also may be done in a follow-on PR.

peterschwarz (approved):

vaporos (approved):

PR #1978, opened 2022-05-23
Fix scabbard store update context

isabeltomb:
Fix the update context diesel scabbard store operation to do an update
rather than delete/insert when udpating a context.

### DIFF
--- a/services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/update_consensus_context.rs
+++ b/services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/update_consensus_context.rs
@@ -18,7 +18,7 @@ use std::convert::TryFrom;
 use diesel::pg::PgConnection;
 #[cfg(feature = "sqlite")]
 use diesel::sqlite::SqliteConnection;
-use diesel::{delete, dsl::insert_into, prelude::*};
+use diesel::{delete, dsl::insert_into, prelude::*, update};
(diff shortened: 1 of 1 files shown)


vaporos (approved):

peterschwarz (dismissed):

services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/update_consensus_context.rs
```diff
@@ -63,16 +63,22 @@ impl<'a> UpdateContextAction for ScabbardStoreOperations<'a, SqliteConnection> {
...
-                            consensus_2pc_context::service_id.eq(format!("{}", service_id)),
-                        ),
-                    )
-                    .execute(self.conn)?;
-
-                    insert_into(consensus_2pc_context::table)
-                        .values(vec![update_context])
-                        .execute(self.conn)?;
+                    update(consensus_2pc_context::table)
+                        .filter(consensus_2pc_context::service_id.eq(format!("{}", service_id)))
```
peterschwarz, line 67: Note, instead of using `format` here, you can just call `service_id.to_string()`

isabeltomb: Updated, will make a PR in the future to update this elsewhere

agunde406 (approved):

PR #1982, opened 2022-05-24
Replace the use of epoch with id in CommitEntry

agunde406:
This removes a dependency between the commit entry used
for the scabbard prototype and the context which should
be internal to consensus.


peterschwarz (approved):

isabeltomb (approved):

agunde406:
This has been decided to not be needed

PR #2000, opened 2022-06-08
Split out Permission from authorization::mod.rs

Caleb-Hill:
This commit moves the `Permission` enum into its own file out of the
mod.rs file. The goal is to keep mod.rs concise and more tightly
focused.


Caleb-Hill:
Should have been part of another PR, commit is included here https://github.com/Cargill/splinter/pull/2003.

PR #2009, opened 2022-06-10
Update ScabbardMessageHandler to use a TimerAlarm after msg is received

agunde406:
This PR updates the Timer to return a TimerAlarmFactory. Returning the factory enables passing an alarm factory to the ScabbardMessageHandlerFactory so that the ScabbardMessageHandler can use an alarm to wake up the Timer after a message has been received.

### DIFF
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -211,7 +211,7 @@ service-timer =[
   "defered-send",
   "runtime-service",
   "service-message-sender-factory",
-  "service-timer-alarm",
+  "service-timer-alarm-factory",
   "service-timer-filter",
   "service-timer-handler",
   "service-timer-handler-factory"
--- a/libsplinter/Cargo.toml
+++ b/libsplinter/Cargo.toml
@@ -138,6 +138,7 @@ experimental = [
     "service-message-sender-factory-peer",
     "service-timer",
     "service-timer-alarm",
+    "service-timer-alarm-factory",
     "service-timer-filter",
     "service-timer-handler",
     "service-timer-handler-factory",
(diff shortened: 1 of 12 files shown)


libsplinter/src/service/timer_alarm_factory.rs
```diff
@@ -0,0 +1,31 @@
...
+//! Contains `TimerAlarmFactory` trait.
+
+use super::TimerAlarm;
+
+/// Used to create new `TimerAlarm` instances.
+pub trait TimerAlarmFactory: Send {
+    /// Returns a new `TimerAlarm`
+    fn new_alarm(&self) -> Box<dyn TimerAlarm>;
+
+    fn clone_box(&self) -> Box<dyn TimerAlarmFactory>;
```
vaporos, line 24: Is this clone_box necessary? I couldn't find where its used in subsequent commits.

agunde406: It is only used directly in the impl of Clone below. That is required because the ScabbardMessageHandlerFactory, which will get a boxed TimerAlarmFactory, must be cloneable.

libsplinter/Cargo.toml
```diff
@@ -211,7 +211,7 @@ service-timer =[
   "defered-send",
   "runtime-service",
   "service-message-sender-factory",
-  "service-timer-alarm",
```
peterschwarz: Shouldn't both of these features exist? Also, this doesn't seem to be in the correct commit.

agunde406: This is updating which feature the `service-timer` requires. Which is required as of this commit. The  "service-timer-alarm-factory" pulls in "service-timer-alarm". No features are actually removed here.

vaporos (approved):

PR #2016, opened 2022-06-16
Scabbard store enhancements

isabeltomb:
- Split service_id (fully qualified service ID) in consensus tables into circuit_id and service_id
- Change fk relationship to point to the scabbard_service table rather than consensus_2pc_context


agunde406 (approved):
Tested

vaporos (approved):

PR #2027, opened 2022-06-22
Implement Supervisor that works with the Consensus Layer 

agunde406:
This PR adds a Supervisor that will handle pending notifications
that have been received from the consensus layer. The supervisor
currently handles each notification in the same way that the
CommandNotifyObserver did.

When a supervisor notification is added to state the Supervisor
should be notified. This must happen after the commands have been
executed, as such the notifier will be added to the timer handler.


services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/list_ready_services.rs
```diff
@@ -108,6 +111,22 @@ where
                     })?,
             );
 
+            ready_services.append(
```
isabeltomb, line 114: should this query also check to make sure that the services with unexecuted supervisor notifications are in the finalized state?

services/scabbard/libscabbard/src/store/scabbard_store/supervisor.rs
```diff
@@ -0,0 +1,218 @@
...
+        &self.action_id
+    }
+
+    /// Returns the notification_type for the notficiation
+    pub fn notification_type(&self) -> &SupervisorNotificationType {
+        &self.notification_type
+    }
+
+    /// Returns the value for the notficiation
+    pub fn request_for_vote_value(&self) -> &Option<Vec<u8>> {
```
vaporos, line 48: Consider include this in the enum to avoid the Option

agunde406: Updated to remove request_for_vote_value and use `SupervisorNotificationType::ParticipantRequestForVote { value: Vec<u8> },` instead

services/scabbard/libscabbard/src/store/scabbard_store/supervisor.rs
```diff
@@ -0,0 +1,218 @@
...
+        &self.request_for_vote_value
+    }
+
+    /// Returns the executed_at time for the notficiation
```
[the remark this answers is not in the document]
agunde406: fixed

services/scabbard/libscabbard/src/store/scabbard_store/diesel/operations/list_ready_services.rs
agunde406: added

isabeltomb (dismissed):

vaporos (approved):

isabeltomb (approved):

vaporos (approved):

PR #2046, opened 2022-07-06
Scabbard store fixes

isabeltomb:
- Update database types that use text type to represent a boolean value to be boolean types
- put underscores in the string options that are multiple words for example 'VOTERESPONSE' -> 'VOTE_RESPONSE'
- add a sqlite migration to change the `created_at` column in the consensus_2pc_action table to be UTC instead of localtime to match the `executed_at` column


agunde406 (approved):
tested

PR #2048, opened 2022-07-11
Add SignedTimestamp artifact, creator and factory 

agunde406:
These components will be used by the new Publisher design

### DIFF
--- a/services/scabbard/libscabbard/src/service/v3/publisher/publisher_context.rs
+++ b/services/scabbard/libscabbard/src/service/v3/publisher/publisher_context.rs
@@ -1,0 +1,42 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


services/scabbard/libscabbard/src/service/v3/publisher/publisher_context.rs
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Contains the publishing context for Scabbard v3
+
+use cylinder::Signer;
+use splinter::service::FullyQualifiedServiceId;
+
+pub struct ScabbardContext {
```
peterschwarz, line 20: Should this be named more specifically the `ScabbardPublishingContext` to differentiate from the consensus context?

peterschwarz (approved):

PR #2051, opened 2022-07-13
Pin Rust version and add package-lock.json

rbuysse:


agunde406 (approved):
Tested, there may be a regression unrelated to this PR in gameroom that will be tracked separately.

Caleb-Hill (approved):

PR #2060, opened 2022-07-20
Add echo service to splinter

isabeltomb:
Add the Echo service to splinter. Echo service is a simple service that can be used for testing a circuit between nodes. An echo service sends messages to its peers at a set frequency.

### Testing:

Start two splinter nodes with experimental features enabled

Create a circuit with echo as the service types between the two nodes:
```
cargo run --manifest-path cli/Cargo.toml --features=experimental circuit propose \
    --url http://127.0.0.1:28080 \
    --key <node2-private-key> \
    --display-name "echo" \
    --node node1::tcp://127.0.0.1:18044 \
    --node node2::tcp://127.0.0.1:28044 \
    --node-public-key node1::<node1-public-key> \
    --node-public-key node2::<node2-public-key> \
    --service a000::node1 \
    --service b000::node2 \
    --service-peer-group a000,b000 \
    --service-type "*::echo" \
    --service-arg "*::frequency=5" \
    --service-arg "*::jitter=2" \
    --service-arg "*::error_rate=0.05" \
    --management tutorial \
    --auth-type trust
```
accept the proposal:
```
splinter circuit vote --accept \
  --key home/alpha.priv \
  --url http://127.0.0.1:18080 <circuit-id>
```
you should start seeing logs similar to the following:
```
[2022-07-20 15:25:19.859] T[MessageHandlerTaskPool-2] INFO [splinter_echo::service::message_handler] [service:ZlNG2-vBdwu::b000] [from:ZlNG2-vBdwu::a000] [id:1] received echo response: "test"
[2022-07-20 15:25:19.859] T[MessageHandlerTaskPool-6] INFO [splinter_echo::service::message_handler] [service:ZlNG2-vBdwu::b000] [from:ZlNG2-vBdwu::a000] [id:1] received echo request, sending echo response: "test"`
```


splinterd/Cargo.toml
```diff
@@ -112,6 +113,7 @@ experimental = [
     "service-endpoint",
     "service-timer-interval",
     "service2",
+    "service2-echo",
```
agunde406, line 116: I think we should change the name of this feature to just `service-echo` or `echo`

services/echo/libecho/src/migrations/diesel/sqlite/src/schema.rs
```diff
@@ -0,0 +1,46 @@
+table! {
```
agunde406, line 1: Is this schema.rs file required? there is another one added in a later commit that is not sqlite specific

agunde406 (changes requested):

services/echo/libecho/src/migrations/diesel/sqlite/src/schema.rs
isabeltomb: It is not required and I thought I already deleted that, should be gone now

splinterd/Cargo.toml
isabeltomb: updated to `service-echo`

agunde406 (changes requested):
One small comment, otherwise tested with sqlite and postgres and LGTM

services/echo/libecho/src/service/message_handler.rs
```diff
@@ -0,0 +1,143 @@
...
+};
+use std::convert::TryFrom;
+use std::time::SystemTime;
+
+use super::EchoMessage;
+
+use crate::service::RequestStatus;
+use crate::store::EchoStore;
+
+// #[derive(Default)]
```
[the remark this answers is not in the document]
isabeltomb: fixed

agunde406 (approved):

peterschwarz (approved):

